Issue #4, opened 2020-07-30
Create golang library for validating packages against specs


ycombinator:
Create a golang library (go module) that provides an API for validating a package against the appropriate specifications. The library will need to bundle all specification versions with it.

* [x] Foundation: https://github.com/elastic/package-spec/pull/12
* [x] Folder structure validation: https://github.com/elastic/package-spec/pull/14
* [x] File contents validation #41

ycombinator:
@mtojek I think the last item in the checklist is done now? If so, could you link to your PR that resolved it, check it off, and close this issue? Thanks!

PR #12, opened 2020-08-06
Golang library for validation: foundation

ycombinator:
This PR creates the foundations for a Golang library for package validation. Concretely, it creates the following:
- Makefile(s) for building the library, which includes building the code but also bundling the package specifications into the library, and
- a go module that provides an API for validating a package against the appropriate specification version. This API can be used as follows:
   ```go 
   import "github.com/elastic/package-spec/code/go/pkg/validator"

   validationErrs := validator.ValidateFromPath("/path/to/package/root/folder")
   ```

   where `validationErrs` will implement the [`error` interface type](https://blog.golang.org/error-handling-and-go#TOC_2.).

Note that this PR does not implement the actual validation logic yet. That is, calling the API will not return any validation errors at the the moment. The actual validation logic will be implemented in follow up PRs.

Related: #4


code/go/pkg/validator/validator.go
```diff
@@ -0,0 +1,13 @@
+package validator
+
+import (
+	"github.com/elastic/package-spec/code/go/internal/validator"
```
mtojek, line 4: I'm not quite convinced about the import path as it contains many package levels. Ideally it would be `github.com/elastic/package-spec/validator`, but I assume that `code/` will contain implementation also for JS.

Fine for me!

code/go/pkg/validator/validator.go
```diff
@@ -0,0 +1,13 @@
...
+)
+
+// ValidateFromPath validates a package located at the given path against the
+// appropriate specification and returns any errors.
+func ValidateFromPath(packageRootPath string) error {
```
mtojek, line 9: I see you run this as static method. I admit I didn't go deeper yet and might be wrong, but do you cache loaded specs in some global validator? just not to load schemas every single time, a user validates a package.

code/go/internal/validator/errors.go
```diff
@@ -0,0 +1,27 @@
+package validator
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ValidationErrors is an Error that contains a iterable collection of validation error messages.
```
mtojek, line 8: nit: an error
nit: an iterable

code/go/internal/spec_test.go
```diff
@@ -0,0 +1,17 @@
+package internal
+
+import (
+	"testing"
+
+	_ "github.com/elastic/package-spec/code/go/internal/spec"
```
mtojek, line 6: nit: you can sort these imports

code/go/pkg/validator/validator.go
ycombinator: ++ I'd love for it to be simpler too but I don't know how to make that happen. Indeed the reason was to allow for other language libraries to be added in the future (I copied the structure from ECS: https://github.com/elastic/ecs/tree/master/code).

code/go/internal/spec_test.go
ycombinator: Thanks for catching this. I need to set this up in my IDE!

code/go/pkg/validator/validator.go
ycombinator: Yes, I'm using https://github.com/rakyll/statik which results in the code/go/internal/spec/statik.go as a result of `make update` and then that file gets compiled in with the rest of the code. So by the time the validation library is called, the specs are in memory already.

code/go/pkg/validator/validator.go
mtojek: Just thinking loud: you can leave specification files in package-spec and depend on them in implementations: go-package-validator, js-package-validator. But I believe it might be an overkill here as it's relatively small.

As I said, I'm happy to leave it as is or add a single facade file on repo root (so all imports would be: `github.com/elastic/package-spec`).

ycombinator: > Just thinking loud: you can leave specification files in package-spec and depend on them in implementations: go-package-validator, js-package-validator. But I believe it might be an overkill here as it's relatively small.

You mean pull out the libraries into their own repos? Or something else?

> As I said, I'm happy to leave it as is or add a single facade file on repo root (so all imports would be: `github.com/elastic/package-spec`).

This could work but I'm not sure what this means for other languages. I think maybe this is something we can revisit once we add JS validation too. I know it would mean updating existing clients at that point but hopefully that's okay since adding a facade would be a backwards-compatible change (at least for some period of time).

code/go/pkg/validator/validator.go
mtojek: With statik you can keep raw files in memory, but you need to initialize structures with loaded data. I was thinking about following this model:

1. Load all specs from in-memory filesystem.
2. Build validators once.
3. Use validators against packages.

I understand it might be a pre-optimization, but hopefully may save memory by not initializing validators on every call.

ycombinator: Yeah, this optimization is definitely beyond the scope of this PR. This PR is mainly about exposing the right API and making sure spec files are bundled in with the library (using statik). 

So I would like to consider this optimization in future PRs, e.g. https://github.com/elastic/package-spec/pull/14. Even then I would consider this a premature optimization and try to defer it to a time when we actually need it, e.g. start to see a perceptible slowness or larger-than-expected memory utilization when running `elastic-package lint`. Until then I would prefer to prioritize getting the API right and a functioning implementation — we can always improve the implementation later, as needed.

code/go/internal/spec_test.go
ycombinator: According to my IDE (goland) these imports are already sorted. I deliberately tried reordering them and my IDE re-sorted them into their current position. 🤔

code/go/internal/validator/errors.go
ycombinator: Fixed in c0fcc8d.

code/go/internal/spec_test.go
ycombinator: Imports should be sorted now.

Issue #19, opened 2020-08-12
[Discuss] Impact of renaming config to policy


ruflin:
It was decided to rename Agent Config to Agent Policy. This also applies to all config blocks inside the config blocks inside the config, which are now also policies.

The main usage of config inside the package spec seems to be `config_template` which could become `policy_template` instead.

ruflin:
@mtojek @ycombinator Thoughts on the above? Perhaps this is also an opportunity to find a better name?

mtojek:
> It was decided to rename Agent Config to Agent Policy. This also applies to all config blocks inside the config blocks inside the config, which are now also policies.

It would be valuable to put a sample config here as the description is a bit blurry. What are the policies? How much do they differ from what we have now?

ruflin:
Policy = Config. There is no difference form a pure yaml perspective structure except the name. One thing that changes is if `dataset.*` was used inside the stream templates it now has to be `data_stream.*`, but this is specific to each package and only effected the log package so far.

ycombinator:
Yeah, I'm also not 100% clear on what are agent configs in relation to packages. To put it differently: where exactly in packages do agent configs show up? It might be helpful to post some links to agent configs in existing packages.

ruflin:
An example config can be found here: https://github.com/elastic/beats/blob/master/x-pack/elastic-agent/elastic-agent.yml But lets also try to talk it through, I organised a meeting tomorrow and then we follow up here.

ycombinator:
@ruflin @mtojek I think we can close this now thanks to https://github.com/elastic/package-spec/pull/50?

mtojek:
Yes. Resolving.

PR #14, opened 2020-08-06
Golang library for validation: folder structure validation

ycombinator:
This PR implements validation of a package's folder structure. Structure of package files is not yet implemented and will be done in a follow up PR.

Related: #4.

PR #24, opened 2020-08-13
Dataset to data stream

### LINKED ISSUE #18: [Discuss] Impact of renaming dataset to data_stream
It was decided recently to rename the `dataset.*` fields to `data_stream.*` fields and quite a few changes were already made for this in 7.9. No changes have happened in the package spec so far. This issue is to discuss what impact this rename has on the package spec.

ycombinator:
This PR updates the spec to rename `dataset` references in packages to `data_stream`.

Concretely, the following elements of packages are changed:
- In the package's root `manifest.yml` file, the `datasets` property must now be called `data_streams` instead.
- In the package's root folder, the `dataset` folder must now be called `data_stream` instead.

Resolves #18.


versions/1/spec.yml
```diff
@@ -17,11 +17,11 @@ spec:
   name: "changelog.yml"
   required: true
   $ref: "./changelog.spec.yml"
-- description: Folder containing dataset definitions
+- description: Folder containing data stream definitions
   type: folder
-  name: dataset
+  name: data_stream
```
ycombinator, line 22: While we're at it, should we rename this folder to `data_streams` (plural), since it typically contains multiple sub-folders, one for each data stream?

ruflin: I personally like that it is singular because each file path is a single data stream: `data_stream/access/`. But I know quite a few people disagree and a discussion me and @exekias have for years 😆  So I'm good with both.

versions/1/data_stream/manifest.spec.yml
```diff
@@ -1,37 +1,37 @@
 ##
-## Describes the specification for a dataset's manifest.yml file
+## Describes the specification for a data stream's manifest.yml file
 ##
 spec:
   # Everything under here follows JSON schema (https://json-schema.org/), written as YAML for readability
   type: object
   additionalProperties: false
   properties:
     name:
```
ruflin, line 9: This would have to become `dataset` as the `dataset.name` property was renamed to `data_stream.dataset`.

versions/1/data_stream/spec.yml
```diff
@@ -1,10 +1,10 @@
 spec:
-- description: Folder containing a single dataset definition
+- description: Folder containing a single data stream definition
```
ruflin, line 2: Stumbled over this as it is not fully current. This a template for all data streams created for a specific `type + dataset` combination. All `logs-nginx.access-*` will use this template, independent of the namespace.

In my head, these are data stream templates.

mtojek: I'm afraid I'm lost here. @ruflin, could you please elaborate more about this? data streams, templates, namespaces

versions/1/spec.yml
mtojek: I vote for the plural, similarly to REST naming conversion:

1. Document is singular: account, customer
2. Set of documents is plural: accounts, customers

so: data_streams

mtojek: btw. do we know the scope of changes caused by this single update?

ycombinator: > btw. do we know the scope of changes caused by this single update?

Once https://github.com/elastic/package-spec/pull/14 is merged (we're pretty close on that one I think), we can run `elastic-package lint` with this PR here and get an idea of the changes that will be needed across all the packages.

versions/1/data_stream/manifest.spec.yml
ycombinator: Fixed in 383c692.

versions/1/data_stream/spec.yml
ycombinator: Yes, I'm lost too. I think it might be useful to add a `definitions.md` or just a `README.md` file in the `versions/1` folder defining all these terms and concepts and how they relate to each other. The spec (so far) describes the syntactic relationships but I also have to admit that I don't fully understand all the semantic relationships between these concepts.

@ruflin can you please make a new PR to add such a file? I think you are the only(?) one who understands all these terms (old and new), the corresponding concepts, and the relationships between them.

versions/1/spec.yml
ycombinator: Looking at some other existing package folders that represent collections of things, we don't have any consistency ☹️ . Some collection folders use singular names, e.g., `kibana/dashboard/`, `kibana/visualizations/`, and `img/`. And some collection folders use plural names, e.g. `docs/` and `dataset/<foo>/fields`. 

I don't have strong opinions in either direction but I slightly favor plural names for folders that represent collections of things. I do have a strong opinion about trying to be consistent 🙂.

ruflin: @ycombinator There is something indeed odd with `visualizations` vs `datashboard`. I thought we took the name of the saved object here but it seems `visualization` is singular in KB :-( ++ on consistency.

ruflin: I checked some packages and couldn't find visualizations? Where did you spot that?

ycombinator: Sorry @ruflin, that was a typo in my comment above. I meant `kibana/visualization` not `kibana/visualizations`. Edited and fixed now.

ycombinator:
@ruflin @mtojek I'm resurrecting this PR after a while. I've rebased it on `master` so it has all the other changes now. Could you please re-review when you get a chance? Thanks!

mtojek:
Ok, so once you push this change to master, we'll block introducing next changes to this repository until the package-registry and all integrations are adjusted?

ycombinator:
> Ok, so once you push this change to master, we'll block introducing next changes to this repository until the package-registry and all integrations are adjusted?

Yes. But to make the transition quick, I would say we should try to get the PRs ready in `package-registry` and `integrations` repos along side this one so we can merge all three at around the same time. We can also merge this PR into `master` and update `elastic-package`, but then make a PR to `integrations` updating `elastic-package` and making the necessary adjustments to the packages.

mtojek:
I think we can safely push this to master. Then:
1. Update dependency on the elastic-package.
2. Adjust package-registry (not sure if this is safe to push changes, @ruflin ?)
3. Adjust integrations.

Points 1. and 2. can go in parallel.

mtojek:
Actually the same question goes to Kibana. Are we safe to introduce such renaming or is it better to schedule it? @ruflin

ruflin:
We should schedule this change together with Kibana. For my two recent breaking changes to the registry @jfsiii has worked some magic that both version work which is also a best case scenario here. If we can do that, it means Kibana merges + Backports this first, then we get it in and roll it out to all of the parts mentioned above. Can you sync up with @jfsiii on this?

ycombinator:
I'm unclear as to why this PR needs to be scheduled with a Kibana PR. This PR just changes the spec. It does not change any packages. So I'm not understanding the impact of this PR on Kibana. What am I missing?

ruflin:
@ycombinator If we get this PR in without the changes ready on the Kibana side, it means if we have other changes to the package-spec and want to roll it out, we can't as it will break things. So just merging it will not have a direct impact on Kibana but it will hold development back on other things in the spec.

jfsiii:
@ruflin I'm not totally clear on the how the spec changes map to Registry & Kibana code/type changes but am happy to sync up with @mtojek. @jen-huang did a lot of `dataset.*` to `data_stream.*` conversion in https://github.com/elastic/kibana/commit/042254f026bd77e4411d056c2af305588c8974ed 

We do reference `datasets` in our `RegsitryPackage` type  https://github.com/elastic/kibana/blob/87ca6ff70c90144f010c1757f309aac8f53608ce/x-pack/plugins/ingest_manager/common/types/models/epm.ts#L70 which mirrors the Registry's Package.Datasets https://github.com/elastic/package-registry/blob/51947a80a0b45e0cd8e0f4c241e4886a89eacd68/util/package.go#L65

ruflin:
@jfsiii @ycombinator Could you sync up to push this forward? Please pull me in if needed. Please directly also discuss https://github.com/elastic/package-spec/issues/19 to make sure everyone is aware of all the changes coming.

mtojek:
@ycombinator I'm not sure why the CI didn't report problems with test directories:

https://github.com/ycombinator/package-spec/tree/dataset-to-data_stream/code/go/internal/validator/test/packages/bad_deploy_variants
https://github.com/ycombinator/package-spec/tree/dataset-to-data_stream/code/go/internal/validator/test/packages/good

There are still "dataset" folders and it failed for me in the local environment.

versions/1/data_stream/manifest.spec.yml
```diff
@@ -68,32 +68,32 @@ spec:
           - name
           - type
   properties:
-    name:
-      description: Name of dataset
+    data_stream:
```
mtojek, line 71: @ycombinator Is this rename correct? `name->data_stream`?

ycombinator: Yes, see https://github.com/elastic/package-spec/pull/24#discussion_r470464135.

mtojek: Ah... and I even approved it :) Ok, let's leave it here.

jen-huang: shouldn't this be `name`->`dataset`, not `data_stream`? I see it was `dataset` up until the last commit f706b65

ycombinator: Hmm, yes, you're right — it should be `dataset` not `data_stream` per https://github.com/elastic/package-spec/pull/24#discussion_r470464135. Fixing...

ycombinator: Fixed in 37fd8917ea45530bf5885e50efa0c76344d0e0a0. Thanks!

ycombinator:
@mtojek Looks like CI just failed as expected: https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Fpackage-spec/detail/PR-24/6/pipeline. Yay!

Fixing...

mtojek (approved):

Issue #20, opened 2020-08-12
Add validation that owner is set


ruflin:
The owner of a package is becoming more important. I think every package "must" specify an owner. The owner can be a group or a single person. The validation script should be able to validate that an owner exists. 

An addition to that would be that it even validates if the group / users exists on Github but that might become excessive and cause permissions problems.

ycombinator:
> The validation script should be able to validate that an owner exists.

Concretely what does it mean for an owner to exist? Are owners the same as a user or repo on GitHub? Or is it a team on GitHub? Based on what you said about permissions I think an owner is either a user or team on GitHub — if so, then yes, we could probably validate the existence of a user but not that of a team.

ruflin:
Yes, owner should be user or team.

jsoriano:
Owner is already required, and we also enforce the presence of the owners in the GITOWNERS file in the integrations repository. Closing this. Please reopen if something is missing.

PR #25, opened 2020-08-13
Making owner required

ycombinator:
This PR makes the `owner` root-level field in a package's root-level `manifest.yml` file required. 

It does **not** add any semantic validation of the value of the field, such as checking if the value corresponds to an actual owner entity in some system. This may not be possible or might require certain security privileges so let's tackle this in a follow up PR.

Resolves partially #20.


versions/1/manifest.spec.yml
```diff
@@ -233,4 +233,5 @@ spec:
   - title
   - description
   - version
-  - release
\ No newline at end of file
+  - release
+  - owner
```
ruflin, line 237: Shouldn't this be `owner.github` as this is the only property we currently use? https://github.com/elastic/package-storage/blob/production/packages/apache/0.1.4/manifest.yml#L49

mtojek: +1 

same question here

ycombinator: Yup, good catch! Will fix, thanks!

ycombinator: Fixed in 57f1dcc.

Issue #38, opened 2020-05-04
Validate YAML files


ruflin:
All available yaml files in the packages should be validated for valid yaml to make sure packages only contain valid yaml.

An example where this would have been useful can be found here: https://github.com/elastic/package-registry/pull/421

mtojek:
Let me try to handle this.

PR #41, opened 2020-08-27
Validate content of folder items

mtojek:
_I'm opening this as draft for visibility of the progress and changes._

Fixes: https://github.com/elastic/package-spec/issues/38

This PR adds support for validating file content (JSON or YAML) against JSON-schema.

TODO:

- [x] extract folder item to a separate file
- [x] call item validate's method
- [x] load JSON-schema from YAML file
- [x] load folder item content based on the content-type (MIME)
- [x] validate the file content and report problems
- [x] code cleanup
- [x] I went through integrations and adjusted few schema files to make them passing (mostly missing field types)
- [x] refs to external item schemas are not supported at the moment
- [x] support inline specs for folder items

Notes:
- YAML fields encoded like: 

```
conditions:
  kibana.version: "^7.10.0"
```

... are not extended to:

```
conditions:
  kibana:
    version: "^7.10.0"
```

If there is a quick and nice fix, I can apply it. At the moment I enforced the upper version.


versions/1/manifest.spec.yml
```diff
@@ -202,10 +198,69 @@ spec:
                 description:
                   description: Longer description of input.
                   type: string
-                  examples: 
+                  examples:
                   - Collecting Apache access and error logs
                 vars:
-                  $ref: "./dataset/manifest.spec.yml#streams/items/properties/vars"
+                  description: Input variables.
```
ycombinator, line 204: Aren't we duplicating this spec? That's why I had the `$ref` originally.

versions/1/manifest.spec.yml
```diff
@@ -216,10 +271,6 @@ spec:
         - name
         - title
         - description
-    datasets:
-      type: array
-      items:
-        $ref: "./dataset/manifest.spec.yml"
```
ycombinator, line 222: Hmm, I wonder why this was there 🤔

versions/1/dataset/_dev/test/pipeline/event.spec.yml
```diff
@@ -12,4 +12,4 @@ spec:
       items:
         type: object
         additionalProperties: true
-      required: true
```
ycombinator, line 15: Thanks for fixing!

versions/1/dataset/_dev/test/pipeline/expected.spec.yml
```diff
@@ -12,4 +12,4 @@ spec:
       items:
         type: object
         additionalProperties: true
-      required: true
```
ycombinator, line 15: Thanks for fixing!

code/go/internal/validator/test/packages/good/dataset/foo/_dev/test/pipeline/test-access-raw.log-expected.json
```diff
@@ -1,5 +1,5 @@
 {
-    "1expected": [
```
ycombinator, line 2: Thanks for fixing!

versions/1/manifest.spec.yml
mtojek: Yes, but the provided loader doesn't support the statik filesystem. I can try to implement a custom resource loader.

Also, the library enforces a canonical path here with schema, so it would be: `file://./dataset/manifest.spec.yml#streams/items/properties/vars`

versions/1/manifest.spec.yml
mtojek: I have no idea, I didn't find any references.

code/go/internal/validator/test/packages/good/dataset/foo/_dev/test/pipeline/test-access-raw.log-expected.json
mtojek: That one was actually on purpose... it would fail when somebody implements the validation :)

code/go/internal/validator/folder_item_spec.go
```diff
@@ -0,0 +1,166 @@
...
+	case itemTypeFolder:
+		return file.IsDir()
+	}
+
+	return false
+}
+
+func (s *folderItemSpec) validate(fs http.FileSystem, folderSpecPath string, itemPath string) ValidationErrors {
+	if s.Ref == "" {
+		return nil // no item's schema defined
```
ycombinator, line 68: What is the schema is defined inline and not referenced via `$ref`?

code/go/internal/validator/folder_item_spec.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func loadItemSchema(fs http.FileSystem, itemSchemaPath string) ([]byte, error) {
+	itemSchemaFile, err := fs.Open(itemSchemaPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "opening schema file failed")
+	}
+	defer itemSchemaFile.Close()
+
+	itemSchemaData, err := ioutil.ReadAll(itemSchemaFile)
```
ycombinator, line 98: Any reason not to use `ioutil.ReadFile` here directly instead of opening the file handle with `fs.Open` first and then reading from it using `ioutil.ReadAll`?

code/go/internal/validator/folder_item_spec.go
mtojek: not sure if I follow you...
this might be bad wording, should be: schema reference not provided. Does it sound better?

code/go/internal/validator/folder_item_spec.go
mtojek: We need to read data from statik resource. I'm afraid that `ioutil.ReadFile` doesn't support it this way.

code/go/internal/validator/folder_item_spec.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, errors.New("schema file is empty")
+	}
+
+	var schema itemSchemaSpec
+	err = yaml.Unmarshal(itemSchemaData, &schema)
+	if err != nil {
+		return nil, errors.Wrapf(err, "schema unmarshalling failed (path: %s)", itemSchemaPath)
+	}
+
+	schemaData, err := json.Marshal(&schema.Spec)
```
ycombinator, line 113: Not for this PR but we could probably convert the schema YAML files to JSON as part of the [`make update` step](https://github.com/elastic/package-spec/blob/f45b05c887b5d62dea8f5c21577520a0e113716c/Makefile#L2) so we don't have to do this `yaml.Unmarshal` + `json.Marshal` dance every time at run time.

mtojek: Yes, this could be an improvement in the future (convertion is always error prone), although I admin that YAML here is much more human readable and easier to interact.

versions/1/manifest.spec.yml
ycombinator: I see. I think the duplication is fine for now but we should probably make comments in both places pointing to the other as a duplicate that needs to be kept in sync whenever any one of the places is changed. Longer term, let's look into implementing a custom resource loader.

mtojek: I will try to focus on this as followup. Got interested with this :)

code/go/internal/validator/folder_item_spec.go
ycombinator: Sorry, I was referring to needing something like this: https://github.com/elastic/package-spec/blob/master/code/go/internal/validator/folder.go#L125-L141. Maybe this is not the right place for it but then it should be added where this method is called from.

code/go/internal/validator/folder_item_spec.go
ycombinator: Ha, I got tripped up by the `fs` there! Totally makes sense why you have it the way it is. Ignore me. 🙂

code/go/internal/validator/folder_item_spec.go
ycombinator: Indeed, which is why we'd keep the source spec files in YAML but only convert them to JSON when `make update` is run.

code/go/internal/validator/folder_item_spec.go
mtojek: Ah, now I see... I totally missed the case where schema is defined inline, didn't know about this. I will improve the implementation.

mtojek: @ycombinator Hm.. I dived into this topic and got a bit confused. Do we really need it? What would a use case for this? This `$ref` refers to a place in folder schema in which we refer to item's schema. It's not part of the JSON-schema, so I believe we can enforce to keep it in a separate file (to separate concerns and prevent from creating a huge all-in-one files).

tl;dr `$ref` in folder schema is always external.

WDYT?

ycombinator: So in the folder schema, there can be two `type`s of items: `folder` (for describing the structure of a sub-folder within the folder) and `file` (for describing the structure of a file within the folder). 

For `type: folder`, we already have a couple of places in the spec where use `$ref` ([example](https://github.com/elastic/package-spec/blob/f45b05c887b5d62dea8f5c21577520a0e113716c/versions/1/dataset/spec.yml#L20)) and a couple other places where we inline the definition instead of using `$ref` ([example](https://github.com/elastic/package-spec/blob/f45b05c887b5d62dea8f5c21577520a0e113716c/versions/1/dataset/spec.yml#L26)).

For `type: file`, I checked, and _at the moment_ we only have places where we use `$ref` (example). We do not yet have any places where we are inlining the spec (aka schema) for the file itself. However, I can see this being useful for simple file specs ([example](https://github.com/elastic/package-spec/blob/master/versions/1/dataset/_dev/test/pipeline/event.spec.yml)).

 So I think there is a use case for it and we should support it.

Also, if we allowing inline specs for sub-folder items but not for file items, we are introducing an inconsistency. We could, of course, resolve this inconsistency by going the other way: that is, we can say that we _only_ allow `$ref`s for item specs, regardless of whether those are folder items or file items; that we do not accept inline specs for either item type. I think this is slightly inconvenient when the item has a small spec but I'd be okay with this approach too. If you prefer this, then let's make a separate PR to first remove support for inline specs for `type: folder` items and adjust the spec files accordingly.

mtojek:
Ok, I fixed the issue with schema referenced inside item specs. The solution is that all "exported" elements must be specified under definitions. Then we can refer to them (out of the box, without JSON pointer magic) via:

`$ref: "./dataset/manifest.spec.yml#/definitions/vars"`

(for input vars)

ycombinator:
> The solution is that all "exported" elements must be specified under definitions

Hmm, according to https://json-schema.org/understanding-json-schema/structuring.html it's not required to put the "exported" elements under `definitions`:

> Since we are going to reuse this schema, it is customary (but not required) to put it in the parent schema under a key called `definitions`:

Maybe this is something required by the library we are using for validating JSON schema. Anyway, good find and it's actually quite nice to have a well-defined location for "exported" vars so 👍 all around!

PR #45, opened 2020-09-10
Adding specs needed for system testing

ycombinator:
This PR adds specs needed for system testing packages via the `elastic-package test system` CLI command. 

## Testing this PR
- You can test this PR in conjunction with https://github.com/elastic/integrations/pull/263. 
- You will need to add a `replace` directive to your `elastic-package`'s `go.mod` file for `github.com/elastic/package-spec/code/go` and point it to your local folder where this PR is checked out. 
- Then run `make build` on `elastic-package`
- Finally, run `elastic-package lint` in the `apache` package folder.


code/go/pkg/validator/validator_test.go
```diff
@@ -1,13 +1,43 @@
...
 func TestValidate(t *testing.T) {
-	pkgRootPath := "../../internal/validator/test/packages/good"
-	errs := ValidateFromPath(pkgRootPath)
-	require.NoError(t, errs)
+	tests := map[string]struct {
+		expectedErrContains []string
+	}{
+		"good": {},
+		"bad_deploy_variants": {
+			expectedErrContains: []string{
```
mtojek, line 18: Such implementation verifies the expected error messages, but it doesn't relate them with particular files. It's vulnerable to pass the test if validation fails for any file.

code/go/internal/validator/test/packages/bad_deploy_variants/_dev/deploy/variants.yml
```diff
@@ -0,0 +1,2 @@
+variants:
+  - FOO: bar
```
mtojek, line 2: For clarity you can leave a comment here what's wrong with this file (missing `default`, misused array).

ycombinator:
@mtojek Thanks for the review feedback. I've addressed all three points. Please re-review when you get a chance. Thanks!

code/go/internal/validator/test/packages/bad_deploy_variants/_dev/deploy/variants.yml
ycombinator: Added in https://github.com/elastic/package-spec/pull/45/commits/790a8e3c14befff47c5a5b620c0fb02434ee7d0e.

code/go/Makefile
```diff
@@ -23,4 +23,4 @@ check-spec:
 
 # Runs tests
 test:
-	@go test -v ./...
\ No newline at end of file
+	@go test -v ./... -count=1 # count is there to bust the test cache
```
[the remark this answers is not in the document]
ycombinator: Added in 73da55544ae34d4787594f2b318b6260d8647a09.

code/go/pkg/validator/validator_test.go
ycombinator: Added in 315cc073fd596651867f7c39ff0e4fa39d6c7262.

code/go/Makefile
```diff
@@ -23,4 +23,6 @@ check-spec:
 
 # Runs tests
 test:
-	@go test -v ./...
\ No newline at end of file
+    # -count=1 is included to bust the test cache. This way, if you run "make test" multiple times
+    # you will get fresh test results each time.
```
mtojek, line 27: nit: you can add here a note that changing source of mocked packages doesn't invalidate the cache.

mtojek (approved):
Ship it!

code/go/Makefile
ycombinator: Done in d6af314.

PR #48, opened 2020-09-18
Support two-character dataset names

mtojek:
This PR adds support for two-character dataset names.

Used for the first time: https://github.com/elastic/integrations/pull/245


ruflin (approved):
Nice to have a test for it too!

PR #49, opened 2020-09-18
Adjusting deploy spec

ycombinator:
This PR adjusts the spec for the `_dev/deploy` folder as per this proposal: https://github.com/elastic/integrations/pull/263#discussion_r490609023.

With this PR the spec will expect a specific folder name under `_dev/deploy` corresponding to a supported deployment model. So a package could have this:

```
_dev/
  deploy/
   variants.yml (optional)
    docker/
      docker-compose.yml    (required)
      Dockerfile   (optional)
      (other, optional files as needed via `additionalContents: true`)
```

or:

```
_dev/
  deploy/
    terraform/
      TBD
```

or:

```
_dev/
  deploy/
    SOME_NEW_MODEL/
      TBD
```

For now, only the Docker-based deployment model is supported. The spec changes in this PR reflect this. In the future when we add other deployment models, we'll need to adjust the spec accordingly, e.g. to expect that there must be exactly one sub-folder under `_dev/deploy` since a package should only specify one deployment model.


ycombinator:
I would expect CI to fail on this PR as it stands. Specifically, I'd expect this test case to fail as the mock package for that test case has a `_dev/deploy` folder but no `docker` folder under it:

https://github.com/elastic/package-spec/blob/d669841b7963c3779e410c9ce581cfb48a66ee97/code/go/pkg/validator/validator_test.go#L19

I wonder why CI is passing 🤔 . I'm going to investigate this before I push a commit to fix the expected test failure in this PR.

ycombinator:
Okay, turns out I had accidentally pushed up the test fix commits already, which is why CI was passing earlier. I temporarily removed them now and CI is [failing](https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Fpackage-spec/detail/PR-49/2/pipeline#step-158-log-34) as expected. Phew! 

So just PEBCAK, nothing to see here 🙈 .

I will push up the test fix commits again now.

mtojek (approved):

PR #52, opened 2020-09-30
Adding ES histogram field to the spec

jonathan-buttner:
Adding the histogram field as a valid field type.

https://www.elastic.co/guide/en/elasticsearch/reference/master/histogram.html


ycombinator (approved):
LGTM. 

We haven’t been doing a good job of updating the [package spec CHANGELOG](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml) but better to start late than never. Would you mind adding a CHANGELOG entry?

ruflin:
@jonathan-buttner Do you know if this change also made it into Kibana and is supported there too?

jonathan-buttner:
> @jonathan-buttner Do you know if this change also made it into Kibana and is supported there too?

hmm if I remember correctly I think we needed this so we could use the histogram field type in the endpoint package. Here's an example: https://github.com/elastic/endpoint-package/blob/main/custom_schemas/custom_endpoint.yml#L636

I don't remember if we needed to make changes in fleet 🤔

Issue #55, opened 2020-09-30
Manifest files: add support for corresponding flat and object properties


mtojek:
Due to different implementations, some properties may be considered as faulty:

```
kibana.version: 1
```

vs.

```
kibana:
   version: 1
```

Same for: `elasticsearch.index_template.mappings`

The goal of this task is to research all places that are required to be adjusted in elastic-package and in package-spec.

mtojek:
As the change is merged to package-spec I will adjust the elastic-package to support flat/object properties.

PR #54, opened 2020-10-05
Require three dashes (---) at the beginning of YAML pipelines

mtojek:
Closes https://github.com/elastic/package-spec/issues/53

PR #56, opened 2020-10-06
Support expanded and flat objects

mtojek:
Issue: https://github.com/elastic/package-spec/issues/55

For ease of merging, I branched out from this PR: https://github.com/elastic/package-spec/pull/54 (if that's an issue I can branch from master).


mtojek:
This is another place where `mapstr.go` is needed. I think we need to put it in a good place.

ycombinator:
> For ease of merging, I branched out from this PR: #54 (if that's an issue I can branch from master).

It depends on how much longer we're going to wait on #54, which depends on how much longer we want to wait for the associated ES issue to at least be discussed. 

I would prefer to branch this PR (#56) from `master` so there's no dependency on #54.

mtojek:
I was thinking the other way round. Let's consider #54 as workaround and once the discussion with ES is finalized we'll adjust the implementation, but I don't mind splitting these PRs.

ycombinator:
That plan also works for me. In that case don’t bother splitting out this PR. I will review #54 now so we can get it in shortly. Then you can just rebase this PR on `master`.

mtojek:
Ok, I rebased against master. It's ready for review.

code/go/internal/validator/test/packages/missing_pipeline_dashes/manifest.yml
```diff
@@ -5,7 +5,8 @@ description: This package is good.
 version: 0.1.2
 release: beta
 conditions:
-  kibana.version: '^7.9.0'
+  kibana:
+    version: '^7.9.0'
```
ycombinator, line 9: I assume the purpose of the new `expanded` test package is to test the case where the kibana version condition is specified like this:

```
kibana:
  version: ...
```

Given that, should we leave this line as-is (flattened syntax) to make sure both syntaxes are being tested?

mtojek: You're right. Fixed.

ycombinator (approved):
LGTM.

PR #59, opened 2020-10-13
Added two missing types: wildcard and version

webmat:
Hello, I've noticed that the types `wildcard` and `version` ([new in 7.10](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/version.html)) are missing from this spec file.


webmat:
@ycombinator First time here, sorry. Done now :-)

Issue #65, opened 2020-10-20
Validate that Kibana asset IDs start with lowercase package names


ycombinator:
For packages that install Kibana assets (saved searches, visualizations, dashboards, etc.) we should ensure that the asset IDs start with the package name, all lowercased. This will help avoid conflicts with the same assets being installed by corresponding Beats modules (for users not using Agent).

This issue will be resolved via multiple PRs (please keep the list below updated with links to the various PRs)
* [x] Validate that Kibana object filenames conform to the ID rules laid out above: #66 
* [ ] Validate that the `id` fields inside Kibana object files have the same value as the object filename (minus the `.json` extension, of course): https://github.com/elastic/package-spec/pull/160
* [ ] Validate that there are no dangling references to Kibana objects in any Kibana object files within the scope of a package. That is, if a Kibana object file in package `P` references another Kibana object with ID `i`, then there must exist a Kibana object file for id `i` in package `P`:

mtojek:
I'm going with this feature.

ycombinator:
In #66 we added validation for the Kibana object _filenames_. However, we still need two more validations that are related:
- to validate that the `id` fields _inside_ Kibana object files have the same value as the filename (minus the `.json` extension, of course), and
- to validate that there are no dangling references to Kibana objects in any Kibana object files within the scope of a package. That is, if a Kibana object file in package `P` references another Kibana object with ID `i`, then there must exist a Kibana object file for id `i` in package `P`.

I have adjusted the checklist in this issue's description to reflect this.

mtojek:
> to validate that the id fields inside Kibana object files have the same value as the filename they are defined in (minus the .json extension, of course), and

@ycombinator It might be hard to implement this just with the spec. It might be possible to reach the goal with FormatCheckers (similarly to https://github.com/elastic/package-spec/pull/83). 

> to validate that there are no dangling references to Kibana objects in any Kibana object files within the scope of a package. That is, if a Kibana object file in package P references another Kibana object with ID i, then there must exist a Kibana object file for id i in package P.

I wonder if it's not an edge case and this way we may introduce too much complexity in validation. I don't remember that dashboards refer to each other between packages. Are we sure we want to proceed with this requirement? It may happen that one dashboard will be adjusted (renamed), so tests for a referencing one will start failing. I would vote for excluding this requirement from the task scope.

ycombinator:
[...]

PR #66, opened 2020-10-21
Validate that dashboard IDs start with package names

mtojek:
Issue: https://github.com/elastic/package-spec/issues/65


versions/1/kibana/spec.yml
```diff
@@ -9,7 +9,7 @@
       - description: A dashboard asset file
         type: file
         contentMediaType: "application/json"
-        pattern: '.+\.json'
+        pattern: '{PACKAGE_NAME}+-.+\.json'
```
ycombinator, line 12: This is an interesting idea - to allow certain keywords in the spec that can act as variables/placeholders whose value depends on the context. TBH, this is not how I had imagined implementing semantic validations but I like it - its simple! Let's go with this idea and see how far we can take it.

We should probably document these keywords somewhere (outside the spec, of course) for spec developers so they know which keywords are available to them and what they mean.

mtojek: Do you think about README or CONTRIBUTING file in package-spec or somewhere else?

code/go/internal/validator/folder_spec.go
```diff
@@ -168,13 +169,13 @@ func (s *folderSpec) validate(folderPath string) ValidationErrors {
...
 
-func (s *folderSpec) findItemSpec(folderItemName string) (*folderItemSpec, error) {
+func (s *folderSpec) findItemSpec(packageName string, folderItemName string) (*folderItemSpec, error) {
 	for _, itemSpec := range s.Contents {
 		if itemSpec.Name != "" && itemSpec.Name == folderItemName {
 			return &itemSpec, nil
 		}
 		if itemSpec.Pattern != "" {
-			isMatch, err := regexp.MatchString(itemSpec.Pattern, folderItemName)
+			isMatch, err := regexp.MatchString(strings.ReplaceAll(itemSpec.Pattern, "{PACKAGE_NAME}", packageName), folderItemName)
```
ycombinator, line 178: Validating that the dashboard file name starts with the package is a good start. But we also need to implement two more related validations:
- validating that the ID of the dashboard, as defined in the dashboard asset JSON file, starts with the package name or, better yet, matches the file name (sans extension) of the dashboard asset JSON file, and
- validating that there aren't any references to the dashboard ID with the package name prefix removed.

For both these additional validations, I think you'll need to define the spec files for Kibana assets. I was hoping to avoid this but I think we have to bite that bullet now. Perhaps we could define them with only the fields we need and leave the rest open-ended (that is, allow additional fields in the asset file that are not explicitly defined in the spec)?

If you want to tackle these additional validations in a separate PR that's fine by me. In that case, just update the issue corresponding to this PR with a checklist of the 3 total validations so we can keep track of the work done vs. remaining.

code/go/internal/validator/folder_spec.go
```diff
@@ -168,13 +169,13 @@ func (s *folderSpec) validate(folderPath string) ValidationErrors {
 	return errs
 }
 
-func (s *folderSpec) findItemSpec(folderItemName string) (*folderItemSpec, error) {
+func (s *folderSpec) findItemSpec(packageName string, folderItemName string) (*folderItemSpec, error) {
 	for _, itemSpec := range s.Contents {
 		if itemSpec.Name != "" && itemSpec.Name == folderItemName {
```
ycombinator, line 174: Should we allow the use of keyword placeholders, e.g. `{PACKAGE}`, in item names as well (not just item patterns)?

versions/1/kibana/spec.yml
ycombinator: I think CONTRIBUTING is the right place for this sort of thing. Let's try to keep the README more for spec _consumers_ and CONTRIBUTING for spec _contributors_, although there might be some overlap/fuzziness, which is okay IMHO.

code/go/internal/validator/folder_spec.go
mtojek: I suspected that you may ask about it. I decided to postpone it until we've a real use-case. So far, this is the only binding between items, so I vote for leaving it only for patterns.

ycombinator: Sounds good to me!

code/go/internal/validator/folder_spec.go
mtojek: I agree with the first point (validating the ID of the dashboard inside the file). I can introduce the schema for dashboard file. We might need to introduce the binding between ID and packageName also there.

Regarding the second point, it's hard to implement it as it's mainly "search and replace" operation and we may fail the validation if there are references between different dashboards (e.g. jump to the system dashboard).

ycombinator: Let's break these out into follow up PRs. Implementing the first point will introduce more files so let's not bloat this PR with those. Implementing the second point might be tricky as you pointed out so let's keep that out of scope for this PR. What you have in this PR here is already useful so let's move that forward first.

versions/1/kibana/spec.yml
mtojek: I added a short one.

mtojek:
I'll merge this PR as it's already an improvement. We can work on follow up PRs then.

ruflin:
Is this only about dashboards or also visualizations, searchs etc?

mtojek:
It's only for dashboards (according to the email thread).

ruflin:
I would assume we also have conflicts on visualizations, search etc?

code/go/internal/validator/folder_spec.go
exekias: > Regarding the second point, it's hard to implement it as it's mainly "search and replace" operation and we may fail the validation if there are references between dashboards in different packages (e.g. jump to the system dashboard).

Visualization used by the dashboard should all be contained within the same integration (we cannot include a visualization from a different integration, as it may not be installed). Are you thinking about links here? Most of what we have is just visualizations, and their IDs should match this pattern.

++ on following up with these cases

exekias (approved):
Thank you for jumping into this so quick!

code/go/internal/validator/folder_spec.go
exekias: Perhaps let's not close #65 until the other PRs get in?

mtojek: Good point, Carlos. I updated the description.

mtojek:
> I would assume we also have conflicts on visualizations, search etc?

I tried to reproduce it and we might need to update ID references one more time, @ycombinator . How to reproduce it:

```
$ elastic-package stack up -d
$ metricbeat setup --dashboards
```

Then navigate to the system dashboard which may be scrambled.

I'm not quite sure about the previous fix https://github.com/elastic/integrations/pull/171 , which only covered dashboards. In the worst case we need to wrap up a similar fix to https://github.com/elastic/integrations/issues/319 for visualization and searches, and the same way iterate on other PRs related to it.

ycombinator:
Thanks @ruflin for pointing out that this validation applies to other Kibana assets as well.

@mtojek I have restructured https://github.com/elastic/package-spec/issues/65 to be more of a meta issue, covering the various types of Kibana assets that need this type of ID validation + making room for multiple, follow-up PRs.

ycombinator:
> I'm not quite sure about the previous fix elastic/integrations#171 , which only covered dashboards. In the worst case we need to wrap up a similar fix to elastic/integrations#319 for visualization and searches

Yes, I will work on these PRs today.

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Using predefined placeholders in filename patterns
+
+There are predefined placeholders that can be used in filename patterns. Standard patterns are simple regular expressions
+like `[a-z0-9]+\.json`. In some cases it might be useful to introduce a harder requirement (binding), e.g. a filename should
+be prefixed with a package name (`{PACKAGE_NAME}-.+\.json`).
+
+Currently, the following placeholders are available:
+
+* `{PACKAGE_NAME}` - name of the package
```
ycombinator, line 13: ❤️

ycombinator (approved):
LGTM.

ycombinator:
>> I'm not quite sure about the previous fix elastic/integrations#171 , which only covered dashboards. In the worst case we need to wrap up a similar fix to elastic/integrations#319 for visualization and searches

> Yes, I will work on a PR today to the @elastic/integrations repo for fixing any visualization and saved search IDs.

I started to create this PR but ran into an issue that needs some discussion. Please take a look: https://github.com/elastic/integrations/issues/334#issuecomment-716271597 and continue the discussion on that issue (not here). Thanks!

mtojek:
> @mtojek I have restructured #65 to be more of a meta issue, covering the various types of Kibana assets that need this type of ID validation + making room for multiple, follow-up PRs.

Thank you for adjusting the issue. I wonder if we shouldn't include all these changes in a single PR. Otherwise we'll end up in few iterations on adjusting integrations and spec. Or I will wait with merging this one, until all packages are in a good shape in terms of dashboard names (to merge elastic-package dependency update).

> @mtojek can you work on any corresponding changes to the migration script?

Yes, that's on my TODO list.

PR #73, opened 2020-10-30
Add `metric_type` and `unit` to fields.yml spec

### LINKED ISSUE #42: Add support for _meta in fields.yml
In https://github.com/elastic/elasticsearch/pull/61941 a standardised way on how to use `_meta` information on fields is documented. On the spec side these standard _meta entries should be supported and validated.

exekias:
`metric_type` is used to store the type of metric for numeric fields
storing time series metrics.

`unit` is used to store the units of any numeric field.

These two fields will be stored under the `_meta` object of the
template.

Closes #42


exekias:
Tentative implementation of https://github.com/elastic/package-spec/issues/42, please don't merge yet, as the discussion is not closed

ruflin (approved):
LGTM

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -52,6 +52,24 @@ spec:
...
+        description: Metric type
+        type: string
+        enum:
+        - counter
+        - gauge
+      unit:
+        description: Unit
+        type: string
+        enum:
+        - byte
```
ycombinator, line 65: I know `byte` is here because ES supports it but it's the same as `b`, right? To avoid confusion should we remove one of them from the allowed values?

ycombinator:
Could you please add an entry to the package spec CHANGELOG: https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml? We haven't done a good job of keeping it updated so it would be nice to use it more.

versions/1/data_stream/fields/fields.spec.yml
exekias: arg actually I copied over the byte units instead of time units 🤦 will fix it soon. There was no consensus around using byte units here, so we stick to just `byte` for now

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -52,6 +52,24 @@ spec:
...
+        - percent
+        - b
+        - kb
+        - mb
+        - gb
+        - tb
+        - pb
```
exekias, line 72: These should be time-units

versions/1/data_stream/fields/fields.spec.yml
ycombinator: Ha, I didn't even notice that 😊 .

exekias:
Thanks for the reviews! this should be ready for another look

ycombinator (approved):
LGTM. WFG.

Issue #71, opened 2020-10-27
Validation: Kibana asset IDs should not contain `-ecs` suffix


ycombinator:
Per the discussion starting from https://github.com/elastic/integrations/issues/269#issuecomment-693368174, it was agreed that Kibana asset IDs in Integrations packages should not contain the `-ecs` suffix. Also, any asset titles/labels/markdown links/etc. should not contain the ` ECS` suffix either (per @andrewkroh's suggestion here: https://github.com/elastic/integrations/issues/339#issuecomment-717238777). We should add validation to the spec for this.

This issue will be resolved via multiple PRs (please keep the list below updated with links to the various PRs)
* [x] Validate that Kibana object filenames do not contain the `-ecs` suffix: 
* [ ] Validate that titles/labels/markdown links/etc. within Kibana object files should not contain the ` ECS` suffix:

ycombinator:
Also validate that the title of the assets doesn't contain an `ECS` suffix, per @andrewkroh's suggestion here: https://github.com/elastic/integrations/issues/339#issuecomment-717238777.

jsoriano:
This doesn't seem to be a problem anymore, the import tool removed these suffixes, and new integrations shouldn't have it.

Looking to the integrations repository, I have only found the `awsfargate` integration to contain dashboards with titles with the ECS prefix. So probably it would be enough with fixing these dashboards, and it is not worth to add an explicit validation for this.

@elastic/obs-cloud-monitoring could you please remove `ECS` from the titles of the dashboards in the `awsfargate` package? Thanks!

kaiyan-sheng:
@jsoriano Sorry I don't see `-ecs` suffix in dashboard/lens/visualization file names. Are we also checking `ECS` in like title names? I think `ECS` here in awsfargate dashboard might represent Amazon Elastic Compute Cloud instead.

zmoog:
Yep! The ECS in the AWS Fargate integration stands for (Amazon) Elastic Container Service. 

Do you think we should remove it?

jsoriano:
Ah ok, sorry, this is perfectly fine then! I think we can close this. Thanks!

zmoog:
We could quickly expand the acronym in some places, but in others, it can look weird because of space constraints and would not match the names AWS uses on their web console.

Here are a couple of examples from titles in lens and visualizations:

[...]

PR #74, opened 2020-10-30
Verify: no ECS suffix for Kibana object filenames

mtojek:
Issue: https://github.com/elastic/package-spec/issues/71

Validate Kibana object filenames in terms of forbidden ECS suffix.


code/go/internal/validator/folder_item_spec.go
```diff
@@ -14,15 +14,16 @@ import (
...
-	Name             string `yaml:"name"`
-	Pattern          string `yaml:"pattern"`
-	Required         bool   `yaml:"required"`
-	Ref              string `yaml:"$ref"`
-	Visibility       string `yaml:"visibility" default:"public"`
-	commonSpec       `yaml:",inline"`
+	Description       string   `yaml:"description"`
+	ItemType          string   `yaml:"type"`
+	ContentMediaType  string   `yaml:"contentMediaType"`
+	ForbiddenPatterns []string `yaml:"forbiddenPatterns"`
```
ycombinator, line 20: If we can _somehow_ make the regular expression work (per our conversation off-PR), we wouldn't need this field, right?

For now, could you move it to below the `Pattern` field in the struct since they are related.

code/go/internal/validator/test/packages/wrong_kibana_filename/manifest.yml
```diff
@@ -1,7 +1,7 @@
 format_version: 1.0.4
-name: good
-title: GOOD
-description: This package is good.
+name: bad
+title: BAD
+description: This package is bad.
```
ycombinator, line 4: Thanks for fixing this, I missed it in an earlier review, sorry!

code/go/pkg/validator/validator_test.go
```diff
@@ -59,19 +59,21 @@ func TestValidateItemNotAllowed(t *testing.T) {
...
 			},
 			"kibana/search": []string{
 				"691240b5-7ec9-4fd7-8750-4ef97944f960.json",
-				"bad-search.json",
+				"missed-search.json",
 			},
 			"kibana/visualization": []string{
 				"defa1bcc-1ab6-4069-adec-8c997b069a5e.json",
-				"bad-visualization.json",
+				"missed-visualization.json",
```
ycombinator, line 76: Why `missed-`?

versions/1/kibana/spec.yml
```diff
@@ -10,6 +10,8 @@
         type: file
         contentMediaType: "application/json"
         pattern: '{PACKAGE_NAME}-.+\.json'
+        forbiddenPatterns:
+          - '.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
```
ycombinator, line 14: ❤️  the comments here.

code/go/internal/validator/folder_item_spec.go
mtojek: Yes, we wouldn't need it.

Apart from that, it's MUCH simpler and easier to read, having two independent patterns. See: https://github.com/elastic/elastic-package/pull/154#discussion_r515173374

code/go/pkg/validator/validator_test.go
mtojek: I renamed to `not-bad` as it's different to `bad`.

Then I discoved another issue with regular expressions - the spec doesn't care about `^$`, so I had to consider them in the validator. It failed datastream regexp, so I adjusted it.

code/go/pkg/validator/validator_test.go
```diff
@@ -59,19 +59,21 @@ func TestValidateItemNotAllowed(t *testing.T) {
...
 				"b7e55b73-97cc-44fd-8555-d01b7e13e70d.json",
-				"bad-dashboard.json",
+				"bad-ecs.json",
+				"bad-foobar-ecs.json",
+				"bad-Foobaz-ECS.json",
 			},
 			"kibana/map": []string{
 				"06149856-cbc1-4988-a93a-815915c4408e.json",
-				"bad-map.json",
+				"not-bad-map.json",
```
ycombinator, line 68: Sorry, I'm a little confused about this name change. What was the problem if you left it as `bad-map.json`?

mtojek: If you take a look at `test/package`, the name `good` is restricted to the package which is 100% aligned to the schema.   This package here is not good - `bad` :) which means that `bad-map.json` is a valid Kibana file, but I need a one with wrong name here.

ycombinator: Ah, got it, thanks for explaining. `not-bad` is quite a twisted way of saying `good` so how about we just give these files UUIDs as names? Anyway I don't see any difference between this file and the one above it (`06149856-cbc1-4988-a93a-815915c4408e.json`) - they are both files with empty JSON objects in them and neither of them have a `-ecs` suffix in the name. Or am I missing something again? 😄

versions/1/data_stream/spec.yml
```diff
@@ -3,7 +3,7 @@ spec:
   contents:
   - description: Folder containing a single data stream definition
     type: folder
-    pattern: '[a-z0-9]([a-z0-9_]+[a-z0-9])?'
+    pattern: '([a-z0-9][a-z0-9]|[a-z0-9][a-z0-9_][a-z0-9_]+)'
```
ycombinator, line 6: Why was this change necessary?

code/go/pkg/validator/validator_test.go
mtojek: The intention was:
* `06149856-cbc1-4988-a93a-815915c4408e.json` is a wrong file without package name prefixed.
* `not-bad-map.json` is a file with a different package name.

I suppose I can fallback to the first option only (keep the `06149856-cbc1-4988-a93a-815915c4408e.json` file only).

versions/1/data_stream/spec.yml
mtojek: `pe` datastream was not valid.

mtojek: BTW I changed it now to `'([a-z0-9][a-z0-9]|[a-z0-9][a-z0-9_]+[a-z0-9])'

code/go/pkg/validator/validator_test.go
ycombinator: I see. Thanks for clarifying. I think the issue is that it's not obvious from the filenames what purpose they are meant to serve in the tests. I would suggest keeping two files, one to test each case that you explained. But:
- maybe instead of calling it `not-bad-map`, call it something like `another-package-map`, only because `not-bad` is confusing ("why didn't they just call it `good`", someone looking at this file might ask), and
- maybe flesh out each of the files just enough to get a description in there, where we can see what purpose that file is serving in tests.

versions/1/data_stream/spec.yml
ycombinator: Understood. BTW, you can change the first option in the regex (the left side of the `|`) to `[a-z0-9]{2}`.

code/go/pkg/validator/validator_test.go
mtojek: > maybe instead of calling it not-bad-map, call it something like another-package-map, only because not-bad is confusing ("why didn't they just call it good", someone looking at this file might ask), and

`another-package-map.json` LGTM , I will adjust the code.

> maybe flesh out each of the files just enough to get a description in there, where we can see what purpose that file is serving in tests.

I can add two properties, ID and description, to add more context to the test case. I wouldn't pollute it with a real Kibana object with all properties as I believe it's not necessary them.

ycombinator: > I can add two properties, ID and description, to add more context to the test case. I wouldn't pollute it with a real Kibana object with all properties as I believe it's not necessary them.

👍  I think it might be good to follow a real Kibana object's schema though. That way if we ever start validating Kibana object structures, we don't suddenly start failing this test. But ++ to only populating the minimum number of properties needed.

versions/1/data_stream/spec.yml
mtojek: Nice, I simplified the regular expression.

code/go/pkg/validator/validator_test.go
mtojek: Done

ycombinator (approved):
LGTM. Thanks for the clarifying changes.

PR #76, opened 2020-11-02
Fix: invalid fields file regex

mtojek:
This PR fixes another issue in fields file (invalid filename).

Spotted in https://github.com/elastic/elastic-package/pull/154 .


ycombinator (commented):
Nice catch! But why is this not failing CI in the integrations repo already? 🤔  There are a number of field definition files there that are using `-` in their names and the bad spec has been around for ~3 months: https://github.com/elastic/package-spec/blame/master/versions/1/data_stream/fields/spec.yml#L6. It would be _really_ good to understand why Integrations CI isn't failing all over the place because something is obviously being missed. 

Probably the easiest place to start would be to keep the new test file you added in this PR but temporarily revert the `pattern` change. That _should_ cause this PR to fail in CI. If it passes, then we know something is wrong in the `package-spec` validation implementation itself. If it fails, that's good but then something's wrong in the Integrations repo CI.

mtojek:
> Nice catch! But why is this not failing CI in the integrations repo already? 🤔 There are a number of field definition files there that are using - in their names and the bad spec has been around for ~3 months: https://github.com/elastic/package-spec/blame/master/versions/1/data_stream/fields/spec.yml#L6. It would be really good to understand why Integrations CI isn't failing all over the place because something is obviously being missed.

It was related to a bug I fixed in the previous PR - constructing regular expressions without `^ $`.

ycombinator:
> It was related to a bug I fixed in the previous PR - constructing regular expressions without `^ $`.

IIUC, then shouldn't we add these anchors to all our filename patterns?

I'm still curious what would happen if you did this:

> Probably the easiest place to start would be to keep the new test file you added in this PR but temporarily revert the `pattern` change.

mtojek:
> IIUC, then shouldn't we add these anchors to all our filename patterns?

They are added in this line:

https://github.com/elastic/package-spec/blob/master/code/go/internal/validator/folder_spec.go#L185

mtojek:
I reverted the pattern change:
https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Fpackage-spec/detail/PR-76/2/pipeline

ycombinator:
> They are added in this line:

> https://github.com/elastic/package-spec/blob/master/code/go/internal/validator/folder_spec.go#L185

I see. I'm trying to decide if this is good or bad.
* It's good because we don't have to add them to each pattern. 
* It's bad because:
   * there's now a bit of "magic" being applied to every folder item `pattern` (but not to `pattern`s in other areas of the spec), and
   * if we ever need to define a `pattern` for a folder item that shouldn't be anchored, we would end up adding `.+` wildcards on the side(s) that shouldn't be anchored just to account for the anchors being added in the code.

So, on the whole, I'm leaning towards removing the anchors "magic" from code and instead defining them explicitly everywhere we need them in folder item `pattern`s. But I'm curious to hear your thoughts on this too. WDYT?

ycombinator:
> I reverted the pattern change:
https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Fpackage-spec/detail/PR-76/2/pipeline

Thanks. The build fails, which is what we would expect. Phew! 😅

mtojek:
It's good reason to add `^ $` to patterns because of next possible validator implemetations (e.g. TypeScript). We should keep as least logic as possible in the source code.

ycombinator:
Alright, @mtojek, let's fix up this PR to assume that anchors are being added in the code so we can get it merged. Then let's make a follow up PR to remove the anchors logic from the code and fix up all the places in the spec we are defining folder item `pattern`s.

ycombinator:
Looks like you made the changes to remove the anchors from code in this PR itself. So ignore my previous comment 🙂 .

mtojek:
Yes, I've already adjusted it and rebased. Let's wait for the CI.

versions/1/data_stream/spec.yml
```diff
@@ -3,7 +3,7 @@ spec:
   contents:
   - description: Folder containing a single data stream definition
     type: folder
-    pattern: '([a-z0-9]{2}|[a-z0-9][a-z0-9_]+[a-z0-9])'
+    pattern: '^([a-z0-9]{2}|[a-z0-9][a-z0-9_]+[a-z0-9])$'
```
ycombinator, line 6: Just want to double check this pattern here. It says we allow underscores in data stream folder names, but we don't allow hyphens. Just want to make sure this is what we want.

mtojek: Yes, it's requirement enforced by Kibana, which uses hyphens to build indices.

ycombinator (approved):
LGTM.

PR #79, opened 2020-11-03
Add support for top level configuration

### LINKED ISSUE #70: Do not ignore vars when there are no inputs
Consider this snippet of an `0.1.0/manifest.yml` example:

```
policy_templates:
  - name: apm
    inputs:
      - type: traces
        vars:
          - name: host
            type: int
            title: APM Server port
            required: true
            show_user: true
            default: 8200
```

If my data stream(s) don't have a `streams` key, the generated policy will just have an `input` with  `streams: []`, and the variable `host` will not be there even if I had filled it in the UI.

For apm-server we don't have streams, so we need that the `vars` defined at the policy level get picked up and rendered in the generated policy.

Not sure on which repo this issue belongs, let me know if it should be transferred to Kibana. 

**Update**: The ideal way forward is that the top level config shows right under `inputs`, and not under a `streams` entry. For this we need hbs templates at the top level too.

jalvz:
Closes #70


ruflin (commented):
Overall LGTM. How will the files inside look exactly? Can there be more than 1 file? What is the content? Would be good to also directly specify this here. If not in the spec, in the PR description.

versions/1/spec.yml
```diff
@@ -28,6 +28,17 @@ spec:
...
+    type: folder
+    name: agent
+    required: false
+    additionalContents: false
+    contents:
+    - description: Folder containing input definitions
+      type: folder
+      name: input
+      required: true
+      additionalContents: false
```
ycombinator, line 41: Related to @ruflin's comment, it would be good to flesh out `contents` here with a spec for the file(s) that could be be contained in this folder.

jalvz:
Thanks both, good idea. Added now, let me know what do you think

versions/1/agent/spec.yml
```diff
@@ -0,0 +1,13 @@
+spec:
+  additionalContents: false
+  contents:
+  - description: Folder containing input definitions
+    type: folder
+    name: input
+    required: true
+    additionalContents: false
+    contents:
+    - description: Package-level template file
```
ruflin, line 10: Probably a better description needed here.

The part I was also looking for is to define, if the content of the file itself is an array or not. Also, are there some required like type? Let me give you 3 examples:

```
inputs:
  - type: foo
    value: bar
```

```
- type: foo
  value: bar
```

```
type: foo
value: bar
```

Which option is it?

@ycombinator Do we have examples already where we also validate the content of certain assets?

jalvz: What I am trying spec out is the optional existence of `/<package>/<version>/agent/input/template.yml.hbs`. I was expecting to adhere to the same constraints than data streams templates: why a `type` or any other field must be defined inside? Is not up to the integration to make sense of whatever is in there?

There is nothing defined for `data_stream/<name>/agent/stream/` contents that I can see either (even thou the default file name is `stream.yml.hbs`)

ruflin: On the stream.yml.hbs files the rules is, it can't be an array and alway only contain a single input. Is this the same for input templates? Is the format the last option from above?

jalvz: Ah, I wasn't aware of that. Yes, it would be the same then, last option.

ruflin: @jen-huang How will this work on the Kibana side. How will Kibana know which file to use? Convention? Or do we need to reference the file in the input part in the manifest.yml? As inputs is an array there, could there be multiple files in this directory?

jen-huang: @ruflin @jalvz Kibana will need a `template_path: string;` field on each input (same as `template_path` on each stream) to know which template from the directory to use. this will give flexibility so that different inputs can use different templates too.

jalvz: thanks @jen-huang , updated the spec to add the `template_path`.

jalvz:
@jen-huang @ycombinator @ruflin does this look right now? something else missing?

versions/1/agent/spec.yml
ycombinator: > Do we have examples already where we also validate the content of certain assets?

@ruflin Do you mean something like this? https://github.com/elastic/package-spec/blob/master/versions/1/data_stream/fields/fields.spec.yml

ruflin: @ycombinator Yes! The way I think of it is that we define what are the required fields and everything else is free form.

ruflin:
What we miss is the definition of the base format of the template file, see conversation above with @ycombinator . But maybe it is best to move forward even without it, so we can still change it (see comment below).

@ycombinator We also have a bit a chicken / egg problem here. We are adding something new to the spec but haven't fully tested it with Kibana and an actual package yet, so it might still change. What is our best approach here?

@jen-huang @ycombinator If you are fine with the changes, lets get them in.

ycombinator:
Ideally the spec will be defined first and it's changes rolled out into `elastic-package` and the `integrations` repo. This ensures that all existing and new packages will conform to the spec. Then we make the corresponding functionality changes in Kibana, etc. 

Also ideally the spec can be as strict as possible so as to catch as many issues as possible early on in a package's development. 

Given that, I would suggest for this PR:
- to spec out the base definition of the template file (per https://github.com/elastic/package-spec/pull/79#discussion_r521922711)
- get @jen-huang's LGTM on this PR from a "do we think this will work for Kibana" perspective. 
- then we merge it, create/modify a package with the changes, and try it out in Kibana. If something needs to be tweaked, we make another PR to the spec and rinse/repeat. But hopefully this step can be avoided as much as possible.

versions/1/agent/spec.yml
jalvz: Gotcha, thanks.
Changed my mind about the format, and now I am requiring a single, top-level `config` field which might hold any arbitrary data.
This is to avoid clashes when merging the template with the input fields, ie. imagine if a user has a `name` field in their agent/input template.

ruflin: Could you share an example? I thing I'm stumbling over the term `config` here. 

What would be the conflict around `name`?

jalvz: I don't have a preference about the actual term, but this is the concern:

`manifest.yml`
```yml
policy_templates:
  - name: apm-server
    inputs:
      - type: apm
        vars:
          - name: name
            default: my-default-name
        template_path: ./agent/input/template.yml.hbs
```

`agent/input/template.yml.hbs`
```yml
name: {{name}}
```

`elastic-agent.yml`
```yml
inputs:
  - id: 0e682c50-183a-11eb-916d-71d55143d422
    name: ??????
    revision: 1
    type: apm
```

What will be the value of the `??????` field? `my-default-name`, as per the template variable, or whatever the user set in UI when creating the integration?

Instead, forcing a (eg.) `config` key:

`elastic-agent-2.yml`
```yml
inputs:
  - id: 0e682c50-183a-11eb-916d-71d55143d422
    name: apm-1 # user defined in UI
    revision: 1
    type: apm
    config:
       # everything from the template 
       name: my-default-name
```

Makes sense?

jen-huang: @jalvz I think the problem you describe is something that could be run into today with streams & stream templates. We generate streams in `elastic-agent.yml` like this:
```yml
    streams:
      - id: logfile-system.auth
        data_stream:
          dataset: system.auth
          type: logs
        paths:
          - /var/log/auth.log*
          - /var/log/secure*
        ...
```

If the package author has `id` or `data_stream.*` fields in their stream template, we would run into the conflict problem you described.

For inputs, we have more of these kind of "reserved" generated field names:

```yml
inputs:
  - id: b8bc5300-edfd-11ea-905a-819b5c00fe02
    name: system-1
    revision: 1
    type: logfile
    use_output: default
    meta:
      package:
        name: system
        version: 0.5.3
    data_stream:
      namespace: default
    streams:
      ...
```

I'm not sure if adding a `config` field at the input level is the right approach though, given same nature of the problem on the stream level. I guess for streams, we've relied on package authors not using `id` and `data_stream.*` in their templates. I wonder if this is something we can enforce via a blocklist during package validation? I recall we added package validation to enforce template syntax correctness, maybe a blocklist can be added there?

@jalvz @ruflin WDYT?

jalvz: I see @jen-huang, thanks.

In the spirit of developer-friendliness, I think we should prevent clashes from happening - for someone outside the ingest management team this problem is not obvious at all (in fact, I didn't realize it can happen with streams already) and consequences are pretty much unknown?

The issue with blocklists is to remember to update them when new fields are added, and more importantly how to make them work backwards: what if a new field `foo` is added to the spec but I already have it in a template?

I think this deserves a discussion before moving the spec to GA, maybe considering a breaking change. I agree with @ycombinator that the more strict the better, and currently there is no definition at all for the stream templates AFAICS.

Is there any other problem with requiring a single top level  key in the templates (named `config` or whatever) other than it is not done for streams? Alternatively, Kibana could "inject" such key behind the scenes, based eg. on the template file name or something like that. But it wouldn't be so explicit.

WDYT?

ruflin: We definitively need the validation that these keys are not used and if a package uses them, it should be reject. We have a spec versioning, so if we add new fields as "reserved" we will increase the package spec and the new package will follow the new spec. I doubt that at the moment we have anything around that could not be fixed without a breaking change, but we should check. If we validate, we will find it.

Assuming we don't have any conflicts, what is the best format we should use?

jalvz: Sure, it will be detected but it will likely force the package maintainer to change the template and deal with 2 variants of the same configuration. To that end the best preventive action would be to come up with keys that will never ever crash (assuming you know a clash in a future version can happen), so why don't be explicit and require it upfront?
 
OTOH, json-schema is already meant for validation, so adding more validation alongisde means that the json-schema spec becomes less reliable (a developer might be easily confused after carefully following the spec and then find out that their package doesn't work). 

Anyways, since that is a separate problem I filed https://github.com/elastic/package-spec/issues/85 and removed the `config` bit from this PR, let me know how it looks now.

jen-huang (approved):
LGTM for Kibana support

ruflin:
@jalvz Any chance you could post here the final content / structure of the template file? I think we are aligned but want to make sure we have it also written in YAML here.

Seems there is a conflict with the generated file.

mtojek (changes requested):
Correct me if I'm wrong, but I understand that this change is backward compatible (it's just an extension). We should be safe with adding it.

Please add a new folder with a sample package for testing purposes in:
https://github.com/elastic/package-spec/tree/master/code/go/internal/validator/test/packages

and enable it in this file: https://github.com/elastic/package-spec/blob/master/code/go/pkg/validator/validator_test.go

jalvz:
23e436d8d484ee9adfad5e9d85147b2899637d2c

versions/1/agent/agent.spec.yml
```diff
@@ -0,0 +1,7 @@
+##
+## Describes the specification for an input template file
+##
+spec:
+  # Everything under here follows JSON schema (https://json-schema.org/), written as YAML for readability
+  type: object
+  additionalProperties: true
```
mtojek, line 7: I understand that this schema file describes the `*.yml.hbs` file, which is Handlebars template. I think it's not possible to easily define a JSON schema for this file as the JSON format can be broken by template placeholders.

versions/1/agent/spec.yml
```diff
@@ -0,0 +1,14 @@
...
+    type: folder
+    name: input
+    required: true
+    additionalContents: false
+    contents:
+    - description: Config template file for inputs defined in the policy_templates section of the top level manifest
+      type: file
+      pattern: '^.+.yml.hbs$'
+      required: true
+      $ref: "./agent.spec.yml"
```
mtojek, line 14: As stated above, the `.yml.hbs` file is not a strict JSON/YAML file if you plan to use placeholders. You can't use (reference) a schema for this.

mtojek (changes requested):
I'm afraid that the CI still fails for this PR.

versions/1/agent/spec.yml
jalvz: right, thanks

mtojek (approved):
Ship it!

code/go/internal/validator/test/packages/input_template/agent/input/template.yml.hbs
```diff
@@ -0,0 +1 @@
+{}
```
mtojek, line 1: nit: It would be nice to put a real content in at least one file, so this `template.yml.hbs` won't be so mysterious.

jalvz:
> @jalvz Any chance you could post here the final content / structure of the template file? 

76738747c9a03a48a926dbacf655f359162b2557

Merging now, thanks all

code/go/internal/validator/test/packages/input_template/agent/input/template.yml.hbs
ruflin: @jalvz Can you open a PR with this? Because it is exactly what I'm looking for to have an example.

@mtojek These test packages are great. Example and testing in one go.

ruflin: BTW, you could even put a short APM Example in here to make it more concrete.

jalvz: @ruflin I added https://github.com/elastic/package-spec/commit/76738747c9a03a48a926dbacf655f359162b2557, is not that what you were looking for?
I don't think I can use this file as a real hbs example because it is referenced in the test (correct me if Im wrong)

Also didn't want to add an APM-like config because this can be anything, it has exactly the same structure as the stream templates.

ruflin: Got it. I missed that you added two template files. The part I stumble over is that you used `group`. This is just an example I assume. It can all be on the top level like `foo: bar`?

If we could use apm example here, @mtojek will know best. My preference is always to have a real example if possible.

jen-huang:
Hi all, I opened https://github.com/elastic/package-registry/pull/655 to allow the package registry to serve the new `template_path` field on input level.

PR #80, opened 2020-11-04
Add pattern property to field

mtojek:
This PR adds new field property to the field definition. The "pattern" property defines field value and will used for type assertion.

Issue: https://github.com/elastic/elastic-package/issues/113


versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -71,6 +71,9 @@ spec:
         - ms
         - micros
         - nanos
+      pattern:
+        description: Field value pattern
```
ycombinator, line 75: Maybe say that it's a regular expression pattern?

Also consider adding an `example`, like https://github.com/elastic/package-spec/blob/33b54fb5403b6cc36449dcba1825b5ed7d02a3e8/versions/1/data_stream/manifest.spec.yml#L19-L20.

mtojek: Rephrased description and added one example.

ycombinator (approved):
LGTM.

PR #82, opened 2020-11-10
Add ignored_fields to test config

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/169

This PR adds `ignored_fields` property definition to the test config.


code/go/internal/validator/test/packages/good/data_stream/foo/_dev/test/pipeline/test-access-raw.log-config.json
```diff
@@ -10,5 +10,10 @@
         "event.category": [
             "web"
         ]
-    }
+    },
+    "ignored_fields": [
+        "foobar",
+        "world",
+        "hello"
+    ]
```
ycombinator, line 18: I read through the use case for this PR: elastic/elastic-package#169. My understanding is that there will be some fields in the indexed documents that will have dynamic values - values whose value will be different each time the test is executed. However, these fields' values should still conform to a specific data type and shape. 

So I would not recommend that we completely allow certain fields to be ignored. Instead, perhaps we can allow users to specify certain fields in the test config and specify a valid pattern for that field's value. So something like:

```suggestion
    "dynamic_fields": {
      "foobar": ".*",
      "a_numeric_field": "\\d+"
    }
```

I'm not married to the name of this setting; just using `dynamic_fields` as a suggestion above.

This way we can allow for dynamic values for specific fields without completely ignoring them. WDYT?

mtojek: > However, these fields' values should still conform to a specific data type and shape.

This is a good argument for dynamic fields.

> Instead, perhaps we can allow users to specify certain fields in the test config and specify a valid pattern for that field's value. 

We'll need to update the pipeline test runner to make the diff more smarter (consider regular expressions), but I think it's possible.

ycombinator (approved):
LGTM.

PR #83, opened 2020-11-16
Validate JSON properties with relative-path format

mtojek:
We decided to proceed with mutex and FormatCheckers. I've adjusted the PR. The solution should thread-safe (but uses locking).

---

~_I'm opening this as draft to discuss possible design ideas._~

Resolves: https://github.com/elastic/package-spec/issues/31

I found `FormatCheckers` in `gojsonschema` that could be used to verify properties with relative paths (screenshot and image src). Unfortunately the `FormatCheckers` is a global object, not thread-safe and can be applied only to statics.

For reference:
https://github.com/xeipuuv/gojsonschema/issues/257
https://github.com/xeipuuv/gojsonschema/pull/300

I'm wondering whether we should guard the piece of code with `Validate()` and `refreshFormatCheckersContext` with a mutex or should we use a different approach:

1. Package domain-aware validation (screenshots and images have sources). So far we didn't need to strongly depend on package domain and it would be nice not to follow this path
2. Fork the package and modify the existing JSON schema draft (to support `type: string` & `subType: relative-path`.
See:
http://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7
http://json-schema.org/draft/2019-09/release-notes.html#format-vocabulary
3. Depend on fork: https://github.com/xeipuuv/gojsonschema/pull/256

My opinion:

~I wouldn't like to implement the new **entire** schema loader that focuses only on sub-type (path). KISS approach: define type-based Go validation functions (compiled into the validator). So far we need to apply them only to package manifests (Image type).~


ycombinator:
Yeah, I'm in favor of KISS. What you have here is fine for now (I realize it's not threadsafe). We can worry about that when it starts to become an issue.

code/go/internal/validator/folder_item_spec.go
```diff
@@ -93,3 +94,9 @@ func (s *folderItemSpec) validate(fs http.FileSystem, folderSpecPath string, ite
 	}
 	return errs
 }
+
+func refreshFormatCheckersContext(currentPath string) {
```
ycombinator, line 98: Given that this is a global instruction, should we just move it into `init` to make that more explicit/obvious? It will also guarantee that the `relative-path` format checker is initialized exactly once at package initialization time.

mtojek: Let me describe the case here:

The method `refreshFormatCheckersContext` is called just before calling `Validate` to modify the global state every time it's executed against a package, because there is no way to sneak a context (with a package path).

ycombinator: Ah yes, of course! I totally missed the `currentPath` argument 🤦.

Okay, I understand what's going on here now and why you have to do it this way. This is good as-is but I made a related comment at the location where this function is being called from.

code/go/internal/validator/folder_item_spec.go
```diff
@@ -78,6 +78,7 @@ func (s *folderItemSpec) validate(fs http.FileSystem, folderSpecPath string, ite
 
 	// validation with schema
 	documentLoader := gojsonschema.NewBytesLoader(itemData)
+	refreshFormatCheckersContext(filepath.Dir(itemPath))
```
ycombinator, line 81: Given that we are essentially creating a new custom format checker for every `filepath.Dir(itemPath)`, perhaps we should also remove this format checker using `defer gojsonschema.FormatCheckers.Remove(...)`?

Additionally I would add a comment here noting that this call is not thread-safe so, in the future, if we need to introduce concurrency within a single `elastic-package lint` invocation, we can add mutexes here or something like that.

mtojek: I can add a mutex to surround this place, but it's just mitigation.

Just wanted to make sure that we understand the risk and consequences here (incl. error messages).

ycombinator: Up to you if you want to add the mutex now or later. I don't think it's going to make much of a performance difference to have it now itself but if you want to KISS feel free to leave it out until we need it.

mtojek: Ok, it wasn't hard to add the mutex, so I did it. I've also adjusted the error message.

code/go/internal/validator/folder_item_spec.go
```diff
@@ -89,7 +100,17 @@ func (s *folderItemSpec) validate(fs http.FileSystem, folderSpecPath string, ite
...
 	var errs ValidationErrors
 	for _, re := range result.Errors() {
-		errs = append(errs, fmt.Errorf("field %s: %s", re.Field(), re.Description()))
+		errs = append(errs, fmt.Errorf("field %s: %s", re.Field(), adjustErrorDescription(re.Description())))
 	}
 	return errs
 }
+
+func loadRelativePathFormatChecker(currentPath string) {
+	gojsonschema.FormatCheckers.Add("relative-path", RelativePathChecker{
```
ycombinator, line 109: Consider making `relative-path` a constant, now that we are using it in a couple of places in code.

versions/1/manifest.spec.yml
```diff
@@ -114,6 +114,7 @@ spec:
           src:
             description: Relative path to the icon's image file.
             type: string
+            format: relative-path
```
ycombinator, line 117: This should be documented in `CONTRIBUTING.md` for anyone who wants to add elements to the spec in the future that could use `relative-path` validation.

mtojek: Fixed

code/go/internal/validator/folder_item_spec.go
mtojek: Fixed

CONTRIBUTING.md
```diff
@@ -10,4 +10,26 @@ be prefixed with a package name (`{PACKAGE_NAME}-.+\.json`).
...
+  description: Relative path to the screenshot's image file.
+  type: string
+  format: relative-path
+  examples:
+  - "/img/apache_httpd_server_status.png"
+```
+
+Currently, the following custom formats are available:
+
+* `relative-path`: Relative path to the package root directory. The format checker verifies if the path is correct and
```
ycombinator, line 34: Is it relative to the package root directory or is it relative to where the folder item spec file is defined?

mtojek: It is relative to the package root (as noted).

ycombinator (approved):
LGTM.

PR #84, opened 2020-11-17
Define README file template in _dev

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/151

This PR adds definition for README file template to be stored in the `_dev/build/docs` directory. This file will be used to regenerate the README file.


code/go/internal/validator/folder_item_content.go
```diff
@@ -49,6 +49,7 @@ func loadItemContent(itemPath, mediaType string) ([]byte, error) {
 			return nil, errors.Wrapf(err, "converting YAML file to JSON failed (path: %s)", itemPath)
 		}
 	case "application/json": // no need to convert the item content
+	case "text/markdown":
```
ycombinator, line 52: Add the same comment here as you did for `application/json`?

versions/1/_dev/build/spec.yml
```diff
@@ -0,0 +1,8 @@
+spec:
+  additionalContents: false
+  contents:
+    - description: Folder containing resources related to docs.
+      type: folder
+      name: docs
+      required: false
```
ycombinator, line 7: Should we make this `true`? IOW, should every package be **required** to have a README?

ycombinator:
BTW, I'm noticing that package README files seem to be under `<pkgroot>/docs/`, e.g. https://github.com/elastic/integrations/tree/master/packages/apache/docs. However, this PR suggests that they should be under `<pkgroot>/_dev/build/docs/`. Why the difference?

versions/1/_dev/build/spec.yml
mtojek: No. It's possible to have a package without a README template. The one that doesn't have at the moment in the `log` package..

ycombinator: Okay, but does it make sense for the `log` package to have a README? Or to ask differently: what would be the downside of defining a README for the `log` package?

mtojek:
> BTW, I'm noticing that package README files seem to be under <pkgroot>/docs/, e.g. https://github.com/elastic/integrations/tree/master/packages/apache/docs. However, this PR suggests that they should be under <pkgroot>/_dev/build/docs/. Why the difference?

Here we're talking about additional **builder** resources that shouldn't be part of a distribution in the package-storage. I put the README template in the `_dev/build/docs` to signal that the README template is used only in the build step (similarly to `_dev/test` and test definitions/configs).

I don't insist on this design, if you prefer to put it under `_dev/docs`, it's fine for me.

versions/1/_dev/build/spec.yml
mtojek: You mean, a README file template.

We can it there, in this case both files will have the same content.

ycombinator:
Ah, I got confused because the PR references elastic/elastic-package#151, which mentions removing the `GenerateDocs` command from integrations. I looked at that command it doesn't seem to operate off any template so I got confused.

I think I understand now: developers will define a README template under `<pkgroot>_dev/...` and then they could run `elastic-package ...` to regenerate their `<pkgroot>/docs/README.md` from it. If this understanding is correct, then I'm good with your proposal of putting the README template under `<pkgroot>/_dev/build/docs/`.

versions/1/_dev/build/spec.yml
ycombinator: Yes, this is partly related to my confusion as described in https://github.com/elastic/package-spec/pull/84#issuecomment-729541408 😄 . 

Quick questions:
- say a developer defines a README template for a package but also accidentally edits the README by hand such that it's no longer in sync with the template. I assume our tooling would catch this, maybe as part of `elastic-package lint` or something?
- for packages that define a README template, it sounds like a developer would run `elastic-package build` to (re)generate the README. Normally, `elastic-package build` places the resulting package into the `build` folder, but in this case it sounds like it would also place an updated `README.md` in the package source folder itself, right?

mtojek: > say a developer defines a README template for a package but also accidentally edits the README by hand such that it's no longer in sync with the template. I assume our tooling would catch this, maybe as part of elastic-package lint or something?

Yes, the `lint` command can detect it in advance. Keep in mind that when developer runs `build` command, it will overwrite the README.md.

>for packages that define a README template, it sounds like a developer would run elastic-package build to (re)generate the README. Normally, elastic-package build places the resulting package into the build folder, but in this case it sounds like it would also place an updated README.md in the package source folder itself, right?

Correct. README files should be stored in the package as they can be visualized in the Github preview online.

ycombinator: Okay, in that case, I agree with keeping the template as optional.

ycombinator (approved):
Just https://github.com/elastic/package-spec/pull/84/files#r525834777 then LGTM.

code/go/internal/validator/folder_item_content.go
mtojek: I added a comment, but not the exact (the reason is a bit different).

PR #88, opened 2020-11-24
Support test config for all pipeline test kinds

mtojek:
Issue: https://github.com/elastic/elastic-package/issues/184

Spotted in: https://github.com/elastic/elastic-package/pull/185


versions/1/data_stream/_dev/test/pipeline/spec.yml
```diff
@@ -31,6 +31,22 @@ spec:
             description: Multi-line configuration
             type: object
             additionalProperties: true
+    - description: Configuration of test case in JSON format
```
ycombinator, line 34: AFAICT this section is now duplicated with the one above it for configuring test cases in raw format, except for `multiline`. I wonder if it would be possible to extract the common bits into a common definition section and reference that from both sections? That way if we change the configuration for one type of pipeline test, the other one will automatically get the same changes.

I think this reuse+extension could be achieved using https://json-schema.org/understanding-json-schema/structuring.html#extending?

ycombinator: If this proves to be too difficult to achieve (e.g. it requires updating a library in the validator), then let's go with what you have here.

mtojek: I extended schemas according to your suggestion. There are common-config and log/json-config schemas. The problem is that all files require to set `additionalProperties: true`, because schemas under `allOf` are validated independently (in their own context).

For reference:
https://stackoverflow.com/questions/38737860/how-to-use-additionalproperties-with-allof-in-json-schema

With the approach we validate only defined properties and allow for any of them (which is not expected).

Actually I vote for reverting this last commit and stick to separate, duplicated schemas.

ycombinator: I see, thanks for trying this out. I agree with you that it's better to revert the commit and go back to duplicated schemas than to set `additionalProperties: true` and loosen the schema in the process. Sorry for the trouble.

mtojek: Thanks! Reverting.

mtojek:
@ycombinator I will wait with merging until the issue with status reporting is fixed.

mtojek:
jenkins run the tests please

ycombinator (approved):
LGTM. WFG.

PR #91, opened 2020-12-07
add hidden field in to data stream manifest

nnamdifrankie:
Issue: https://github.com/elastic/security-team/issues/548

This PR adds the the hidden fields to data stream manifest file. The hidden field allows fleet to mark the data stream as hidden https://github.com/elastic/elasticsearch/pull/63987 . This hidden feature is needed for issue 548 to hide the data stream from views. 

### `hidden` field
Further, the `hidden` field will correspond with this ES change: https://github.com/elastic/elasticsearch/pull/63987 .  We want to give the flexibility for package authors to denote a `data stream` as hidden for purposes such as this where a `data stream` is not intended for app specific usage and should be hidden to reduce user confusion.

Our specific use case, again, has to do with a `data stream` that is intended to be queried by a separate telemetry server.  As a result, we want the `data stream` to be hidden to reduce user confusion.


versions/1/changelog.yml
```diff
@@ -20,4 +20,7 @@
     link: https://github.com/elastic/package-spec/pull/79
   - description: Allow `condition` under inputs.
     type: enhancement
-    link: https://github.com/elastic/package-spec/pull/87
\ No newline at end of file
+    link: https://github.com/elastic/package-spec/pull/87
```
mtojek, line 23: I think that this PR link is not valid: https://github.com/elastic/package-spec/pull/91

versions/1/data_stream/manifest.spec.yml
```diff
@@ -96,6 +96,9 @@ spec:
       - traces
       examples:
       - metrics
+    hidden:
+      description: Specifies if a datastream is hidden
```
mtojek, line 100: nit: data stream

mtojek (approved):
only some nit-picks, but LGTM!

PR #94, opened 2020-12-09
Support lens Kibana type

mtojek:
This PR enables support the Lens type in Kibana.

Spotted while working on the Apache migration: https://github.com/elastic/integrations/issues/370


ycombinator:
Not sure if we need to update the import-beats script in the integrations repo for this change as well?

ycombinator (approved):
LGTM.

mtojek:
> Not sure if we need to update the import-beats script in the integrations repo for this change as well?

Hm... I'm not sure if there any lens dashboards. It depends also on how the script is implemented, I will check that later.

PR #101, opened 2021-01-05
Update system tests to support multiple cases

adriansr:
This updates the spec to support multiple test cases per system test.

From now, instead of the test configuration being defined in `config.yml`, multiple files are allowed in the form `test-name-config.yml`.


versions/1/data_stream/_dev/test/system/spec.yml
```diff
@@ -1,8 +1,35 @@
...
-    - description: System test configuration
+    - description: Configuration of test case in yaml format.
       type: file
-      pattern: '^config.yml$'
+      pattern: '^test-[a-z0-9_.-]+-config.yml$'
       contentMediaType: "application/x-yaml"
-      required: false
\ No newline at end of file
+      required: false
+      content:
```
ycombinator, line 9: I think we should not define the contents just yet. We can define them later but once if we define them in the spec i may be difficult to change the structure without breaking backwards compatibility.

versions/1/data_stream/_dev/test/system/spec.yml
```diff
@@ -1,8 +1,35 @@
...
+        type: object
+        additionalProperties: false
+        properties:
+          input:
+            description: Input to use for test case.
+            type: string
+            additionalProperties: false
+          service:
```
ycombinator, line 17: This makes sense at the moment since the only service deployer supported today is Docker Compose. But in the future we plan add other service deployers so we may need to revisit this property.

ycombinator (approved):
LGTM.

Issue #64, opened 2020-10-20
[Feature Request] Dynamic variables in data stream names


jalvz:
As per the current [APM Index Strategy proposal](https://docs.google.com/document/d/1y56a9fjkLi6Zen5qGC_JKYM9ljYpBA5W0fgdWilcwYc/edit?ts=5f6c1aab#), we need the ability to use `service.name` as a [data stream name](https://github.com/elastic/package-spec/blob/master/versions/1/data_stream/manifest.spec.yml#L16), or part of a data stream name.

`service.name` is a top level indexed field in APM documents ([example](https://github.com/elastic/apm-server/blob/master/docs/data/elasticsearch/generated/transactions.json), [spec](https://github.com/elastic/apm-server/blob/master/docs/spec/service.json)), and having it in the `data_stream` would give us the guarantee that each APM service will have its own indices. This would improve performance, allow users to have more granular security and retention policies, and simplify the APM experience in general.

So for metrics data, for instance, we would have indices looking like:
`metrics-backend_service-production`
`metrics-backend_service.profiles-production`
`metrics-frontend_service-staging`

Same thing for other types (traces, logs). 

Right now, I believe folders in the package need to match exactly the data stream name set in `manifest.yml`, so we would need to circumvent that limitation as well.

jalvz:
cc @elastic/apm-server

axw:
As above, the current proposal is to have `service.name` as a prefix in the data set, under the assumption that this would be ideal for RBAC or other places where we might want to identify data streams for a given APM service; we can alternatively make it a suffix if it that is more conventional.

CC @ruflin

ruflin:
In the above, we have `*service.profiles` and just `*service`. These are 2 different datasets? But `metrics-backend_service-production` and `metrics-frontend_service-staging` are the same? 

Taking this, when you ingest data, the data stream names are created out of `metrics-{service.name}_service-{data_stream.namespace}`? And the dataset template would apply to `metrics-*_service-*`?

What are the limitations on the service name? It should not contain `-` for sure. It would be nice if we could potentially restrict it further to not have accidental matches? We should test on how we can make the patterns match best, be it with a prefix or postfix. The reason I prefer postfix is because it goes from generic to specific. So metrics is the most generic part, service in the middle and the name of the service is the most specific one concerning the mapping. But I think it will work both ways around.

jalvz:
Sorry, I didn't explain myself well.

[...]

PR #102, opened 2021-01-06
Adds support for templates index pattern

jalvz:
See motivation in https://github.com/elastic/package-spec/issues/64#issuecomment-755342698

It will require Kibana support, basically if this new property is `true`, the index pattern generated for the templates should be `<type>-<dataset>.*-*`
I will open a ticket in Kibana if this is accepted.


ruflin:
@jalvz Can you share a bit more details on how this will be used?

jalvz:
@ruflin you have been too fast :) 
Backlink in description now

versions/1/data_stream/manifest.spec.yml
```diff
@@ -166,6 +166,14 @@ spec:
           type: object
           additionalProperties: false
           properties:
+            index_patterns:
+              description: index_patterns section of index template
```
mtojek, line 170: nit: I prefer to keep a human readable description (without underscore).

versions/1/data_stream/manifest.spec.yml
```diff
@@ -166,6 +166,14 @@ spec:
           type: object
           additionalProperties: false
           properties:
+            index_patterns:
```
mtojek, line 169: Is it possible to define a regular expression for the field value?

mtojek (commented):
Please modify the "good" test package, so we can be sure that the change works as intended.

ruflin (changes requested):
I'm currently blocking this PR as we should first complete the discussion in https://github.com/elastic/package-spec/issues/64#issuecomment-755342698

I like the simplicity of this change but I worry it has many side effects. On the Stack side we loose the knowledge if the templates / mappings / index patterns for a data stream are all aligned or if a dev might have just introduced an invalide index pattern.

jalvz:
Updated as per https://github.com/elastic/package-spec/issues/64#issuecomment-757925161

ruflin:
@skh @jfsiii @jen-huang Could you chime in from the Kibana side if this is feasible?
@mtojek @ycombinator Not happy with the name yet (see conversation in the linked issue), so if you have ideas for better names ...
@jalvz LGTM but lets wait on the feedback from the Kibana side to not merge this too soon.

versions/1/data_stream/manifest.spec.yml
```diff
@@ -111,6 +111,10 @@ spec:
       type: string
       examples:
       - diagnostics
+    match_all:
```
simitt, line 114: How about `dataset_dynamic` or `dataset_is_pattern` as this only changes whether or not the `dataset` part is a string or a pattern)? 
Releated to the description - we could then change it to something like _indicates whether the dataset name can contain a dynamic part only determined during runtime_.

jalvz:
Filed Kibana issue: https://github.com/elastic/kibana/issues/88307

versions/1/data_stream/manifest.spec.yml
```diff
@@ -111,6 +111,10 @@ spec:
       type: string
       examples:
       - diagnostics
+    dataset_is_prefix:
+      description: if true, the generated index pattern will contain the dataset as a prefix only
```
skh, line 115: Is this only about the field `index_patterns` in an ES index template (this is what I guessed from the discussions), or also about Kibana index patterns (this is how this line reads without context)?

ruflin: It is only about the ES index patterns. It should not affect the Kibana ones as these are based on the types and include all datasets anyways.

skh: Then this description should be changed accordingly.

jalvz:
Hey all, are there any more objections to this?

simitt (commented):
LGTM

mtojek (approved):
LGTM, but let's leave the final decision to @ruflin .

EDIT:

please rebase the branch against master and update the changelog.

ruflin (approved):
LGTM

@mtojek @ycombinator Side discussion: Should this go into https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml or where is the right place for this?

mtojek:
> @mtojek @ycombinator Side discussion: Should this go into https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml or where is the right place for this?

Yes, this is the right place.

PR #104, opened 2021-01-12
Cleaning up go.mod and adding command

jonathan-buttner:
Not sure why the statik.go file changed again 🤔  is it because the go.mod and go.sum files changes?


mtojek (commented):
Hm... isn't it because of the changelog.yml which is included in statik.go?

jonathan-buttner:
> Hm... isn't it because of the changelog.yml which is included in statik.go?

Oh yeah that could be.

versions/1/changelog.yml
```diff
@@ -27,7 +27,7 @@
   - description: Add `hidden` fields to data stream manifest file.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/91
-  - description: Revert: Allow `condition` under inputs.
+  - description: 'Revert: Allow `condition` under inputs.'
```
jonathan-buttner, line 30: This was complaining in vscode because `description: Revert:` is invalid yml I think.

mtojek (approved):
Ship it!

PR #105, opened 2021-01-12
Add field to allow skipping of tests

ycombinator:
This PR adds a field to the pipeline and system test configuration files, allowing tests to be skipped.

Related: https://github.com/elastic/elastic-package/issues/218


mtojek (commented):
I'm not in favor of skipping tests, but we may need it in the future. I would enforce the description and link with regexp :) Something like these (all fields required):

```yaml
skip:
   url: 
   explanation:
```

WDYT?

ycombinator:
@mtojek thanks, I like this idea. It forces users to offer a short explanation for the skip and URL that has more details. 👍

ycombinator:
@mtojek I addressed your review feedback. A few points to note:
- I went with `link` instead of `url` because we already have a precedent for `link` here: https://github.com/elastic/package-spec/blob/930b44726a527ac599f55eebdfca114f049bdeff/versions/1/changelog.spec.yml#L37
- I went with `reason` instead of `explanation` because `reason` is shorter. 😄 
- I don't like that the definition of `skip` is repeated in two places. I tried refactoring it into a common file, similar to what we have with https://github.com/elastic/package-spec/blob/930b44726a527ac599f55eebdfca114f049bdeff/versions/1/data_stream/manifest.spec.yml#L35 but I was not successful. I kept getting a `Reference /spec.yml#/definitions/skip must be canonical`.
  - The value of my `$ref` property looked like `"../spec.yml/#definitions/skip"` and in the `data_stream/_dev/test/spec.yml` file I had a `spec.definitions.skip` property.
  - Just for testing I even tried to create a `common.yml` file in the same folder as the referencing spec file and changed the value of `$ref` to `"./common.yml/#definitions/skip"` but that also gave the same error.
  - If you have some ideas as to what might be causing this error or what else I should try, please let me know.

mtojek:
I looked into this problem and found a workaround: https://github.com/mtojek/package-spec/tree/test-skip-fix-1

Most likely the inline content doesn't allow for jumping over remote resources or the location is not evaluated correctly.

ycombinator:
Thanks @mtojek. I merged your workaround into this PR. Now I'm seeing a different test failure, which I'm investigating:

```
=== RUN   TestValidateFile/good
    validator_test.go:50:
                Error Trace:    validator_test.go:50
                Error:          Received unexpected error:
                                found 1 validation error:
                                   1. item [test-access-event.json-config.json] is not allowed in folder [../../internal/validator/test/packages/good/data_stream/foo/_dev/test/pipeline]
                Test:           TestValidateFile/good
```
```

**Comment by ycombinator** [comment]
@mtojek This PR is ready for your re-review when you get a chance. Thanks!

File: versions/1/data_stream/_dev/test/pipeline/config_raw.spec.yml [hunk 1]
```diff
@@ -0,0 +1,20 @@
+spec:
```
mtojek, line 1: I think the config is applicable in both modes now: raw and JSON. It's described also in docs for pipeline tests.

ycombinator: There appears to be a slight difference in the configs for the raw and JSON modes.

Raw: https://github.com/elastic/package-spec/blob/930b44726a527ac599f55eebdfca114f049bdeff/versions/1/data_stream/_dev/test/pipeline/spec.yml#L14-L36

JSON: https://github.com/elastic/package-spec/blob/930b44726a527ac599f55eebdfca114f049bdeff/versions/1/data_stream/_dev/test/pipeline/spec.yml#L37-L55

It looks like the raw mode supports `multiline` whereas the JSON mode does not. 

That's why I had to end up creating two different spec files: `config_raw.spec.yml` and `config_json.spec.yml`.

mtojek: Ah, I see! In this case I think it's correct. Sorry about confusion.

mtojek (approved):

Issue #108, opened 2021-01-18
[Discuss] Process on changes to the package-spec


ruflin:
The package-spec has become a central part of everything around packages, integrations, Fleet, Elastic Agent. Any change or addition to the package-spec can have wide ranging consequences. This issue is to discuss on how we make changes to the package-spec and ensure it does not have any unexpected side effects.

To better explain the problem I'll use an example. The addition of a new asset type like `elasticsearch transform` affects many parts. The package-registry must know how to delivery it, Kibana / Fleet must know how to update / downgrade / install / remove the asset, it must fit into the indexing strategy, elastic-package must know how to test it. So before adding a new assets, it must be checked how it should be added and which parts are effected by it.

The above applies to addition of all asset types but also to changes / additions to the manifest files. How do we ensure nothing breaks? At the same time, we should ensure that the package-spec stays open for contributions from everyone and don't put too much burden on potential changes.

One idea to start with is having a detailed checklist when modifying the package-spec. Any change should go through this checklist.  In addition, it would be nice to have experts from each area that can be pinged for help. Someone wanting to add a new asset for Kibana might not have all the expertise on how this affects elastic-package. So each area could have an expert that can be pinged.

ycombinator:
Thanks for starting this much-needed discussion, @ruflin.

At a high level, I would say the process looks something like this:

1. Whenever there's a potential change that impacts the package spec, this change must be proposed in an issue in the `package-spec` repository. This gives us an opportunity to understand which parts of the stack might be impacted by this change and pull in relevant experts to get their opinions. The initial proposal should cover these areas:
   - what problem the proposal is solving. This provides context and could help shape the solution.
   - where the solution will need to be implemented, aka which parts of the stack will be impacted. Its okay if the initial proposal doesn't get this 100% right; the discussion in the proposal issue will help clarify this. But having some idea initially will help get the ball rolling in terms of pulling in the right people for the discussion.

2. Once we have consensus on the proposal issue, we modify the issue description to include an **ordered** checklist of tasks that need to be resolved to make the change happen in a safe way.  For example, maybe Kibana needs to implement support for a new property in the spec first, then only when the support is implemented, the spec can itself be modified, which would then allow packages to define this property and have these packages still be valid. At this point the proposal issue serves as a meta issue for the safe implementation of the change.

3. We file issues in each of the repositories corresponding to the checklist items and update the checklist with links to these issues.

[...]

PR #110, opened 2021-01-19
Add multi property to streams

leehinman:
## What does this PR do?

Adds multi property to streams.

## Why is it important?

Useful when input for stream may be defined multiple times with only a
few parameters changing.  For example queries for osquery, or event
providers for winlog.

## Checklist

## Related issues

- Relates elastic/kibana#83141


ruflin:
Based on https://github.com/elastic/package-spec/issues/108 would be good to pull in some folks to get feedback on this:

* Integrations / package-registry / package-spec: @mtojek @ycombinator 
* Kibana: @jen-huang

ruflin:
I'm wondering if what we want to repeat is the stream or the full input? Is there a case that we have an input with multiple data streams and we want only one of the data streams to be repeated?

mtojek (commented):
I can see a bunch of use cases around database generic queries. Nice. I'm not convinced about the `multi` property name, but I can't propose anything better (collection? expandable?).

~@leehinman could you please add this property to the test package (the `good` package)?~ (nevermind, I missed it's there)

versions/1/data_stream/manifest.spec.yml
```diff
@@ -156,6 +156,12 @@ spec:
           enabled:
             description: Is stream enabled?
             type: boolean
+          multi:
+            description: Can stream be defined multiple times
+            type: boolean
+            default: false
+            examples:
```
mtojek, line 163: nit: well, this is a boolean field. I don't think we need to give an example here :)

mtojek (approved):
Nothing blocking from the package-registry perspective. I'll leave the final word to the Kibana's representive, @jen-huang .

jen-huang:
> I'm wondering if what we want to repeat is the stream or the full input? Is there a case that we have an input with multiple data streams and we want only one of the data streams to be repeated?

++ to @ruflin's question. We discussed this on the Kibana issue back in November: [example of only stream repeating](https://github.com/elastic/kibana/issues/83141#issuecomment-728521427), [example of full input repeating](https://github.com/elastic/kibana/issues/83141#issuecomment-728991787).

Which do you expect Kibana to generate?

ycombinator:
@ruflin @jen-huang, just going through PRs today after being on PTO for a week. Seems like this PR is currently blocked on this question:

> I'm wondering if what we want to repeat is the stream or the full input? Is there a case that we have an input with multiple data streams and we want only one of the data streams to be repeated?

Looking at the two examples that @jen-huang linked to in her comment, @ruflin can you explain why [repeating the full input](https://github.com/elastic/kibana/issues/83141#issuecomment-728991787) in the agent config would be better than [repeating the data stream](https://github.com/elastic/kibana/issues/83141#issuecomment-728521427)? I think understanding the pros/cons of each approach might help move this discussion forward to a resolution.

ruflin:
For the input vs stream part: I don't think one is better than the other, it is more the question on what is trying to be solved here. Looking at https://github.com/elastic/kibana/issues/83141 There are 2 checkboxes, one for the input and one for the stream. But as there is only 1 stream, the two should probably be merged (other issue). If both are merged, the only thing that is left is the input.

Where the two differ is on shared config: Repeating the stream would allow to reuse the input config for multiple streams. Repeating the input allows to have different input config for each entry. 

In the case of osquery, it seems nothing is shared. So the outcome would be the same. Is osquery really using a single data stream for all the data? Or is it more similar to apm and the input configs could even be global.

What would be clarifying I think would be seeing a mockup on how it should look in the UI and decide if input configs should be shared or not.

jamiehynds:
> Is osquery really using a single data stream for all the data? Or is it more similar to apm and the input configs could even be global.

@james-elastic could you confirm which approach the new osquery integration has taken?

aleksmaus:
@jamiehynds as of now it's a single output datastream. 
The results of all the queries that osquery executes end up in the same ES datastream. 
Each query was configured as input, and we wanted to be able to run multiple queries at different time intervals. That's why we are asking to be able to configure multiple. If there is a better way to accomplish that, let us know.

mtojek:
@ruflin @leehinman is this PR still valid?

ruflin:
I think it is still valid but would be good to get clarification on https://github.com/elastic/package-spec/pull/110#issuecomment-767401108 from @leehinman

leehinman:
> For the input vs stream part: I don't think one is better than the other, it is more the question on what is trying to be solved here. Looking at [elastic/kibana#83141](https://github.com/elastic/kibana/issues/83141) There are 2 checkboxes, one for the input and one for the stream. But as there is only 1 stream, the two should probably be merged (other issue). If both are merged, the only thing that is left is the input.
> 
> Where the two differ is on shared config: Repeating the stream would allow to reuse the input config for multiple streams. Repeating the input allows to have different input config for each entry.
> 
> In the case of osquery, it seems nothing is shared. So the outcome would be the same. Is osquery really using a single data stream for all the data? Or is it more similar to apm and the input configs could even be global.
> 
> What would be clarifying I think would be seeing a mockup on how it should look in the UI and decide if input configs should be shared or not.

For Windows Event logs I think repeating the input is a better fit, this is to support Custom Windows Event Logs, so each entry might need access to every option.

jen-huang:
> For Windows Event logs I think repeating the input is a better fit, this is to support Custom Windows Event Logs, so each entry might need access to every option.

It sounds like we agree on repeating the full input for this use case. With that in mind, might it be better to add `multi: true` to the package's `policy_templates.inputs` on the [main manifest](https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L190) instead of at the stream level? I think this is a more accurate representation of what we want to achieve in the UI. If we find a case in the future where we want repeatable streams, in addition to repeatable inputs, then we can add the stream-level flag at that point.

What do you think @mtojek @ruflin?

ruflin:
@jen-huang Agree

jen-huang (changes requested):
After understanding what we want the resulting agent yaml to look like, I'd like to see this property moved to the `policy_templates.inputs` level on the [main package manifest](https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L190) instead of at the stream level. This makes more sense from a Kibana perspective too.

leehinman:
> After understanding what we want the resulting agent yaml to look like, I'd like to see this property moved to the `policy_templates.inputs` level on the [main package manifest](https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L190) instead of at the stream level. This makes more sense from a Kibana perspective too.

Made the change and rebased to master.  Let me know if we need to change anything else.  Thanks.

ruflin:
Thanks @leehinman Would be good for @jen-huang to have a quick look at it when she is back.

@mtojek @sorantis  Will this work as expected with thew new introduce of multiple policy_templates?  I assume yes but want to make sure I don't miss something.

jen-huang (approved):
LGTM from Kibana side. Thanks for the discourse and making the changes!

mtojek:
Please make sure to correct all conflicted files and merge the PR if only the status is green.

PR #111, opened 2021-01-20
Support Terraform deployment files

mtojek:
## What does this PR do?

This PR introduces new Terraform-based deployment service definitions. Deployment definitions will enabled for both data stream and package (reuse). 

## Why is it important?

We're moving on with service deployer for AWS, GCP, etc.

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/issues/89


versions/1/data_stream/_dev/spec.yml
```diff
@@ -5,4 +5,9 @@ spec:
...
       name: test
       required: false
-      $ref: "./test/spec.yml"
\ No newline at end of file
+      $ref: "./test/spec.yml"
+    - description: Folder containing configuration related to deploying the package's integration service(s)
+      type: folder
+      name: deploy
+      required: false
+      $ref: "./../../_dev/deploy/spec.yml" # Use "deploy" spec for package
```
ycombinator, line 13: Is this what we want? It implies that there can be data stream level deployment definitions for Docker Compose-based deployments as well, not just Terraform-based deployments. Having such consistency is really nice but I want to make sure it's something that's actually going to be supported by `elastic-package` as well. Otherwise the spec is a bit misleading.

mtojek: Yes, I modified the `elastic-package` here: https://github.com/elastic/elastic-package/pull/228 . 

As you noticed, It means that we will support deploy definitions on the data stream level. Actually we may leverage from this change, as currently developers build a single Docker image for multiple data streams. See: https://github.com/elastic/integrations/tree/master/packages/zeek/_dev/deploy/docker

If you prefer to support `tf` only, I can adjust the spec.

mtojek:
I will solve all conflicts (`statik.go` and `changelog.yml`) once everything else is fine.

versions/1/data_stream/_dev/spec.yml
ycombinator: Ah nice, I hadn't studied https://github.com/elastic/elastic-package/pull/228 in depth before, sorry! Given what you have there, what you have here is good as-is. Thanks!

ycombinator (approved):
LGTM once merge conflicts are addressed and CI is green.

Issue #112, opened 2021-01-20
Need to be able to create a package without data_stream.


aleksmaus:
Presently we had to add a fake-ish data_stream for the Fleet Server package in order to pass the linter checks
https://github.com/elastic/integrations/pull/544

Otherwise were getting a error
```
Error: checking package failed: linting package failed: found 1 validation error:
   1. expecting to find [data_stream] folder in folder [/Users/amaus/elastic/integrations/packages/fleet_server]
```

PR #113, opened 2021-01-21
Make data_stream optional for a package spec

aleksmaus:
## What does this PR do?

Allows data_stream folder to be optional in the package

## Why is it important?

This flexibility was needed for the Fleet Server package, issue: https://github.com/elastic/package-spec/issues/112

## Checklist

## Related issues
Closes https://github.com/elastic/package-spec/issues/112


code/go/internal/validator/test/packages/missing_data_stream/kibana/visualization/good-visualization-abc-1.json
```diff
@@ -0,0 +1 @@
+{}
```
ruflin, line 1: you can probably skip / remove all the kibana files as they should not be needed for the test package.

code/go/internal/validator/test/packages/missing_data_stream/manifest.yml
```diff
@@ -0,0 +1,41 @@
...
+            type: text
+            title: Hosts
+            multi: true
+            required: true
+            show_user: true
+            default:
+            - http://127.0.0.1
+owner:
+  github: elastic/foobar
+screenshots:
```
ruflin, line 28: I think scnreenshot and icons are not required, lets remove it to keep the example package simpel.

ruflin (commented):
LGTM.

@aleksmaus You mentioned that the package also passed before you made the change to set the required to false? This makes me suspicous.

mtojek (commented):
Keep in mind that you'll have to update dependency on the package-spec in the elastic-package too (and in integrations).

code/go/internal/validator/test/packages/missing_data_stream/manifest.yml
aleksmaus: yeah, was not sure what was needed and what was not. just copied over the "good" package and removed the "data_stream" dir. will update

aleksmaus:
> LGTM.
> 
> @aleksmaus You mentioned that the package also passed before you made the change to set the required to false? This makes me suspicous.

I didn't put the test for the right validator. Had to dig a bit to understand which one is validating what. Got the failing test with missing ```data_stream```, then adjusted the spec to have the ```data_stream``` optional and the test passed.

code/go/internal/validator/test/packages/missing_data_stream/manifest.yml
```diff
@@ -0,0 +1,27 @@
+format_version: 1.0.4
```
ruflin, line 1: @mtojek Cant' comment on the `.empty` commit but that looks like a hack :-) Should the README be required? @aleksmaus Is this because a `docs` directory is required?

mtojek: It's because the docs directory is open (doesn't have any limitation). We may enforce some rules, but this way we may block some user files. 

For reference: https://github.com/elastic/package-spec/blob/master/versions/1/spec.yml#L30

mtojek (approved):

code/go/internal/validator/test/packages/missing_data_stream/manifest.yml
ruflin: I like the idea that it is enforce. And I think we required in the UI the README.md file, so we could enforce this one too?

mtojek: https://github.com/elastic/package-spec/pull/116

PR #116, opened 2021-01-25
Enforce README.md file

mtojek:
## What does this PR do?

This PR requires the presence of README.md file.

## Why is it important?

The rest of tooling requires this file or assumes it's present. We need to make it consistent.

## Checklist

## Related issues

- https://github.com/elastic/package-spec/pull/113#discussion_r563535752


ruflin (commented):
LGTM. Will this work with all the existing packages?

mtojek:
> LGTM. Will this work with all the existing packages?

If it doesn't, then I will adjust it. I believe we have a hard requirement on README.md presence.

versions/1/docs/spec.yml
```diff
@@ -0,0 +1,8 @@
+  spec:
+    additionalContents: true
```
ycombinator, line 2: Are we expecting to see docs other than `README.md` in the docs folder? Looking through the `integrations` packages, I'm only seeing `README.md` files under docs. I'd prefer that we are stricter now by setting `additionalContents: false` and then loosen the restriction later if/when we have a use case for it.

ruflin: The initial design for the docs directory was that it could contain multiple files which can also reference / include each other but for simplicity we ended up going with just a single file for now. So at the moment, I don't think it should contain any other file so we can lock it down. We can still open it up later if we have new requirements.

mtojek: Fixed!

ycombinator (approved):
LGTM.

PR #120, opened 2021-01-28
Add alias type

justinkambic:
## What does this PR do?

Attempting to add an `alias` field type.

## Why is it important?

We need [a field](https://github.com/elastic/integrations/compare/master...justinkambic:new-new-synth-package#diff-4875b7c06dfbe454fb30579ac609f80c5f5908a0d8a3947763b79bd801bfeb23R7) of this type for synthetics package.

## Checklist

- [ ] I have added test packages to [`code/go/internal/validator/test/packages`](https://github.com/elastic/package-spec/tree/master/code/go/internal/validator/test/packages) that prove my change is effective.


ycombinator:
@justinkambic would you mind posting a link to exactly where in the synthetics package you require the alias? It might help provide additional context. Thanks!

@ruflin we had a discussion about alias types. I'm okay with adding this but just want to get your thoughts in case I'm missing something. Thanks!

justinkambic:
> would you mind posting a link to exactly where in the synthetics package you require the alias?

I've updated the description

mtojek (changes requested):
Please run `make update && make test`, so you can get the updated `statik.go` and verify if tests still work.

And the CHANGELOG please.

ruflin:
In general I don't see an issue to add support for the `alias` type. But @justinkambic the field you linked to seems odd:

```
    # - name: url
    #   type: alias
    #   path: url.full
    #   migration: true
    #   description: >
    #     Service url used by monitor.
```

How will this work? A field can't be an object and an alias at the same time AFAIK? Also in case it would work, how is this field used? I see the `migration: true` flag in there which means this was a field used for migration from 6.x to 7.0 so should definitively not be in the integrations packages. Were the fields reviewed or just copy / pasted from Heartbeat fields.yml? @andrewvc for awareness.

ycombinator:
> I see the migration: true flag in there which means this was a field used for migration from 6.x to 7.0 so should definitively not be in the integrations packages. Were the fields reviewed or just copy / pasted from Heartbeat fields.yml?

This is just a copy/paste for now. The actual package PR has not yet been reviewed so unnecessary fields like `migration: true` that no longer make sense in the context of packages would be removed during review.

ycombinator:
@justinkambic Could you add an entry to the CHANGELOG please? The link to it is in the checklist in the PR description.

justinkambic:
Sorry, yes I will do it now.

ycombinator (approved):
LGTM but I'd like @ruflin to review this change as we've had some discussions in the past about field aliases in packages.

ruflin:
@justinkambic I do not want to block this PR but a quick comment on https://github.com/elastic/package-spec/pull/120#issuecomment-769642588 would be good to clarify. I'm worried if we just get the PR in the discussion will not happen.

justinkambic:
> a quick comment on #120 (comment) would be good to clarify

Like @ycombinator mentioned, these fields are a result of copy/paste, so we haven't removed unnecessary fields like that `migration: true` yet. This particular `alias` field just simplifies querying `url.full`, as far as I know. @andrewvc might want to comment any additional insight.

After encountering errors trying to finish setting up our package, we felt that `alias` had no real cause _not_ to be in the spec, which is why we decided to open this PR.

ruflin (approved):
Approving to not block the change on adding alias.

@justinkambic I probably miss something but I would assume the package does not work in this state.

mtojek (approved):

PR #124, opened 2021-02-03
Adding contribuing guide section + issue template for change proposals

### LINKED ISSUE #108: [Discuss] Process on changes to the package-spec

ycombinator:
## What does this PR do?

Introduces a section in the Contributing Guide + a GitHub issue template to guide users on how and why to propose changes to the Package Specification in this repository.

## Why is it important?

So we follow a consistent process that promotes discussion amongst relevant parties + a safe approach to introducing changes to the Package Specification and Elastic Stack components.

## Related issues

- Closes #108.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Change_Proposal.md
+++ b/.github/ISSUE_TEMPLATE/Change_Proposal.md
@@ -5,5 +5,5 @@ labels: discuss
 title: "[Change Proposal] "
 ---
 
-Please read the section on [Change Proposals in the Contributing Guide](/CONTRIBUTING.md#change-proposals) and flesh out
-this issue accordingly. Thank you!
+Please read the section on [Change Proposals in the Contributing Guide](/CONTRIBUTING.md#change-proposals) 
+and flesh out this issue accordingly. Thank you!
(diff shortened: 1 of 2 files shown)


mtojek (approved):
I like the point no. 4, which makes the spec's issue a meta-issue for rolling out changes. The PR looks good to me, but we may lack a way of advertising this guide (people might not be aware of the spec repo and the CONTRIBUTING inside).

ruflin:
@ycombinator Great to have this documented. Can you make sure this is communicate widely to raise awareness?

PR #130, opened 2021-02-08
Support kubernetes service deployer

mtojek:
## What does this PR do?

This PR defines files used by the Kubernetes service deployer.

## Why is it important?

These files will define a custom service deployed on top of the Kubernetes cluster. The Elastic Agent can monitor this service (e.g. nginx deployed in the cluster). If there aren't any files required to be installed in the cluster, it's mandatory to set the `.empty` file as a mark for the test runner to use Kubernetes service deployer (empty directory `_dev/deploy/k8s` won't be versioned).

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/issues/239


mtojek:
@ycombinator is ready for review, but please take a look at the https://github.com/elastic/elastic-package/pull/246 to verify if the proposal here is correct.

code/go/internal/validator/folder_item_content.go
```diff
@@ -17,8 +17,8 @@ func loadItemContent(itemPath, mediaType string) ([]byte, error) {
 		return nil, errors.Wrap(err, "reading item file failed")
 	}
 
-	if len(itemData) == 0 {
-		return nil, errors.New("file is empty")
+	if len(itemData) == 0 && mediaType != "" {
```
ycombinator, line 20: It took me a second to realize that this additional check was required to properly handle the `.empty` file case. In the case of the `.empty` file, `len(itemData) == 0` but `mediaType == ""` so this check will evaluate to `false`, which is what we want. Without the additional `mediaType != ""` clause, this check would have evaluated to `true` and returned an error, which is not what we want.

Would you mind adding a comment above this check explaining why we need to check both the length of `itemData` and the media type?

ycombinator (approved):
Left a minor comment about adding an explanatory comment. 🙂 

Otherwise, LGTM.

code/go/internal/validator/folder_item_content.go
mtojek: Fixed.

mtojek: I'll merge it once the CI is green and doesn't complain.

Issue #17, opened 2020-08-11
Package CHANGELOGs


ycombinator:
Forked off from discussion about package CHANGELOGs in https://github.com/elastic/package-spec/pull/14.

We need to decide how best to capture CHANGELOGs for packages. Some ideas that have been proposed so far:
- Writing a file at the root of the package's folder,
- Generating them from commit messages,
- Generating them from PR descriptions

[UPDATE] After discussion below and consensus on where and how package CHANGELOGs should be stored, I'm adding a few checklist items below for next steps to implement the agreed-upon changes:

* [x] Create PR to package spec to make CHANGELOG files required: https://github.com/elastic/package-spec/pull/131
* [x] Once package spec PR is approved **but not yet merged**: 
   * [x] Update `integrations` repo with skeleton CHANGELOG files for each integration package: https://github.com/elastic/integrations/pull/675
   * [x] Inform non-`integrations` repo package owners about adding CHANGELOG files: email sent
* [x] Once existing packages have been updated with skeleton CHANGELOG files:
   * [x] Merge package spec PR: https://github.com/elastic/package-spec/pull/131
   * [x] Create/review/merge PR to update `elastic-package` with package spec changes: https://github.com/elastic/elastic-package/pull/252
   * [x] Create/review/merge PR to update `integrations` repo with `elastic-package` changes: https://github.com/elastic/integrations/pull/675

ycombinator:
Some comments, in chronological order, copied over from #14 for ready reference:

---

@mtojek said:

> I'm allergic to changelogs, probably @ruflin would like to elaborate on them :) I suppose we will have to start using them, but I'm not sure if people get mad, when you ask them to fill them up.

---

@ruflin said:

> @mtojek For changelog: We need a way to tell users of the packages on what change between versions. I'm not tied at all to changelogs but we need a solution.

---

@ycombinator said:

> My 2 cents: I like the proposed CHANGELOG yml format. It captures what we need, but in a structured fashion. So one can imagine generating human-readable changelog documentation from it, etc.

>> I'm allergic to changelogs,

> @mtojek could you elaborate on why this is? maybe there's a drawback that I'm not seeing currently.

---

@urso said:

> And I'm getting tired of yml :)

What issue does using YAML solve here?

---

@ycombinator said:

[...]

PR #131, opened 2021-02-08
Introducing package CHANGELOGs

ycombinator:
## What does this PR do?

This PR enhances the package spec to require `CHANGELOG.yml` files at the root folder of every package.

## Why is it important?

So package consumers can track changes to packages across versions.

## Checklist

## Related issues

- Related: #17


mtojek (approved):
Nice! Looking forward to seeing this in action.

ycombinator:
@mtojek Thanks for reviewing this PR but I thought of a couple of things after I put up this PR so I wanted to raise them here and get your opinions:

1. According to the current implementation in this PR, the `changelog.yml` file must go at the root folder of the package. I also considered adding it to the `docs` folder but decided against it as I thought we should keep the `docs` folder for any **human-readable** formats like `.md`. I can imagine, in the future, `elastic-package build` generating a `docs/changelog.md` file from `changelog.yml` if we have a use case for it.

2. According to the current implementation in this PR, the `changelog.yml` file contains changes for **all** versions of the package. An alternative approach would be for the `changelog.yml` file to contain changes **only for the current version** of the package. This has some advantages and disadvantages:
   * _Advantage_: every package version does not have to contain it's own changelog entries and also changelog entries from past package versions. If a package is released frequently, the `changelog.yml` in each subsequent version would not grow larger and larger over time.
   * _Advantage_: the package author would not need to repeat the version in the `changelog.yml` from the `manifest.yml`. We can simply assume that the contents of `changelog.yml` apply to the version mentioned in the `manifest.yml`.
   * _Disadvantage_: when bumping up a package version, the package author would need to remember to clear out the existing `changelog.yml` and start with a clean slate for the new version.
   * _Disadvantage_: not sure if this is a disadvantage _per se_, but if a client (e.g. the Fleet UI) wants to consume the **entire** set of changes between two versions of a package or for all versions of a package, we would need a mechanism to stitch together changelogs from individual package versions into a larger list. This could be done via a new Package Registry API.

mtojek:
> According to the current implementation in this PR, the changelog.yml file must go at the root folder of the package. I also considered adding it to the docs folder but decided against it as I thought we should keep the docs folder for any human-readable formats like .md. I can imagine, in the future, elastic-package build generating a docs/changelog.md file from changelog.yml if we have a use case for it.

Most likely it's personal preference, but it's easier to find if there is a changelog package if it's in the root, close to the `README.md`. I'm fine with other solution too (put it in `docs/`).

> According to the current implementation in this PR, the changelog.yml file contains changes for all versions of the package. An alternative approach would be for the changelog.yml file to contain changes only for the current version of the package. This would have some advantages and disadvantages:

I think it might be easier to keep the changelog file additive and don't clean it before releases. Actually it might be convenient not to jump over versions to check when a feature has been released.

ycombinator:
Thanks for the feedback, @mtojek. Based on it, I'm going with the changes in this PR as-is.

ruflin:
Thanks for getting this in. Looking forward to see this showing up in the packages.

Issue #132, opened 2021-02-10
[Discuss] Proposal: support integration sets


mtojek:
Hi everyone,

The goal of this issue is to discuss challenges, benefits and drawbacks of introducing a new package structure (potentially v2?), which can satisfy the requirement of service granularity (reported in https://github.com/elastic/package-spec/issues/122). I've prepared a sample package to visualize the proposal.

MySQL pseudo-integration: https://github.com/mtojek/integrations/tree/v2-integration-sets/packages/mysql-2 

Requirements:
* fit in the new service granularity model proposed by @sorantis
* have single manifest file (for package and all data streams), so it would be easier for Kibana and EPR to load and process package description
* keep large contents (pipeline definitions, icons, screenshots, agent's configs) in separate resources
* support extending resources, e.g. manifests contain vars on every level, fields are on multiple levels in the directory tree

Notes/observations:
* Integration policies defines groups of inputs: logs, metrics, packets (packetbeat), synthetics (uptime). I suppose we'll need also traces
*  With this approach, we'll drop the original idea of a package - a group of dashboards, pipelines, etc. The new concept changes the meaning of not just integrations.
* `data_stream` names are aligned with the manifest structure, e.g.:
  ```yaml
  integration_sets:
  - name: basic
    integration_policies:
      logs:
        inputs:
          - name: error
            ilm_policy: custom-hot-warm-logs
            vars: ~ # ...
  ```
  .. gives - `logs-mysql-basic.error-shared`. I suppose there should be an option for the end-user to correct/change it.

Please do not consider this is as future approach or a kick-off for adjustments, it's rather an exercise to understand and estimate the amount of work in multiple places. It would be nice to hear some feedback from all parties.

EDIT:

Do not look at the manifest as it's the final/target format. Please rather focus on concepts/requirements.

mtojek:
Let me ask few people to join the discussion.

Kibana: @jen-huang @skh 
AWS Integration/Integrations: @kaiyan-sheng @masci 
EPR: @ruflin @ycombinator 
Fleet: @ph @nchaulet 

It would be great to justify first if we're able to introduce such changes and when would be the right time.

ph:
Let's keep @mostlyjason @andresrc in the loop

sorantis:
[Proposed integration architecture](https://docs.google.com/document/d/15_-HSYvctps9M4BcyLaF614W_uys3ZPkWqK83Jk9N7c/edit#)

sorantis:
[...]

Issue #106, opened 2021-01-12
Should pipeline test config be in YAML?


ycombinator:
The system test configuration is expressed as YAML and many other package files are also expressed as YAML. However, pipeline test configuration files (example: https://github.com/elastic/package-spec/blob/master/code/go/internal/validator/test/packages/good/data_stream/foo/_dev/test/pipeline/test-access-raw.log-config.json) is expressed as JSON. Should we change this to be YAML as well?

## Rollout plan for this change
* [x] Enhance package spec to allow `*.yml` pipeline test configuration files in addition to the currently-allowed `*.json` ones. Mark the `*.json` file format as deprecated (exactly how to do this is TBD). This will give package authors some time to adjust their pipeline test configuration file formats: https://github.com/elastic/package-spec/pull/139.
* [x] Update all pipeline test configuration files in the `integrations` repo from `*.json` => `*.yml`: https://github.com/elastic/integrations/pull/739.
* [x] Notify package stakeholders that `*.json` pipeline test configuration files are being removed and replaced by `*.yml` ones.
* [x] After a few weeks:
   * [x] Notify package stakeholders that support for `*.json` pipeline test configuration files is being removed.
   * [x] Remove support for `*.json` pipeline test configuration files from package spec: https://github.com/elastic/package-spec/pull/150.
   * [x] Notify package stakeholders that support for `*.json` pipeline test configuration files has been removed.

andrewkroh:
I noticed this inconsistency and was wondering about it. +1 for using YAML for both.

ycombinator:
I'm going ahead with this change now. I've updated the issue description with a plan on how to roll out this change safely. Please review it.

mtojek:
@ycombinator I think you can move forward with this issue.

ycombinator:
Thanks for the reminder, @mtojek. I'll move this forward today.

ycombinator:
All steps in the rollout plan (see issue description) are now done. Closing.

PR #137, opened 2021-02-19
Support policy groups

mtojek:
## What does this PR do?

This PR introduces policy groups.

Package preview: https://github.com/mtojek/package-spec/tree/132-support-policy-groups/test/packages/input_groups 
_Don't focus on the `policy_groups` name, it's a working name for the test package, in reality it will be `aws`, `nginx`, etc._

## Why is it important?

We need to ensure better granularity for policies, e.g. for the AWS integration.

## Checklist

## Related issues

- https://github.com/elastic/package-spec/issues/132


test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,119 @@
...
+      - src: /img/aws-dynamodb-overview.png
+        title: AWS DynamoDB Overview
+        size: 1702x1063
+        type: image/png
+    vars:
+      - name: dynamodb_foobar_property
+        type: bool
+        title: DynamoDB Specific Property
+policy_groups:
+  - name: ec2_logs
```
sorantis, line 106: `policy_groups` sounds like a good name, but why do we need them to be policy set specific? I was thinking they'd be generic groups that inputs can refer to.

```yaml
policy_groups:
  - name: logs
    title: "Logs" 
  - name: metrics
    title: "Metrics"
....

policy_templates:
  - name: ec2
    title: AWS EC2
    description: Collect logs and metrics from EC2 service
    inputs:
      - type: s3
        title: Collect logs from EC2 service
        description: Collecting EC2 logs using S3 input
        policy_group: logs
        vars: ...
      - type: aws/metrics
        title: Collect metrics from EC2 service
        description: Collecting EC2 metrics using AWS CloudWatch
        policy_group: metrics

```

sorantis: @mtojek @ruflin @ycombinator @kaiyan-sheng

mtojek: Sure that's configurable. I wanted to highlight that it's possible to have even 2 different logs groups. Let me adjust it.

mtojek: Fixed

sorantis: the input should also be updated to reflect the fix :)

sorantis:
I think we're getting close! We also would like each policy set to be part of one or more categories, i.e. if the user browses the Database category, they should be able to see DynamoDb tile too, so it should be possible to define categories for each policy set. E.g.
```yaml
policy_templates:
  - name: dynamodb
    title: AWS DynamoDB
    description: Collect logs and metrics from DynamoDB service
    categories: [AWS, Cloud, Database]
```

mtojek:
> I think we're getting close! We also would like each policy set to be part of one or more categories, i.e. if the user browses the Database category, they should be able to see DynamoDb tile too, so it should be possible to define categories for each policy set. E.g.

I think we can do this. It would require also a slight modification in the package-registry to cover this change. We have a predefined set of categories that are supported: https://github.com/elastic/package-spec/blob/master/versions/1/manifest.spec.yml#L61 (enum).

sorantis:
@mtojek who's curating this list?

mtojek:
Here is a thread in which we considered introducing another category: https://github.com/elastic/package-registry/pull/671#issuecomment-762328923

You can see all people taking part.

mtojek:
I think I've covered the entire proposal with schema, marking it as ready for review.

jen-huang:
Tagging myself to do some exploratory testing and requirements gathering on Kibana side. Thanks for putting the changes in PR, this makes it much easier.

test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,115 @@
...
+    screenshots:
+      - src: /img/aws-dynamodb-overview.png
+        title: AWS DynamoDB Overview
+        size: 1702x1063
+        type: image/png
+    vars:
+      - name: dynamodb_foobar_property
+        type: bool
+        title: DynamoDB Specific Property
+policy_groups:
```
sorantis, line 103: we should call it something else, since effectively we're grouping inputs, the name should reflect that.

sorantis: we're still stuck on aligning names, so apologies in advance if such comments may disrupt this PR. But we're getting there.

kaiyan-sheng: How about we keep `groups` here?

test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: ec2_foobar_property
+        type: bool
+        title: EC2 Specific Property
+  - name: dynamodb
+    title: AWS DynamoDB
+    description: Collect logs and metrics from DynamoDB service
+    categories:
+      - datastore
+    inputs:
+      - type: s3
```
kaiyan-sheng, line 81: We don't have dynamodb logs right now so should we remove this log section here?

kaiyan-sheng:
@mtojek Will we still have a `data_stream` directory at the [top level](https://github.com/mtojek/package-spec/tree/132-support-policy-groups/test/packages/policy_groups) with `docs`, `img` and `manifest.yml`?

test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,111 @@
...
+    size: 1702x1063
+    type: image/png
+vars:
+  - name: access_key_id
+    type: text
+    title: Access Key ID
+  - name: secret_access_key
+    type: text
+    title: Secret Access Key
+policy_templates:
```
ycombinator, line 33: @mtojek is this the array from which Kibana only shows the first element today? Or is that something different?

test/packages/policy_groups/manifest.yml
ycombinator: I'm -1 for calling it just `groups` as that's rather generic; I'd prefer to call it something more descriptive and/or specific.  

Maybe `input_groups` per @sorantis [comment](https://github.com/elastic/package-spec/pull/137/files#r579334776)?

Yay, naming 😄.

ycombinator:
> Will we still have a `data_stream` directory at the top level with docs, img and manifest.yml?

I have the same question. I think we still need this otherwise where else would the ingest pipelines, field definitions, etc. for each data stream be stored? @mtojek if you agree, could you add a couple of data stream folders to this PR please so we can see the fuller picture of the package? Thanks!

test/packages/policy_groups/manifest.yml
ycombinator: Or maybe @jen-huang can weigh in on my question since she's already watching this PR. 🙂

ycombinator:
@sorantis Clarifying question: based on the proposed package structure in this PR, which element in the package's `manifest.yml` would you like to see as separate tiles when a user searches for "aws" (besides the "main" AWS tile itself)?

test/packages/policy_groups/manifest.yml
mtojek: This is just a sample spec showing format opportunities and different cases, I suppose we can leave it here :) But you're right, Kaiyan, the AWS manifest in Integrations will have a different form (real configuration).

mtojek:
@kaiyan-sheng @ycombinator Yes, data streams stay as is :) I will add them to the package to get rid of confusion.

test/packages/policy_groups/manifest.yml
kaiyan-sheng: Sounds good to me :) Thanks for using AWS here for the example!

test/packages/policy_groups/manifest.yml
jen-huang: yes, this is what Kibana uses the first element of today

jen-huang:
(I'll be OOO next week so wanted to give this at least a quick look-over.) The changes so far look good from a Kibana perspective, there is work that needs to be done to support multiple `policy_templates` and the new `policy_groups` but I'm not spotting any obvious breaking changes. I assume that there should be no changes to the agent YAML that Kibana generates?

++ on showing data streams here for clarity

test/packages/policy_groups/img/aws-dynamodb.svg
```diff
@@ -0,0 +1 @@
...1 7.47-24.05 20.31-34.55 46.93-7.7 20.32-9.11 22.18-15.64 23.58-3.97.7-39.22 1.4-78.45 1.4-67.94 0-71.91.23-81.72 4.9C7.54 596.8 1.94 631.82 22.49 651.9c13.08 12.84 17.04 13.31 88.72 13.54 106 0 125.38-3.04 145.69-24.05l7.94-8.17 6.3 13.54c12.14 25.45 28.72 38.76 51.6 41.56 24.52 2.8 42.49-10.97 56.03-43.19 13.31-31.98 39.93-147.56 63.28-272.93 5.6-29.88 10.51-54.4 11.21-54.4.7 0 26.85 140.09 58.14 311.22 31.28 170.91 58.37 314.73 60.24 319.17 4.2 10.51 9.11 15.87 19.85 21.25 18.45 9.57 43.19 3.04 54.4-14.01 5.14-7.71 7.24-16.34 13.07-57.67 12.36-85.22 33.84-204.06 36.87-204.06.7 0 4.67 5.37 8.4 11.91 14.48 24.75 37.82 38.06 66.54 38.29 29.18 0 40.63-9.34 72.15-58.84l16.11-25.21 57.2-1.17c56.5-1.17 57.44-1.17 63.27-6.77 7.94-7.47 11.44-19.15 10.27-34.09-1.4-15.88-8.17-28.72-20.08-37.12l-9.57-6.77-57.43-1.17c-69.58-1.4-77.51-.23-94.33 14.94-6.3 5.84-17.74 19.84-24.98 31.29-7.47 11.44-13.78 20.78-14.01 20.31-.24-.23-2.57-12.61-5.37-27.32-8.64-48.8-19.38-69.81-40.86-80.55-22.41-11.21-48.33-6.31-64.21 11.91-14.47 16.34-30.12 56.03-43.66 110.67-3.5 14.47-6.77 25.68-7.24 24.52-.23-1.4-26.38-142.66-57.9-314.03-63.04-342.74-58.6-323.83-78.21-333.87-10.96-5.61-28.71-6.08-38.51-.71z"/></svg>
```
ruflin, line 1: Nit: Do you need the `aws` prefix here? I would assume if not, it would automatically match the group name?

test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,111 @@
+format_version: 1.0.0
+name: aws
+title: AWS
+version: 0.3.17
+license: basic
+description: AWS Integration
+type: integration
+categories:
```
ruflin, line 8: @sorantis Could have a group/set its own categories? Is it always a subset of the package categories?

test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,111 @@
...
+        type: bool
+        title: EC2 Specific Property
+  - name: dynamodb
+    title: AWS DynamoDB
+    description: Collect logs and metrics from DynamoDB service
+    categories:
+      - datastore
+    inputs:
+      - type: s3
+        title: Collect logs from the DynamoDB service
```
ruflin, line 82: I was first thinking if we should have a global way to specify inputs but this here shows nicely that each config of the input has a different title / description as it is specific to this case. So this is great.

test/packages/policy_groups/manifest.yml
```diff
@@ -0,0 +1,111 @@
...
+    screenshots:
+      - src: /img/aws-dynamodb-overview.png
+        title: AWS DynamoDB Overview
+        size: 1702x1063
+        type: image/png
+    vars:
+      - name: dynamodb_foobar_property
+        type: bool
+        title: DynamoDB Specific Property
+policy_groups:
```
ruflin, line 103: Ignoring naming on my end for now ;-)

I'm wondering if the groups should be defined as part of each policy_template or global. Having it local would allow to give more descriptive descriptions. At the same time, I don't think we use it today. Global requires much fewer configuration.

An alternative option is to just not have it in the package at first. Only 2-3 groups are supported, required by the spec and hardcoded into Kibana. Like this everyone is forced to use `logs/metrics` as groups.

ruflin (commented):
This looks promising. @mtojek Could you extend the example with one case where a single policy_template with a data_stream has more than 1 input agent template? I think there are a few packages which have it.

test/packages/policy_groups/img/aws-dynamodb.svg
mtojek: This is not a prefix for the `aws` integration, just a "random" string :) I think I'm in favor of consistency with other icon definitions in package manifest where there is no magic around file prefixes/package/data stream - just a simple file reference. It might be not straightforward for users to guess why some icons are skipped/filtered.

test/packages/policy_groups/manifest.yml
mtojek: Actually your both options are related. I agree that global definitions would require less configuration properties and might be easier to setup. The second option is an interesting solution - we can try to kick off the feature with predefined groups and expose only them to users. We need to remember that we won't escape from these tabs in the future :)

mtojek:
@ycombinator @jen-huang I've imported data streams definitions for AWS EC2 and DynamoDB

@ruflin I've added `spamfirewall` (Barracuda) data stream, which contains multiple agent templates (hbs).

sorantis:
> @sorantis Clarifying question: based on the proposed package structure in this PR, which element in the package's `manifest.yml` would you like to see as separate tiles when a user searches for "aws" (besides the "main" AWS tile itself)?

I think each entry in `policy_templates` should be a tile. I.e. if the user searches for "aws", the search results should include all supported AWS services, even the ones from a different package, like AWS Fargate.

test/packages/policy_groups/manifest.yml
sorantis: In some cases it's a bigger set, i.e. DynamoDB is an `aws` service, a `cloud` service and a `datastore` service. `datastore` here is specific to DynamoDB and will not be applicable to other, for example compute related services.

sorantis:
FYI, I updated the document to reflect the definitions we're using in the `manifest.yml` + it's more aligned with security. 
Changes: 
`integration instance` -> `integration policy`
`integration policy set` -> `integration policy template`
`integration policy` -> `input group` (or data group, or TBD. Basically it's the Logs/Metrics groups)

test/packages/policy_groups/img/aws-dynamodb.svg
ruflin: 👍  This also has the advantage that the same icon can be reused ;-)

test/packages/policy_groups/manifest.yml
ruflin: The question is what happens if someone clicks on a category, what will the filter return:

* All integration sets + package
* Only package
* Only integration sets

Based on the above we should know if we need categories inside the integration sets / policy templates.

test/packages/input_groups/manifest.yml
```diff
@@ -0,0 +1,132 @@
...
+        type: image/png
+    vars:
+      - name: dynamodb_foobar_property
+        type: bool
+        title: DynamoDB Specific Property
+  - name: barracuda
+    title: Barracuda logs
+    description: Collect Barracuda logs from syslog or a file.
+    inputs:
+      - type: udp
```
ruflin, line 107: Do we pull in here the additional vars from the data_stream manifest.yml? How do we know which one to select? What if there are multiple data_streams which configure `udp`? I remember I asked the question below but don't remember the answer :-(

mtojek: > Do we pull in here the additional vars from the data_stream manifest.yml?

I think so.

mtojek: > How do we know which one to select? What if there are multiple data_streams which configure udp? 

I don't reckon if we came up with a solution. Do you think we should introduce a "policy_templates" field to data stream manifest or do you see other options?

test/packages/policy_groups/manifest.yml
ycombinator: I'd prefer to start with the most restrictive option we can get away with and then make things less restrictive / more flexible depending on actual usage. So I'm ++ for hardcoding the group names (`logs` or `metrics`) for now. 

How far do we take this hardcoding? I assume not just the group `name`s are hardcoded but also the `title`s — "Logs" or "Metrics". What about the descriptions — should those be somewhat generated, e.g. "Logs for $package_name package"?

test/packages/input_groups/manifest.yml
ycombinator: Yes, my understanding is also that such a problem (or is it a feature? 😎) already exists. I'm wary of changing this behavior, unless we can keep the _default_ behavior the same as it is today.

ycombinator:
> I think each entry in `policy_templates` should be a tile. I.e. if the user searches for "aws", the search results should include all supported AWS services, even the ones from a different package, like AWS Fargate. 

So we will need to decide (most likely outside the scope of this PR) whether the package registry's search API response represents a list of packages (status quo) or a list of packages + policy templates.

Also, I'm not sure how to make this part work under the current proposal:

> ... even the ones from a different package, like AWS Fargate. 

Just to get the obvious question out of the way first: is there a reason AWS Fargate is it's own package and not a data stream under the `aws` package?

test/packages/policy_groups/manifest.yml
mtojek: Renamed to `input_groups`.

test/packages/policy_groups/manifest.yml
mtojek: @sorantis Any thoughts about hardcoding group names (logs, metrics, later on - checks)?

test/packages/input_groups/manifest.yml
mtojek: What do you think about putting this is in a separate PR if this became an issue? We know that it can be solved with an additional field - `policy_templates` in `data_streams`.

test/packages/policy_groups/manifest.yml
mtojek: Yeah and extending to what @ruflin has said, it's more natural to keep DynamoDB, EC2 and others as separate packages ;)

mtojek:
I think we should move towards the final decision about this pull request. Apart from:
* category vs subcategories for policy templates
* hardcoded input groups

do you see any other concerns around this PR?

sorantis:
> Just to get the obvious question out of the way first: is there a reason AWS Fargate is it's own package and not a data stream under the `aws` package?

I guess it's due to the different configuration options and API endpoints, but @kaiyan-sheng will know more.

> I think we should move towards the final decision about this pull request. Apart from:
> 
> * category vs subcategories for policy templates
> * hardcoded input groups
> 
> do you see any other concerns around this PR?

I would avoid hardcoding input groups, one of the use cases we had in mind is to distinguish between operational and security logs (for setting different retentions). If we hardcode the input groups now, it will be hard for us to have this flexibility.

mtojek:
> I would avoid hardcoding input groups, one of the use cases we had in mind is to distinguish between operational and security logs (for setting different retentions). If we hardcode the input groups now, it will be hard for us to have this flexibility.

I'm not sure if this PR assumes retention adjustments. It seems that input groups are more like UI tabs. Data retention is related rather to data stream configuration. Do we want to include retention settings in this PR or follow ups when necessary? I would vote for iterative approach, keep in mind that the first candidate is the AWS integration.

test/packages/policy_groups/manifest.yml
sorantis: If the user click on say "Datastore" then they should be able to see all integration that cover databases, data stores, etc. which also includes cloud services. Since tiles are (will be) generated from policy_templates then I think it's the option 3 in our list. 
This creates another question though. What happens if the user searches for AWS only, do we show AWS as a tile? The answer yes if we create a policy template (integration set) that is called AWS. 

This should work automatically for other integrations, for example NGINX. When the user searches for Nginx we show the Nginx tile because there's a policy template for it, we don't have to use package level information for it.
Screen Shot 2021-02-23 at 11 25 10

sorantis:
@mtojek fair enough. In my mind input groups serve two purposes: 1) separate config options on tabs; 2) set different retentions at the input group level. The question is then how would the user set different retentions for operational vs. security logs, if it's not the input groups that define such separation? I think if don't hardcode input groups it will be easier for the UI to reuse them for setting retention policies.

mtojek:
>  I think if don't hardcode input groups it will be easier for the UI to reuse them for setting retention policies.

I don't have a strong opinion on that, I suppose it's rather a UI design issue (@ruflin and @jen-huang can speak for Kibana). You brought here "operational logs" and "security logs", which means that it already violates (or extends?) the default split (logs-metrics-checks). If this is the approach we'd like to follow, let's stick to user defined input groups.

test/packages/input_groups/manifest.yml
ycombinator: Yes, I'm ++ to taking this specific discussion to a follow up issue/PR.

kaiyan-sheng:
> Just to get the obvious question out of the way first: is there a reason AWS Fargate is it's own package and not a data stream under the `aws` package?

Yep @sorantis is right. In order to collect metrics from `awsfargate`, you have to deploy metricbeat/agent inside the same task as the ECS containers you want to monitor. It does not require AWS credentials (access key, secret key).

ycombinator:
> Yep @sorantis is right. In order to collect metrics from `awsfargate`, you have to deploy metricbeat/agent inside the same task as the ECS containers you want to monitor. It does not require AWS credentials (access key, secret key). 

Thanks @kaiyan-sheng.

Thinking about it some more, I think the proposal in this PR is unrelated to the _italicized_ part of the following use case:

> I think each entry in `policy_templates` should be a tile. I.e. if the user searches for "aws", the search results should include all supported AWS services, _even the ones from a different package, like AWS Fargate_. 

The proposal in this PR will cover the non-italicized part. 

As a follow up we can figure out how to make the Package Registry's search API return the `aws` package, each of its `policy_templates`, and other related packages like `AWS Fargate` when the user searches for the `aws` package. The solution might involve relaxing the matching strictness in the Package Registry's search API code OR introducing additional search keywords at the package / policy_templates level OR something else. Either way, I'm good with deferring this discussion to a follow up issue/PR. 

My main concern was that nothing being introduced in this PR would later get in the way of solving for the italicized part of the above use case, and I think we are good on that front.

test/packages/input_groups/manifest.yml
ruflin: The problem does not exist today because we only have 1 policy template so we need all inputs. But with multiple policy templates, not all data_streams are needed for each one. So I think we need to solve it here.

mtojek: I'm looking for propositions/ideas. I've mentioned one idea above.

ruflin: I think both ways you proposed are feasible. I'm leaning in the direction of specifying it in the policy template. Let me explain why.

Lets assume I built a policy_template for a use case. Now I decide, I also want to allow users to configure data_stream x. All I need to do, is update my policy template and add an entry for it.

Having it in the policy template also makes it easy to understand, which data streams are used. I don't need to go to each data stream to check if the policy is mentioned in there.

mtojek: I'm afraid it (_specifying it in the policy template_ ) would violate the open-closed principle, but we can turn a blind eye here. Let me try to introduce it.

ruflin: Just a proposal, happy to discuss other solutions :-)

mtojek: I added the `data_streams` property. Actually I'm fine with both solutions, I wouldn't say it's a concept hard to understand :)

ycombinator: I'm good with the proposed solution with two comments:

1. **Default behavior**: not specifying `data_streams` in a policy template needs to be interpreted as if all data streams have been specified. This would guarantee backwards compatibility. I assume such default behavior logic will need to go in the Fleet UI code?

2. **Referential integrity**: We'll need a way to validate that data streams specified under `data_streams` actually exist in the package.

Neither comment above blocks this PR; just things to address in follow up PRs.

mtojek: @ycombinator do you have any preference where `data_streams` proporty should be placed, either in `policy_template` or `input`? For reference: https://github.com/elastic/package-spec/commit/ed04c0f3a449872d3f281ea250c921b340a43668#commitcomment-47498856

test/packages/policy_groups/manifest.yml
sorantis: Speaking of how to express categories in the manifest file, I suggest we keep the top level categories section required, and introduce an optional categories section to a policy template.
This way each policy template will inherit top level categories and also has the option to add more policy template specific categories if needed.
Thoughts? @mtojek @ycombinator @ruflin

mtojek: In other words, categories on the root level (in manifest) would be common buckets valid for all policy templates, right?

sorantis: yes, with the option to extend this list for each policy template. I.e. for AWS common buckets could be `cloud`, `aws` and DynamoDB could optionally extend this list with `datastores` by defining it at the policy template level.

test/packages/policy_groups/manifest.yml
sorantis: responded here https://github.com/elastic/package-spec/pull/137#issuecomment-784100742

mtojek:
We had an online discussion around the left issues and I think we came to the agreement:
* input groups stays as is
* policy template defines additional property `data_streams` to select particular data streams that are compatible with a policy template
* top categories are common, categories on the policy template level extend them

@ycombinator @ruflin please approve/comment on the PR

@jen-huang please double check if the PR is fine from Kibana dev perspective and also approve/comment here

ycombinator (approved):
LGTM. Really nice work iterating on this proposal and driving consensus!

I know this proposal will impact `kibana` and `package-registry` code at minimum.  I think it will also need a follow up PR in `package-spec` to enforce referential integrity between `data_streams` mentioned in policy template items and folders under `<package root>/data_stream/`. I don't think it affects `elastic-package` code (but I could be wrong about that). And, of course, eventually we'll want to introduce the enhancements enabled by this proposal to packages in the `integrations` repo.

So right before we merge this PR, could we create a checklist in https://github.com/elastic/package-spec/issues/132 that will capture each of the above tasks in an ordered fashion (topological sort on the dependency graph)? This will ensure that we can safely rollout this change and have a single place to track progress on the rollout. Some examples of such rollout checklists can be found in https://github.com/elastic/package-spec/issues/106 and https://github.com/elastic/package-spec/issues/17.

sorantis (approved):
Great to see this moving closer to the finish line! Thank you all for your involvement.

Agree with @ycombinator on creating a list of changes that need to be covered by related components. From the priority perspective, I'd like the focus be on Phase 1 described in the doc.

mtojek:
@jen-huang could you please double check if these spec changes are fine for Kibana?

jen-huang (approved):
The changes look good to me from a Kibana perspective.

I would like to do more thorough validation by testing it with a local Kibana instance but that would need package registry changes so that the new fields are served through the package info endpoint. I saw that the [registry changes item](https://github.com/elastic/package-spec/issues/132#issuecomment-788007017) calls for checking with AWS package changes, but if it would be faster to get a draft PR up with a test package (even this nice `input_groups` package done here 😉), that would be super helpful.

And just to double check: we expect Kibana to be backwards compatible for packages without input groups too, right?

mtojek:
> And just to double check: we expect Kibana to be backwards compatible for packages without input groups too, right?

Right, I would say this is an extension to current packages.

PR #141, opened 2021-02-25
Support static tests

mtojek:
## What does this PR do?

This PR adds support for static tests.

## Why is it important?

Static tests should be skipped, so we have to define a test config for them.

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/issues/226


ycombinator (approved):
LGTM.

PR #142, opened 2021-02-25
Add kibana/security-rule asset type

rw-access:
cc @FrankHassanabad, @spong, @brokensound77, @kevinlog  

## What does this PR do?
* Add a `security-rule` asset type for rules that will be delivered to Security » Detection Engine » Detection rules part of Kibana.

## Why is it important?
This is a dependency for:
* https://github.com/elastic/package-storage/pull/843
* https://github.com/elastic/kibana/pull/91949

## Checklist
**Note**: I haven't added a test package yet, because I'm expecting requested changes. Once we converge on the spec, I think it makes sense to add a test package.

- [ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.

## Related issues
No related public issues, just these PRs

- https://github.com/elastic/package-storage/pull/843
- https://github.com/elastic/kibana/pull/91949


versions/1/kibana/spec.yml
```diff
@@ -65,3 +65,25 @@
         type: file
         contentMediaType: "application/json"
         pattern: '^.+\.json$'
+    - description: Folder containing rules
```
mtojek, line 68: Please create a sample package and add it to the `test/packages`.

versions/1/kibana/spec.yml
```diff
@@ -65,3 +65,25 @@
...
+      type: folder
+      name: "rules"
+      required: false
+      contents:
+      - description: An individual rule file for the detection engine
+        type: file
+        contentMediaType: "application/json"
+        pattern: '^{PACKAGE_NAME}-rule-.+\.json$'
+      - description: Changelog for all of the rules in the folder
```
mtojek, line 77: Is it really necessary to introduce additional changelog for a package? Is it possible to use the general one for the package?

If so, could you please write a spec for this file? I suppose it's possible contrary to the rules files (as for other Kibana objects and multiple schemas).

versions/1/manifest.spec.yml
```diff
@@ -47,8 +47,10 @@ spec:
       type: string
       enum:
       - integration
+      - rules
```
mtojek, line 50: I'm hesitant to introduce a new package type this way (silently), also `rules` sounds generic to me. So far the integration type is the only supported one and I'm not sure what would be implications of introducing this.

Any thoughts, @ruflin?

versions/1/kibana/spec.yml
```diff
@@ -65,3 +65,25 @@
...
+        type: file
+        contentMediaType: "application/json"
+        pattern: '^{PACKAGE_NAME}-CHANGELOG\.json$'
+    - description: Folder containing timeline templates
+      type: folder
+      name: "timelines"
+      required: false
+      contents:
+      - description: A Kibana timeline template
```
mtojek, line 86: I'm not aware of differences between rules and timelines, but according to the description they're bound together? Is the `timeline` a Kibana object type?

mtojek (changes requested):
Please make sure that the PR passes the CI verification. Quick check:

```
make update && make check && make test
```

ruflin:
@rw-access Are rules and timeline tied together? Could a user just ship a rule or just a timeline?

versions/1/manifest.spec.yml
ruflin: I think what we need to discuss here is how these packages should show up in the package Manager.

@rw-access I would take this to a separate PR.

versions/1/kibana/spec.yml
```diff
@@ -65,3 +65,25 @@
         type: file
         contentMediaType: "application/json"
         pattern: '^.+\.json$'
+    - description: Folder containing rules
+      type: folder
+      name: "rules"
```
ruflin, line 70: Is `rules` the saved object type? Thinking out loud: Is this `security-rules`? Lets stick to the SO type.

versions/1/changelog.yml
```diff
@@ -79,4 +79,6 @@
   - description: Deprecate JSON-formatted pipeline test configs in favor of YAML-formatted ones
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/139
-
+  - description: Add rules package type with kibana/rules and kibana/timelines
```
ruflin, line 82: Are the two tied together? Otherwise I would split rules and timeline in two separate PR's.

versions/1/manifest.spec.yml
rw-access: For this package type, EPR is essentially used for storage but not much else.

Here's some ways where they differ:
- Detection rules packages aren't installed in the traditional manner via the Fleet pages,
- The rules in the package are always "installed" to a specific space
- Detection rules packages shouldn't show up in the Fleet Integrations page and they should be hidden. `internal: true` might be a better way to accomplish this than a new type

versions/1/kibana/spec.yml
rw-access: Some rules have dependencies on timelines, and right now in Kibana, there is an interrelationship between the two.

In Kibana today, we compile both into the release and from the Detection Engine page, users currently have one button to separate both. Maybe "detection_engine" is a better name for the package, since rules, timelines and ML jobs are all relevant and coupled.



@FrankHassanabad, @spong and @andrew-goldstein could add more detail.

versions/1/manifest.spec.yml
rw-access: Since these packages have no relationship with Agent and would be hidden from the Integrations page, I didn't think `type: integration` was appropriate. Something like `type: artifact` seems more appropriate and still generic

versions/1/kibana/spec.yml
ruflin: Lets go with a single global changelog.

versions/1/manifest.spec.yml
ruflin: This is helpful. Do you expect that rules can also be added for example to an `nginx` package for nginx specific rules in the future? What happens if someone adds rules to a package?

versions/1/kibana/spec.yml
rw-access: Removed this asset type, descoping the concept of a rule-wide changelog for now

versions/1/manifest.spec.yml
rw-access: We had a lot of discussions, and the `rules` custom type for a package isn't necessary. Thanks for helping me get there, @ruflin.
Updated the PR so this thread is no longer relevant

versions/1/kibana/spec.yml
rw-access: Went with `security-rule`. These folder names are typically in singular form right?

versions/1/changelog.yml
rw-access: Done. Removed timelines from the scope of this PR

versions/1/kibana/spec.yml
ruflin: yes. And I think in most cases we use `_` and not `-`. @mtojek ?

ruflin:
@rw-access Could you add a sample package for testing this?

versions/1/kibana/spec.yml
rw-access: I added this asset to `test/packages/good`

rw-access: Unless this should be it's own test package

versions/1/kibana/spec.yml
rw-access: That looks right, updated to `security_rule`

versions/1/kibana/spec.yml
```diff
@@ -65,3 +65,12 @@
...
         pattern: '^.+\.json$'
+    - description: Folder containing rules
+      type: folder
+      name: "security_rule"
+      required: false
+      contents:
+      - description: An individual rule file for the detection engine
+        type: file
+        contentMediaType: "application/json"
+        pattern: '^rule-.+\.json$'
```
ruflin, line 76: Is the `rule` prefix in the name required?

ruflin (commented):
For later: Should we do on the package side an validation of the rules file? 

@ycombinator do we apply some automatic formatting to it?

versions/1/kibana/spec.yml
rw-access: no, I can loose this to just `*.json`

rw-access: Just `^.+\.json$` as of ba237067d01321af83c722ce03fb9fb93a312a81

ycombinator:
> For later: Should we do on the package side an validation of the rules file?

The package spec does allow for defining the structure of file contents but we have deliberately not specified structures for other Kibana saved objects like dashboards, etc. because we're treating those as Kibana implementation details (at least for the moment). The only basic validation is that the files match a certain mime type and are (potentially) named a certain way, both of which are being done in this PR.

> @ycombinator do we apply some automatic formatting to it?

Automatic formatting to the JSON files? Yes, `elastic-package format` will take care of it.

ycombinator (approved):
LGTM.

mtojek (approved):

ruflin (approved):

PR #147, opened 2021-03-10
Introduce optional test results

mtojek:
## Why is it important?

This PR updates the package spec with optional test results in pipeline tests. It may happen that the ingest pipeline won't produce any data (used drop processor) and this is the expected output.

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/issues/284


versions/1/data_stream/_dev/test/pipeline/expected.spec.yml
```diff
@@ -5,11 +5,16 @@ spec:
...
+          additionalProperties: true
+        - type: "null"
   properties:
     expected:
       description: Expected test result
       type: array
       items:
-        type: object
-        additionalProperties: true
+        $ref: "#/definitions/optionalTestResult"
```
ycombinator, line 19: Nit: why not just inline the definition right here? Is it not possible or did you define it separately for some other reason?

ycombinator (approved):
Left a nitpick which you can ignore (I'm more curious about the thinking behind it). LGTM.

versions/1/data_stream/_dev/test/pipeline/expected.spec.yml
mtojek: Yes, it wasn't possible. It failed with a strange error (trying to figure out if it's array of strings, array of objects, but definitely not null)

PR #151, opened 2021-03-17
Add NOTICE.txt file at the root level of a package

### LINKED ISSUE #149: [Change Proposal] Optional NOTICE.txt file for integrations
Related to #142

Some parts of these integrations could included code licensed with permissive licenses (MIT, BSD3, Apache 2, etc.). Even though the final license of all our integrations is Elastic License, Elastic License 2.0 or SSPL, we still have to preserve the original license notices when copying the code or binaries around.

We should have a generic way of including a NOTICE.txt file for each integration and deliver that to Kibana with the rest of the assets.. The most relevant use case currently is the Detection engine rules, which currently copy https://github.com/elastic/detection-rules/blob/main/NOTICE.txt into Kibana in https://github.com/elastic/kibana/blob/master/x-pack/plugins/security_solution/server/lib/detection_engine/rules/prepackaged_rules/notice.ts.

rw-access:
## What does this PR do?
Closes #149

Adds a new optional NOTICE.txt file to a package. This isn't an asset but is the NOTICE.txt for the whole package.

## Why is it important?
Some assets can be derived from licensed material and require us to propogate NOTICE.txt. The current use-case is for the `security_rule` asset type, because some rules (basically query + metadata) have been derived from license materials.

We will need a way to present this information to the user. I imagine the specific integration page is sufficient.

## Checklist
- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective. **I added an asset to the "good" package**

## Related issues
#142 (PR)
https://github.com/elastic/package-storage/pull/843 (PR)


ruflin (approved):
LGTM. But lets way on a 👍 from @ycombinator or @mtojek .

@rw-access As soon as we get it in, could you follow up with an issue in Kibana to make sure we also add the logic in Kibana to expose it. Otherwise it will be pretty hard for any user to find it.

versions/1/spec.yml
```diff
@@ -19,6 +19,11 @@ spec:
     name: "changelog.yml"
     required: true
     $ref: "./changelog.spec.yml"
+  - description: The package's NOTICE file
+    type: file
+    contentMediaType: "text/plain"
```
mtojek, line 24: Is it necessary to define the content type here provided that no action is taken based on this? Did you try without this definition?

rw-access: I just made an assumption that this was necessary. I'm not intimately familiar with the details but am glad to remove if that is preferred.

ycombinator: I'm okay leaving this in. We may not take any action on this content type today but if we do in the future, the NOTICE file would automatically get covered.

ycombinator (approved):
LGTM.

rw-access:
Thanks @ycombinator and @ruflin!

PR #153, opened 2021-03-23
Define version pattern in a single place

ycombinator:
## What does this PR do?

This PR defines the pattern for a version in a single place and references it from multiple version properties.

## Why is it important?

So the pattern for a version doesn't need to be duplicated in multiple places, introducing a potential for drift and inconsistencies.

## Checklist

- [ ] ~I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.~ Existing packages have versions in them already.
- [ ] ~I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml).~ No changelog entry as this is a refactoring, not a user-facing change.


mtojek (approved):
LGTM

PR #157, opened 2021-04-06
Consistency in names and descriptions of input groups

ycombinator:
## What does this PR do?

Renames all references to policy groups to input groups.

## Why is it important?

Consistency. Before this PR sometimes we called input groups as input groups and sometimes we called them as policy groups in the spec, which can be confusing.

## Checklist

- [ ] ~I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.~


versions/1/manifest.spec.yml
```diff
@@ -170,27 +170,26 @@ spec:
...
         properties:
           name:
-            description: Name of policy group.
+            description: Name of input group.
             type: string
             examples:
-              - Logs
-              - Metrics
+              - logs
+              - metrics
```
ycombinator, line 184: I think we want to treat the `name` property like an ID since it will be referenced from within `policy_templates`. So I think lowercasing makes sense here.

versions/1/manifest.spec.yml
```diff
@@ -170,27 +170,26 @@ spec:
...
           title:
-            description: Title of policy group.
+            description: Title of input group.
             type: string
             examples:
               - Logs
               - Metrics
-              - Checks
```
ycombinator, line 191: I'm not sure why we included this example, as there doesn't seem to be a corresponding `name` or `description` example for it. So I think we should remove it.

mtojek: I think the "examples" field is not an enum. It just gives samples the user can use. It doesn't mean it has to be consistent across examples for other field values.

mtojek (approved):

versions/1/manifest.spec.yml
ycombinator: I understand that it just gives sample values. I just think it's confusing if there isn't consistency between this field and the `name` field, especially since the first two examples here line up exactly with the ones in the `name` field. 

Maybe one solution might be to use examples here that are completely different from the `name` values? I think one related question is: how/where do we expect this `title` field to be used?

mtojek: > Maybe one solution might be to use examples here that are completely different from the name values?

It's a good idea, might be less misleading.

>  I think one related question is: how/where do we expect this title field to be used?

I see this as a safe workaround for multi-words names, let's say: magic_checks/multichecks vs "Magic checks".

ycombinator: Thanks, I will update the examples to be completely different from the `name` values and include some multi-word ones.

mtojek (approved):
Changes look good to me

Issue #47, opened 2020-09-17
[META] Semantic validations


ycombinator:
The package spec allows for syntactic validations of packages. These validate the _structure_ of package contents. They have been implemented via https://github.com/elastic/package-spec/pull/14, https://github.com/elastic/package-spec/pull/41, and several minor follow up PRs.

There is also a need for semantic validations. These validate that values found in package contents conform to further rules that may not be expressible in the spec itself. For example:

* [x] Validate that icon and screenshot sources exist: https://github.com/elastic/package-spec/issues/31
* [ ] Validate that dashboard IDs start with package names: https://github.com/elastic/package-spec/issues/65
* [ ] Ensure consistency between CHANGELOG and manifest: https://github.com/elastic/package-spec/issues/167

ruflin:
* Validation for ingest pipelines: https://github.com/elastic/package-registry/blob/master/util/dataset.go#L168

ycombinator:
Hey @ruflin, I'm starting to look into this issue now. Could you elaborate on your comment (https://github.com/elastic/package-spec/issues/47#issuecomment-694693842) please? The link in the comment is now dead so I can't figure out what you meant. Thanks!

ruflin:
If I remember correctly this was about validating that a pipeline name that was used in the manifest actually exists. Basically any file that can be referenced from the manifest file like icons, readmes, pipelines, etc. should be validated that it exists.

mtojek:
@ycombinator Would you mind updating the status of this issue? You have made a significant progress here recently.

ycombinator:
There is no direct update to this issue because of the recent progress. However, I updated https://github.com/elastic/package-spec/issues/65 which is linked from this issue.

Separately, I also added https://github.com/elastic/package-spec/issues/167 to the checklist on this issue.

jsoriano:
Checklist was completed, closing this. Please open new issues for additional validations.

PR #160, opened 2021-04-20
Semantic validations

ycombinator:
## What does this PR do?

This PR introduces semantic validations to the Package Spec. Semantic validations are validations of the package spec that cannot be achieved simply by adding declarative rules to the package spec, but instead may require some amount of programming to express the rules.

To demonstrate how semantic validations can be written, this PR [implements](https://github.com/elastic/package-spec/pull/160/files#diff-8ada00731e0da6d7feec20bf139d37220e0433f0373b3b005f33bbdb2bd2215b) the following semantic validation rule:

> Validate that the `id` fields inside Kibana object files have the same value as the object filename (minus the .json extension, of course)

A follow up PR will implement a second semantic validation rule:

> Validate that there are no dangling references to Kibana objects in any Kibana object files within the scope of a package. That is, if a Kibana object file in package `P` references another Kibana object with ID `i`, then there must exist a Kibana object file for id `i` in package `P`

## Why is it important?

To enforce semantic constraints on package elements.

## Checklist

- [ ] ~I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml).~ There is no change to the spec.

## Related issues
- Related #47 
- Related #65


code/go/internal/validator/semantic/kibana_matching_object_ids.go
```diff
@@ -0,0 +1,55 @@
...
+			idPath = "$.rule_id"
+		}
+
+		objectID, err := objectFile.Values(idPath)
+		if err != nil {
+			errs = append(errs, errors.Wrapf(err, "unable to get Kibana object ID in file [%s]", filePath))
+			continue
+		}
+
+		fileID := strings.Replace(filepath.Base(filePath), filepath.Ext(filePath), "", -1)
```
mtojek, line 47: nit: maybe add a comment here: fileID is the filename without extension. Few seconds of life saved :)

code/go/internal/validator/spec.go
```diff
@@ -54,5 +58,11 @@ func (s Spec) ValidatePackage(pkg Package) ValidationErrors {
 		return errs
 	}
 
-	return rootSpec.validate(pkg.Name, pkg.RootPath)
+	// Syntactic validations
+	errs = rootSpec.validate(pkg.Name, pkg.RootPath)
+
+	// Semantic validations
+	errs = append(errs, semantic.ValidateKibanaObjectIDs(pkg.RootPath)...)
```
mtojek, line 65: What do you think about transforming this line a bit. I'm thinking about the following form:

```go
var rules []RuleFunction
rules := append(rules, semantic.ValidateKibanaObjectIDs,
    semantic.ValidateKibanaFoo,
    semantic.ValidateKibanaFoobar,
    semantic.ValidateKibanaFoobaz,
)

err := rules.Validate(pkg.RootPath)
```

code/go/pkg/validator/errors.go
```diff
@@ -1,8 +1,8 @@
 package validator
 
 import (
-	"github.com/elastic/package-spec/code/go/internal/validator"
+	"github.com/elastic/package-spec/code/go/internal/errors"
 )
 
 // ValidationErrors is an Error that contains a iterable collection of validation error messages.
-type ValidationErrors validator.ValidationErrors
+type ValidationErrors errors.ValidationErrors
```
mtojek, line 8: I see that it will change the API, but as long as the source code is versioned we should be safe.

code/go/internal/validator/semantic/kibana_matching_object_ids.go
```diff
@@ -0,0 +1,55 @@
...
+		if fileID != objectID {
```
mtojek, line 48: oh, it doesn't complain about different type here? (string vs interface)

mtojek (commented):
I left a couple of questions/ideas in comments. The idea is great as it's much simpler than building a dedicated AST.

We need to remember that it will produce additional technical debt on porting the validator to other languages (but it's acceptable).

As this PR changes interfaces for package-spec, which approach is better:
1. Push this PR to master, then to elastic-package, and integrations (I assume you'll have to fix some packages) and then iterate on new rules.
2. Push this PR to master, add more rules (at least the TODO one) and then push everything further.

ycombinator:
> We need to remember that it will produce additional technical debt on porting the validator to other languages (but it's acceptable).

Agreed. This is why I created a new Golang package, `semantic`, which is where we should keep all these semantic validation implementations. That way, at least it's easy to find all the semantic validations in a single place when porting them over to another language.

BTW, I think we have a similar issue with the [format checkers](https://github.com/elastic/package-spec/blob/master/CONTRIBUTING.md#defining-property-format) implementation. Maybe we should move [all format checker code](https://github.com/elastic/package-spec/blob/0a25dbffb1b1d0222b6836652a103bd6fdc9da39/code/go/internal/validator/folder_item_spec.go) into this `semantic` package as well? WDYT?

code/go/internal/validator/semantic/kibana_matching_object_ids.go
ycombinator: Yes, this works as explained in https://golang.org/ref/spec#Comparison_operators:

> Interface values are comparable. Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value `nil`.
> A value `x` of non-interface type `X` and a value `t` of interface type `T` are comparable when values of type `X` are comparable and `X` implements `T`. They are equal if `t`'s dynamic type is identical to `X` and `t`'s dynamic value is equal to `x`.

code/go/pkg/validator/errors.go
ycombinator: Yeah, I had to do this to avoid an import cycle. Otherwise I would've preferred to keep the API the same.

If we really want to stay backwards compatible here, we could introduce type aliases like so:

```
package validator

import "github.com/elastic/package-spec/code/go/internal/errors"
  
type ValidationErrors = errors.ValidationErrors
```

However, I think this is overkill considering that, as far as we know, there is nothing that is using this type outside of the `package-spec` module. I checked `elastic-package` and even it does not use it; it simply relies on the `error` interface returned by `validator.ValidateFromPath` instead.

ycombinator:
> As this PR changes interfaces for package-spec, which approach is better:
> 1. Push this PR to master, then to elastic-package, and integrations (I assume you'll have to fix some packages) and then iterate on new rules.
> 2. Push this PR to master, add more rules (at least the TODO one) and then push everything further.

I think option 1 is better, especially since there may be fixes to some integrations packages as a result. That way the PR to `integrations` can not only update the `elastic-package` dependency but also have these fixes, but also remain relatively small in size. 

Later we can similarly make a second set of PRs for the second semantic validation rule (the one marked as TODO in this PR).

code/go/internal/validator/semantic/kibana_matching_object_ids.go
ycombinator: Added in a114b83.

code/go/internal/validator/spec.go
ycombinator: I implemented something like this in e0aa080. Let me know what you think.

code/go/internal/validator/semantic/kibana_matching_object_ids.go
mtojek: TIL!

code/go/pkg/validator/errors.go
mtojek: Definitely an overkill, let's stick to the current form. We know all consumers of this package.

mtojek (approved):
> BTW, I think we have a similar issue with the format checkers implementation. Maybe we should move all format checker code into this semantic package as well? WDYT?

SGTM, you can open an issue to refactor them too or modify in one next PRs (up to you).

ycombinator:
> SGTM, you can open an issue to refactor them too or modify in one next PRs (up to you).

Okay, I will do this in a follow up PR soon after this one is merged.

ycombinator:
@mtojek I made a non-trivial change to make CI green again on this PR: https://github.com/elastic/package-spec/pull/160/commits/4230c1a123210e4ef833ff71844f0954b571d557. Please re-review. Thanks!

mtojek (approved):

PR #161, opened 2021-04-21
Relocate format checkers

ycombinator:
## What does this PR do?

This PR relocates the format checkers implementation into the new `semantic` package.

## Why is it important?

With this change, all non-declarative package validation rules are contained in a single place, the `semantic` package. This will make porting such rules to another language easier.

## Checklist

- [ ] ~I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.~ Test packages already exist.
- [ ] ~I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml).~ No changes to spec.

## Related issues

- Follow up to https://github.com/elastic/package-spec/pull/160.

### DIFF
--- a/code/go/internal/validator/folder_item_spec.go
+++ b/code/go/internal/validator/folder_item_spec.go
@@ -8,17 +8,14 @@ import (
 	"regexp"
 	"sync"
 
+	"github.com/elastic/package-spec/code/go/internal/validator/semantic"
+
 	ve "github.com/elastic/package-spec/code/go/internal/errors"
 
 	"github.com/pkg/errors"
 	"github.com/xeipuuv/gojsonschema"
 
 	"github.com/elastic/package-spec/code/go/internal/yamlschema"
-)
-
-const (
-	relativePathFormat   = "relative-path"
-	dataStreamNameFormat = "data-stream-name"
 )
 
 type folderItemSpec struct {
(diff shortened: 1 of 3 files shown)


mtojek (commented):
Is everything alright with this PR? I see it includes changes introduced in https://github.com/elastic/package-spec/commit/92ee6e42f2e89456ae4ef5f66531abbc4a23ead4 . Is that intended?

ycombinator:
@mtojek Ah, good catch. I made this PR before #160 was merged. It needs a rebase on `master`. That's why I had left this PR in draft mode but I forgot about that and put it into review yesterday 🤦. Rebasing on `master` now...

code/go/internal/validator/folder_item_spec.go
```diff
@@ -8,6 +8,8 @@ import (
 	"regexp"
 	"sync"
 
+	"github.com/elastic/package-spec/code/go/internal/validator/semantic"
+
```
mtojek, line 12: nit: sorting imports

mtojek (approved):
It might be confusing for users that this semantic validator requires special handling - Load/Unload. I would be for adjusting it to the concept `func do() validationErrors`, but I don't think it's easy. Let's merge this one!

PR #164, opened 2021-04-27
Define Github owner field pattern

mtojek:
## What does this PR do?

This PR defines the Github owner pattern for the manifest field.

## Why is it important?

We need to make sure that the field accepts private users or company teams.

## Checklist


test/packages/bad_github_owner/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.0.1"
+  changes:
+    - description: Initial draft of the package
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/0 # FIXME Replace with the real PR link
```
ycombinator, line 6: Replace this with https://github.com/elastic/package-spec/pull/164?

ycombinator (approved):
LGTM.

test/packages/bad_github_owner/changelog.yml
mtojek: Not necessary, this package has been created with `elastic-package create package`. We can leave it here ;)

ycombinator: I know, just as a convention I've been using PRs where a test package was introduced as the link in that test package's changelog. You can see it in some other test packages. That way its easy to understand _why_ a test package was introduced (what was the context for it).

ycombinator: But this is not a hard requirement, just a convention. So if you want to leave this as-is, that's fine with me.

mtojek: Makes sense, let me update it quickly!

mtojek: Fixed

PR #170, opened 2021-04-30
Add overview on what a package is

ruflin:
This adds a high level overview on what a package actually is. I could not find a place where we have this documented yet in public.


ruflin:
Just found https://github.com/elastic/kibana/issues/88499#event-4657823567 which might help to expand this docs.

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
 ## Version Compatibility between Packages and Specifications
 
 A package specifying its `format_version` as `x.y.z` must be valid against specifications in the semantic version range `[x.y.z, X.0.0)`, where `X = x + 1`.
+
+
+# What is an Elastic Package?
+
+An Elastic Package is a collection of assets for the Elastic Stack. In addition, it contains manifest files which contain additional information about the package. The exact content and structure of a package are described by the package spec. In the following is a high level overview of a package.
```
ycombinator, line 68: Maybe add something about how packages also contain information about configuring them via the Fleet UI?

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
...
+
+A package is downloaded as a .zip file for the package-registry by Fleet inside Kibana. The zip file is then unpacked and each asset is installed into the related API.
```
ycombinator, line 70:
```suggestion
A package is downloaded as a .zip file from the package-registry by Fleet inside Kibana. The zip file is then unpacked and each asset is installed into the related API.
```

ycombinator: I think we can leave out the `as a .zip file` here. Feels like an implementation detail.

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
...
+
+## Asset organisation
+
+In general, the assets are organised by {stack-component}/{asset-type}. For example assets for Elasticsearch ingest pipelines are in the directory `elasticsearch/ingest-pipeline`. The same logic applies to all Elasticsearch, Kibana and Elastic Agent assets.
```
ycombinator, line 74: s/directory/folder/ for consistency with other docs in this repo.

```suggestion
In general, assets within a package are organised by {stack-component}/{asset-type}. For example assets for Elasticsearch ingest pipelines are in the folder `elasticsearch/ingest-pipeline`. The same logic applies to all Elasticsearch, Kibana and Elastic Agent assets.
```

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
...
+
+There is a special directory `data_stream`. All assets inside the data_stream directory must follow the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). `data_stream` can contain multiple directory, each with the name of the dataset (or postfix). Inside this directory, the same structure as before for {stack-component}/{asset-type} applies. The difference is that for all these assets, Fleet during installation enforces naming rules related to the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). All assets in this directory belong directly or indirectly to data streams.
```
ycombinator, line 76:
```suggestion
There is a special folder `data_stream`. All assets inside the `data_stream` folder must follow the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). `data_stream` can contain multiple folders, each with the name of the dataset (or postfix). Inside this folder, the same structure as before for {stack-component}/{asset-type} applies. The difference is that for all these assets, Fleet during installation enforces naming rules related to the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). All assets in this folder belong directly or indirectly to data streams.
```

ycombinator: What is a dataset? 😄 

If we are going to introduce new terminology (relative to the rest of the documentation here) and especially if that terminology sounds similar but isn't quite the same as other terminology (dataset vs. data stream), we need to explain what this new terminology is.

Otherwise it's going to setup readers for confusion. Some will assume dataset is the same as data stream. Others will just be left confused and start questioning their understanding of data streams, etc.

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
...
+
+In contrast, any asset added on the top level will be picked up as json document, pushed to the corresponding Elasticsearch / Kibana API and used as is. In most scenarios, only the data_stream assets are need, there are exceptions where global assets are required to get more flexibility.
```
ycombinator, line 78:
```suggestion
In contrast, any asset added on the top level will be picked up as json document, pushed to the corresponding Elasticsearch / Kibana APIs and used as is. In most scenarios, only data stream assets are needed. There are exceptions where global assets are required to get more flexibility.
```

ycombinator: Last sentence seems somewhat vague and not very actionable. Maybe give some examples of assets for that last sentence? Or examples of the type of flexibility?

If I'm a reader of the package spec documentation, I'm probably here to understand what goes where so I can do the right thing when building my package. Having concrete examples wil help with this.

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
...
+  * Visualization
+  * Index patterns
+  * ML Modules
+  * Map
+  * Search
+  * Security rules
+* Other
+  * fields.yml
+
+The special asset `fields.yml` is used to generate out of a single definition Elasticsearch Index Templates and Kibana index patterns. The exact definition can be found in the package spec.
```
ycombinator, line 100:
```suggestion
The special asset `fields.yml` is used to generate Elasticsearch Index Templates and Kibana index patterns. The exact definition can be found in the package spec.
```

README.md
```diff
@@ -61,3 +61,40 @@ As mentioned above, package specifications are versioned. Versions follow the [s
...
+
+There is a special directory `data_stream`. All assets inside the data_stream directory must follow the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). `data_stream` can contain multiple directory, each with the name of the dataset (or postfix). Inside this directory, the same structure as before for {stack-component}/{asset-type} applies. The difference is that for all these assets, Fleet during installation enforces naming rules related to the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). All assets in this directory belong directly or indirectly to data streams.
+
+In contrast, any asset added on the top level will be picked up as json document, pushed to the corresponding Elasticsearch / Kibana API and used as is. In most scenarios, only the data_stream assets are need, there are exceptions where global assets are required to get more flexibility.
+
+## Supported assets
+
+In the following each currently supported asset type is listed for a quick overview. The package spec will always contain the fully up-to-date list:
```
ycombinator, line 82:
```suggestion
For a quick overview, these are the assets typically found in an Elastic Package. The package spec will always contain the fully up-to-date list.
```

ycombinator (commented):
Thanks for adding this overview, @ruflin. Do you think we should move it up even higher in the README — probably makes sense to explain (at a high level) what a package is, before we go into documentation about the package spec?

ruflin:
I was thinking the same when writing: Should we put it on top, or at least parts of it? Happy to move it.

README.md
ruflin: I linked to the data stream naming scheme that should explain it. But nobody is going to jump there to read up on it first. At the same time I do not want to dive to deep into this overview. Any ideas for a way out here?

README.md
ruflin: Added

README.md
ruflin: Hm, not sure. I often explain a package as a glorified zip file.

README.md
ruflin: Modified it slightly to leave out dataset.

README.md
ruflin: Added an ILM policy example.

ruflin:
Thanks for the review @ycombinator . I modified the content a bit according to your review and move it to the top.

README.md
ycombinator: What I mean is that the fact that a package is "downloaded as a .zip file" is entirely incidental as far as a package author or package consumer (end-user) is concerned. If I'm either of these personas, why should I care how a package is encoded in transport between the package registry and Kibana? Also tomorrow if we decide to deliver packages as `.tar` files or something else, are we going to remember to update this location here?

README.md
```diff
@@ -11,6 +11,46 @@ This repository contains:
...
+
+There is a special folder `data_stream`. All assets inside the `data_stream` folder must follow the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). `data_stream` can contain multiple folders, each with the name of that describes the content. Inside this folder, the same structure as before for {stack-component}/{asset-type} applies. The difference is that for all these assets, Fleet during installation enforces naming rules related to the [Data Stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme). All assets in this folder belong directly or indirectly to data streams.
+
+In contrast, any asset added on the top level will be picked up as json document, pushed to the corresponding Elasticsearch / Kibana APIs and used as is. In most scenarios, only data stream assets are needed. There are exceptions where global assets are required to get more flexibility. This could be for example an ILM policy that applies to all data streams.
```
ycombinator, line 29: Thanks for adding the ILM example. Just copy editing:

```suggestion
In contrast, any asset added on the top level will be picked up as json document, pushed to the corresponding Elasticsearch / Kibana APIs and used as is. In most scenarios, only data stream assets are needed. There are exceptions where global assets are required to get more flexibility. This could be, for example, an ILM policy that applies to all data streams.
```

ycombinator (approved):
Made one more minor copy-editing suggestion. After that LGTM.

Thanks for writing this up; we need more "high level" introductory content like this!

ruflin:
Thanks for the detailed review @ycombinator and especially the suggest parts. Makes it so easy to get changes in.

Issue #167, opened 2021-04-28
Semantic validation: ensure consistency between version and changelog


mtojek:
There is a risk that developers won't update the CHANGELOG files, so there won't be consistency between version in manifest and changelog entries.

@ycombinator it looks like a great candidate for another validation rule :)

mtojek:
I'll try to add this rule.

Issue #140, opened 2021-02-24
[Change Proposal] Enforcing structure inside `changelog.yml` files


ycombinator:
## Problem statement

With https://github.com/elastic/package-spec/pull/131, packages are now expected to have a `changelog.yml` file in their root folders. That PR merely introduced the structure (syntax) for package changelogs. As a next step it would be good to introduce some semantics as well and try to enforce them.

There are two related questions to consider:

1. in the package's `changelog.yml` file, how should unreleased changes be stored?

2. what should be the relationship of the `version` specified in a package's `manifest.yml` file to any of the versions specified in the package's `changelog.yml` file? 

## Proposal

The structure of package's `changelog.yml` files is expected to be like so:

```yaml
- version: x.y.z
   changes:
   - description: ...
      type: ...
      link: ...
   - description: ...
      type: ...
      link: ...
- version: p.q.r
   - description: ...
      type: ...
      link: ...
   - description: ...
      type: ...
      link: ...
- version: a.b.c
   - description: ...
      type: ...
      link: ...
   - description: ...
      type: ...
      link: ...
```

* We introduce a special version, `next`, that is expected to be **first** entry in the top-level array in `changelog.yml` files (where `x.y.z` is in the snippet above). All the `changes` under this version will be unreleased ones.
* We expect that the **second** entry in the top-level array in `changelog.yml` files (where `p.q.r` is in the snippet above) matches exactly the value of `version` in the package's `manifest.yml` file.

This allows a package developer to accumulate unreleased changes in the changelog under the first entry (`next`). When the package developer is ready to release a new version of the package, they do the following, all as part of a single PR:
1. bump up the value of `version` in their package's `manifest.yml`,
2. replace `next` with the new value of `version` in their package's `manifest.yml`, and
3. add a new `next` entry at the top of their package's `changelog.yml` file with an empty list of `changes` under it.

ycombinator:
@mtojek @ruflin WDYT? 

Anyone else we should pull into this discussion?

mtojek:
I think I talked with @masci about this. Massi suggested to leverage from semver and introduce here th "prerelease" tag. We can tweak our scripts later on to check the relation between:
- version in manifest and prerelease in `changelog.yml`
- version in manifest and version in `changelog.yml`

> 1. bump up the value of version in their package's manifest.yml,
> 2. replace next with the new value of version in their package's manifest.yml, and
[...]

PR #172, opened 2021-05-06
Ensure version consistency between manifest and changelog

mtojek:
## What does this PR do?

This PR introduces additional validation to ensure consistency between manifest and changelog files.

## Why is it important?

This rule is required as developers tend to forget about updating changelog files.

## Checklist

## Related issues

- Fixes https://github.com/elastic/package-spec/issues/167


ycombinator:
I'm fine with the implementation the way it is, but if you want you can also use the [helpers from the `pkgpath` package](https://github.com/elastic/package-spec/blob/master/code/go/internal/pkgpath/files.go).

ycombinator:
I'm not sure if the validation rule should be that the **first** entry in `changelog.yml` should match the version in the `manifest.yml`. I was thinking it should more be the **second** entry. 

Additionally the first entry should be a special version like `next` or something where package developers can accumulate multiple changes for the next version.

I proposed something like this in https://github.com/elastic/package-spec/issues/140#issue-814976510 as well.

mtojek:
@ycombinator I transformed the source code to use `pkgpath` helpers and to check if the manifest version is referenced "somewhere" in the changelog.

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -0,0 +1,88 @@
...
+	f, err := pkgpath.Files(manifestPath)
+	if err != nil {
+		return "", errors.Wrap(err, "can't locate manifest file")
+	}
+
+	if len(f) != 1 {
+		return "", errors.New("single manifest file expected")
+	}
+
+	val, err := f[0].Values("version")
```
ycombinator, line 43: Doesn't this need to be `$.version`? Just `version` works?

ycombinator (approved):
Left a comment/question about `$.version` vs `version`. Otherwise, LGTM.

ycombinator (approved):
LGTM.

PR #174, opened 2021-05-12
Fix missing version

mtojek:
## What does this PR do?

This PR improves the version integrity check to prevent panic in case of missing/undefined version.

## Why is it important?

It prevents panics. 

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/pull/354


test/packages/missing_version/manifest.yml
```diff
@@ -0,0 +1,22 @@
+format_version: 1.0.0
+name: inconsistent_version
+title: "Missing Version"
+license: basic
+description: "This is a test package."
+type: integration
+version:
```
ycombinator, line 7: I'm surprised this passes validation, given https://github.com/elastic/package-spec/blob/ade1f5216c0d1320c83269f533b6531ec48ff85b/versions/1/manifest.spec.yml#L281 and https://github.com/elastic/package-spec/blob/ade1f5216c0d1320c83269f533b6531ec48ff85b/versions/1/manifest.spec.yml#L103.

ycombinator: Actually, I'm not sure this will even pass basic YAML validation. So I'm wondering why we need this extra check for the missing version.

mtojek: Actually it passed somehow and resulted with panic. I found this issue with help of CI (see: https://github.com/elastic/elastic-package/pull/354).

ycombinator: I think it's because we are not short-circuiting in our validation logic. Maybe if there are errors here we should short circuit and fail fast?

https://github.com/elastic/package-spec/blob/ade1f5216c0d1320c83269f533b6531ec48ff85b/code/go/internal/validator/spec.go#L64

mtojek: I like this idea. I adjusted the PR.

code/go/internal/validator/spec.go
```diff
@@ -61,16 +61,17 @@ func (s Spec) ValidatePackage(pkg Package) ve.ValidationErrors {
 	}
 
 	// Syntactic validations
-	errs.Append(rootSpec.validate(pkg.Name, pkg.RootPath))
+	errs = rootSpec.validate(pkg.Name, pkg.RootPath)
+	if len(errs) != 0 {
+		return errs
+	}
```
ycombinator, line 67: With this change, do we even need the rest of the changes in this PR? I would assume the lack of a `version` value in `manifest.yml` would get caught at this point and we would never get to the semantic validations?

mtojek: Yes, but on the other hand that code is unsafe - lack of type check. I would be for leaving it as is (less error prone code).

ycombinator: Fair point. Let's keep the type assertion.

ycombinator (approved):
LGTM.

PR #175, opened 2021-05-13
Fix security_rule format

mtojek:
## What does this PR do?

This PR fixes the expected format of security rules.

## Why is it important?

We need to properly validate IDs.

## Checklist

## Related issues

- https://github.com/elastic/integrations/pull/987


ycombinator (approved):
LGTM.

code/go/internal/validator/semantic/kibana_matching_object_ids.go
```diff
@@ -32,18 +32,26 @@ func ValidateKibanaObjectIDs(pkgRoot string) ve.ValidationErrors {
...
 
+		// Special case: object is of type 'security_rule'
+		if filepath.Base(filepath.Dir(filePath)) == "security_rule" {
+			ruleID, err := objectFile.Values("$.attributes.rule_id")
+			if err != nil {
+				errs = append(errs, errors.Wrapf(err, "unable to get rule ID in file [%s]", filePath))
+				continue
+			}
+
+			if ruleID != objectID {
```
rw-access, line 49: There's no reason these have to match.

The `rule_id` is only relevant to the Security solution.
The top level `id` is the saved object ID. It's possible that in the future we want better named SO ids. But we can loosen up this spec up then if that happens.

PR #176, opened 2021-05-31
Define common test config for pipeline tests

mtojek:
## What does this PR do?

This PR defines a common test config for pipeline tests.

## Why is it important?

With this feature there won't be necessary anymore to duplicate the same test config per every pipeline test.

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/issues/289


P1llus (approved):
LGTM

Issue #63, opened 2020-10-15
[Discuss] Avoiding duplication of ECS field definitions


andrewkroh:
Currently each dataset duplicates the definitions of the ECS fields that it uses (data type, descriptions, examples, etc). This puts a burden on the package maintainers to copy the ECS definitions into the dataset and keep it in sync with ECS.

It would be simpler to develop and maintain a package if the dataset only required listing the names of ECS fields that the module uses (and the ECS version). When the package is build the full field definitions for the specified fields can be imported from ECS.

The dataset would declare:
- ECS version
- List of ECS fields (maybe allowing for patterns like `host.*`, but that's part of the discussion)

The build step would create field declarations in YAML format that conform the package spec.

andrewkroh:
I'm not sure this affects the specification. Assuming that the build step used in `elastic/integrations` generates packages that comply to the package-spec then no changes are necessary.

mtojek:
@ycombinator and I had a discussion about possible mitigation of ECS duplication. One of the ideas we talked about bases on introducing a new JSON/YAML file in `<data_stream>/_dev` directory with listed ECS fields. This file would be processed during the build step and will result in the generated `fields/ecs.yml` with generated ECS content using the versioned fields from the ECS repository.

The `_dev` is an extension to the package specification format that contains resources used to build or test packages, not strictly package assets like fields, configs, images, etc.

ruflin:
If we add this to the package spec, I think it should not be a requirement. A dev that still wants to create all fields inside the dataset should be able to. I'm ok to make it a generic feature but would at first iterate in integrations on it.

mtojek:
> If we add this to the package spec, I think it should not be a requirement.

It must be in the package-spec to preserve the standard across all packages. It's not strictly the package specification, but the `_dev` extension. We decided to define `_dev` together in the package specification and not the create separate on and depend on the package-spec (KISS).

> A dev that still wants to create all fields inside the dataset should be able to

Regarding the requirement, I'm rather fan of hard requirements which make the standard stricter. Let's say the `ecs.yml` is restricted and is only generated during the build step. I'm open for discussion, but this will help us to solve the original problem. Unless developers should modify/overwrite the ECS field definitions.

> I'm ok to make it a generic feature but would at first iterate in integrations on it.

The feature will appear only in integrations as the `_dev` folder is not copied to the package-storage.

ruflin:
[...]

PR #181, opened 2021-06-02
Define build time dependencies

mtojek:
## What does this PR do?

This PR introduces support for build-time dependencies.

## Why is it important?

We need to checkout field definitions from the ECS repository and use them to compile up-to-date and synced fields. 

## Checklist

## Related issues

- https://github.com/elastic/package-spec/issues/63
- https://github.com/elastic/package-spec/pull/179


versions/1/manifest.spec.yml
```diff
@@ -273,6 +273,25 @@ spec:
           pattern: '^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\/[a-zA-Z0-9-]+))$'
       required:
       - github
+    buildDependencies:
```
jsoriano, line 276: Use snake_case for coherence with other fields.
```suggestion
    build_dependencies:
```

versions/1/manifest.spec.yml
```diff
@@ -273,6 +273,25 @@ spec:
...
+        properties:
+          name:
+            type: string
+            description: Name of the dependency
+            enum:
+              - ecs
+          reference:
+            type: string
+            description: Source reference
+            pattern: 'git@[a-z0-9]{40}'
```
jsoriano, line 291: Allow the use of tags too? (Can be extended in the future in any case)

test/packages/good/manifest.yml
```diff
@@ -40,3 +40,6 @@ icons:
     title: system
     size: 1000x1000
     type: image/svg+xml
+buildDependencies:
+  - name: ecs
+    reference: git@7a11f3e1f7d8f05e8f06623fd98254b8cb34c441
```
jsoriano, line 45: I wonder if there should be a `build` object, in case there are more build options in the future.
```suggestion
build:
  dependencies:
  - name: ecs
    reference: git@7a11f3e1f7d8f05e8f06623fd98254b8cb34c441
```

mtojek: Hm.. sounds like a more future-oriented design. Let me adjust the schema.

versions/1/manifest.spec.yml
mtojek: I will follow the other approach (with `build` object).

versions/1/manifest.spec.yml
mtojek: The reason why I chose a commit hash is because it's always the same for all develops. If we put a tag here (e.g. `git@1.9.0`), the elastic-package will build a slightly different package based on the commit marked with the tag. I was thinking about a switch for `elastic-package build`, for example: `--update`, which can update references to latest dependency versions.

Actually I see a problem with this design: how can I pick the latest commit for ECS 1.9? Let me suggest something.

test/packages/good/manifest.yml
mtojek: Fixed

versions/1/manifest.spec.yml
mtojek: Ok, I adjusted the manifest to include additional field "version". The update procedure would be as follows:

1. Check ECS dependency considering version (e.g. `1.9`).
2. Fetch latest commit from the branch `1.9`.
3. Update `reference` with the commit hash.

test/packages/good/manifest.yml
```diff
@@ -40,3 +40,8 @@ icons:
     title: system
     size: 1000x1000
     type: image/svg+xml
+build:
```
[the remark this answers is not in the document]
mtojek: There are pros and cons for both options.

`it will also stay in the package when downloaded by Kibana. Do we want that?`

Pros: 
* Kibana can link directly to ECS docs or repo to refer to the revisions
* We don't need to reverse engineer the ECS version used by the built package

Cons:
* Manifest contains new entries for dependencies

`Or should it be really a build time definition that is removed?`

I would be against modifying the manifest to get rid of dependencies. Maybe we can get rid of the `build` level and just leave dependencies.

`Maybe a separate file?`

Hm.. we can extract this logic to `deps.yml`. WDYT?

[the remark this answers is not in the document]
mtojek: > One advantage of depds.yml is that there could be a command elastic-package deps update (just made it up) that automatically updates references in the deps.yml file. I would not necessarily do the same with the manifest.

Yes, that's the plan. I was thinking about `elastic-package build --update`, which will try to pull latest commit references.

versions/1/manifest.spec.yml
jsoriano: @mtojek having a `version` and a `reference` can be confusing, even contradictory if someone sets a reference that doesn't correspond to a `version`. I think that we can go on by now with your original approach with the `reference` only, and we can improve it later to support other use cases.

> how can I pick the latest commit for ECS 1.9? 

> Fetch latest commit from the branch `1.9`.

I wouldn't try to support branches, at least by now, configured branches may lead to unreproducible builds, same code build at different moments can result in packages with different fields.

ruflin: ++

mtojek:
Ok, I extracted dependencies to an external file named "deps.yaml".

test/packages/good/deps.yml
```diff
@@ -0,0 +1,4 @@
+build:
```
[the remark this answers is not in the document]
mtojek: I'm ok with `build.yml` too, although this may suggest to introduce more side features into it :) We have also `_dev` directory, maybe this one is a better fit? For example: `_dev/build/deps.yml`. WDYT?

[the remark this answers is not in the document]
mtojek: I adjusted the PR to use `build.yml` in the package root directory

versions/1/manifest.spec.yml
mtojek: Fixed

jsoriano (approved):

test/packages/good/_dev/build/build.yml
```diff
@@ -0,0 +1,3 @@
+dependencies:
```
ruflin, line 1: Just approved the other PR but we have one more name here: `dependency`.

I assume the ecs dependency can only be used once, should we add it by name instead of array?

```
dependencies:
  ecs:
    reference: foo
```

mtojek: Replaced array with objects. I assume it's just to prevent custom dependencies that users may try to introduce and we don't want them now.

jsoriano (approved):
I like it with objects too :slightly_smiling_face:

ruflin (approved):

PR #183, opened 2021-06-08
Validate if field groups don't have units and metric types defined

mtojek:
## What does this PR do?

This PR introduces another semantic validation check to verify if field groups don't use units and metric types.

## Why is it important?

Field groups shouldn't use units and metric types.

## Checklist

## Related issues

- https://github.com/elastic/integrations/issues/1091


fearful-symmetry (approved):
not 100% familiar with the package-spec, but otherwise LGTM.

Issue #182, opened 2021-06-07
Use go-licenser to include license headers


mtojek:
Introduce similar mechanism of including license headers at the beginning of every Go code file.

mtojek:
Let me work on this.

PR #184, opened 2021-06-08
Ensure license headers for Go files

mtojek:
## What does this PR do?

This PR ensures that all .go files have proper license headers.

## Why is it important?

Legal reasons.

## Related issues

- Fixes https://github.com/elastic/package-spec/issues/182


mtojek:
self-approved

kaiyan-sheng (approved):
👍

PR #186, opened 2021-06-09
Fix: enable field group validation

mtojek:
This PR enables the field group validation (missed in previous PR).


mtojek:
emergency fix

PR #191, opened 2021-06-21
Relax ECS dependency on build manifest

mtojek:
## What does this PR do?

This PR relaxes the restriction on the ECS depedency, so it doesn't depend on Git hashes only.

## Why is it important?

The common use case might be switching the dependency on ECS 1.8 to ECS 1.9. With Git tags it will be human readable comparing to commit hashes.

## Checklist

## Related issues

- https://github.com/elastic/elastic-package/pull/379


jsoriano (approved):
:+1:

Issue #189, opened 2021-06-16
Semantic validation: forbid two same version entries in the changelog file


mtojek:
Spotted in https://github.com/elastic/integrations/pull/1127/files#r652445644 

There should be one entry per version in the changelog file. We need to improve semantic validation here.

mtojek:
I'm going with one (looks like a typical side-quest) :)

PR #193, opened 2021-06-21
Improve semantic versioning for version integrity

mtojek:
## What does this PR do?

This PR modifies the semantic versioning for version integrity to allow for "-next" entries and prevent duplicated version entries in changelog.

## Checklist

## Related issues

- Fixes https://github.com/elastic/package-spec/issues/189


jsoriano:
Umm, after implementing this, what would be the proper format to include multiple changes in a single version?

jsoriano:
> Umm, after implementing this, what would be the proper format to include multiple changes in a single version?

For example this case looked straightforward to me with two entries with the same version :sweat_smile: https://github.com/elastic/integrations/pull/1153/files#r655491549

mtojek:
Exactly the way as @kaiyan-sheng proposed in https://github.com/elastic/integrations/pull/1153/files#r655601336 :) That's the expected form hence this PR to prevent other options.

test/packages/same_version_twice/changelog.yml
```diff
@@ -0,0 +1,21 @@
...
+- version: "0.0.2"
+  changes:
+    - description: Second
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/193
+- version: "0.0.1"
+  changes:
+    - description: First
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/193
```
jsoriano, line 21: Is there any test case with a version block with multiple changes?

jsoriano (approved):
Great! Thanks for the clarifications.

test/packages/same_version_twice/changelog.yml
mtojek: Good catch! I added multiple entries in few changelogs.

jsoriano: Thanks!

PR #194, opened 2021-06-22
Add support for specifying a constant_keyword value

andrewkroh:
## What does this PR do?

For the `constant_keyword` field type Elasticsearch allows for the value to be specified in the mapping.
If the value is not specified in the mapping then the value is set based on the first document that is
indexed.

Package data streams will be able to explicitly declare a constant_keyword value and will not have to
include the field with every document sent.

https://www.elastic.co/guide/en/elasticsearch//reference/7.12/keyword.html#constant-keyword-params

## Why is it important?

This gives packages more control over constant_keyword field values. It can be used optimize storage for an integration since you could omit the constant_keyword field entirely an event's `_source`.

## Checklist


ruflin (approved):
LGTM

fearful-symmetry (approved):
Seconding, LGTM

test/packages/good/data_stream/foo/fields/external-fields.yml
```diff
@@ -3,4 +3,7 @@
   description: Event family
   fields:
     - name: category
-      external: ecs
\ No newline at end of file
+      external: ecs
+    - name: module
```
mtojek, line 7: nit: could you please put `event.module` in a different `fields.yml`? The intention of `external-fields.yml` was to keep only referenced ones :) something like `ecs.yml` would be good.

test/packages/good/data_stream/foo/fields/external-fields.yml
```diff
@@ -3,4 +3,7 @@
...
+      type: constant_keyword
```
mtojek, line 8: I see that this is an inconsistency comparing to the original source - ECS repository, which defines it as `keyword`:

```
    - name: module
      level: core
      type: keyword
      ignore_above: 1024
      description: 'Name of the module this data is coming from.
        If your monitoring agent supports the concept of modules or plugins to process
        events of a given source (e.g. Apache logs), `event.module` should contain
        the name of this module.'
```

Do we plan to adjust the ECS schema too (`event.module`)?

This brings me to the final question - currently the import logic in elastic-package imports fields as they are defined. In this case (`event.module`) the definition would be altered with local properties (`value`, `type`). I'm making sure this is expected behavior.

test/packages/good/data_stream/foo/fields/external-fields.yml
andrewkroh: Absolutely, I should have realized by the name that this wasn't the right place. I moved it to `some_fields.yml` and made the example generic.

test/packages/good/data_stream/foo/fields/external-fields.yml
andrewkroh: > Do we plan to adjust the ECS schema too (event.module)?

AFAIK there are plans at the moment to change the definition of `event.module`. But there have been some talks of loosing the specification to allow implementors to use either keyword or constant_keyword.

// fyi: @elastic/ecs 

> In this case (event.module) the definition would be altered with local properties (value, type). I'm making sure this is expected behavior.

I was expecting that packages would be required to declare their full definition of `event.module` if they make it a `constant_keyword`.

mtojek: If you prefer you can leave overriding for further improvement (not to block this PR).

mtojek (approved):
Ship it

test/packages/good/data_stream/foo/fields/external-fields.yml
ebeahan: > AFAIK there are plans at the moment to change the definition of event.module.

I've heard some discussion, but I'm not aware of a decision being made yet.

I suggested allowing `keyword` and `constant_keyword` to be interchangeable and avoid requiring a specific indexing strategy for all ECS users who wish to use `event.module`.

PR #196, opened 2021-07-07
Add dependabot

v1v:
## What

Delegate the bump automation to `dependabot`. Please read below

## Why

Simplify recurrent tasks

## Next steps

`Allow` is not supported to declare the list of dependencies to be bumped, something that is supported for some other package systems. See https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#allow

The proposal is, delegate the bump automation to dependabot and for each dependency that should not be taking in consideration then:

- ignore it explicitly as stated in https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#creating-ignore-conditions-from-dependabot-ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - dependencies
+      - go
+  	open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,7 @@
 version: 2
 updates:
   - package-ecosystem: "gomod"
-    directory: "/"
+    directory: "/code/go"
     schedule:
       interval: "daily"
     labels:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,4 +8,4 @@ updates:
     labels:
       - dependencies
       - go
-  	open-pull-requests-limit: 10
+    open-pull-requests-limit: 10


kuisathaverat (approved):

mtojek (commented):
elastic/package-spec is not a standard Go package, as it exists in the "code/go" directory. See: https://github.com/elastic/package-spec/tree/master/code/go

Will dependabot autodiscover it?

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
```
v1v, line 5:
```suggestion
    directory: "/code/go"
```

v1v:
It wasn't until I just added the folder:
- https://github.com/elastic/package-spec/pull/196/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R5

Thanks :)

mtojek (approved):
Green from me too :)

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/code/go"
+    schedule:
+      interval: "daily"
+    labels:
+      - dependencies
+      - go
+  	open-pull-requests-limit: 10
```
jsoriano, line 11: Is indentation correct here? (it seems to be at the same level as labels items).

jsoriano (commented):
Nice!

.github/dependabot.yml
v1v:
```suggestion
    open-pull-requests-limit: 10
```

jsoriano (approved):

PR #200, opened 2021-07-09
[dependabot][mergify] automate merge and add reviewers

v1v:
## What does this PR do?

- Add reviewers
- Automerge if CI is happy. Although, if no approval is needed why do we need the reviewers?

## Why is it important?

Look mum no hands!

## Issue

See https://github.com/elastic/integrations/pull/1291#issuecomment-876963997

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
     schedule:
       interval: "daily"
     labels:
-      - dependencies
-      - go
+      - automation
+    reviewers:
+      - "mtojek"
+      - "jsoriano"
     open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,5 @@ updates:
     labels:
       - automation
     reviewers:
-      - "mtojek"
-      - "jsoriano"
+      - "elastic/integrations-developer-experience"
     open-pull-requests-limit: 10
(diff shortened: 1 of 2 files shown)


cachedout (approved):

.github/dependabot.yml
```diff
@@ -6,6 +6,8 @@ updates:
     schedule:
       interval: "daily"
     labels:
-      - dependencies
-      - go
+      - automation
+    reviewers:
+      - "mtojek"
+      - "jsoriano"
```
jsoriano, line 12: I'd prefer to put a [github] team here so we can more easily manage notifications and we don't need need to change this if the people related to the project changes. But we can go on like this by now.

kuisathaverat (approved):

mtojek (approved):

.github/dependabot.yml
jsoriano: @v1v and I have discussed about this today and we have created a new "integrations-developer-experience" github subteam for this, so we can use it to assign reviewers.

@mtojek sounds this approach good to you?

mtojek: SGTM! I like the idea :)

v1v:
```suggestion
      - "elastic/integrations-developer-experience"
```

mtojek (approved):

jsoriano (approved):

PR #152, opened 2021-03-18
Support ML module Kibana type

jgowdyelastic:
Adds support for ML module asset types.

ML Module's are now supported in the form of saved objects in kibana. This PR adds the support for installing the modules via Fleet's integration packages.

Once a Module has been installed via Fleet, it is available for use in ML. This involved calling ML's "setup" endpoint which will create the contained Jobs, with optional overrides, and then optionally starting the Jobs.

An ML Module installed via Fleet should be free to be deleted or upgraded with all other assets in the package.
Deleting an ML Module will not remove any previously created Jobs.
Upgrading the ML Module will not update or affect in any way any previously created Jobs.

- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.
- [x] I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml).

## Related issues

Related to:
https://github.com/elastic/package-spec/issues/148
elastic/kibana#92855
https://github.com/elastic/kibana/pull/94950

PR #204, opened 2021-08-04
Support ML Model asset type

alvarezmelissa87:
Adds support for ML model asset types.

This PR adds the support for installing ML models via Fleet's integration packages. It adds a top level `elasticsearch` directory with `ml_model` type and top level `ingest_pipeline` type.

## Why is it important?

ML models are installed in elasticsearch directly so they are top level assets. There is no other package with top level elasticsearch assets so I've added this new directory to support those. 

## Checklist

## Related issues

-


ruflin (commented):
Before we add the models to the spec, lets figure out all the details around how it is installed in Kibana, upgraded, downgraded, uninstalled: https://github.com/elastic/package-spec/blob/master/CONTRIBUTING.md#change-proposals Otherwise we end up with something in the spec but don't know yet how it is handled.

-------


@ruflin  - thanks for the suggestion! I've created the PR here https://github.com/elastic/package-spec/issues/216

versions/1/elasticsearch/spec.yml
```diff
@@ -0,0 +1,28 @@
+spec:
+  additionalContents: false
+  contents:
+    - description: Folder containing ML model assets
```
mtojek, line 4: I understand that ML model is more related to Elasticsearch than Kibana? I'm asking about this as we used to install such resources as Kibana items like security rules, index patterns and ML module: https://github.com/elastic/package-spec/pull/152

Is this ML model different than ML module already added to Kibana?

versions/1/elasticsearch/spec.yml
```diff
@@ -0,0 +1,28 @@
...
+      type: folder
+      name: ml_model
+      required: false
+      contents:
+        - description: An ML model file
+          type: file
+          contentMediaType: "application/json"
+          pattern: '^{PACKAGE_NAME}-.+\.json$'
+    - description: Folder containing Elasticsearch Ingest Node pipeline definitions
```
mtojek, line 13: Do you need the ingest pipeline to be installed as unrelated from data stream?

mtojek (commented):
Left few design questions, once we agree on this we can polish other parts (tests, changelog, etc.)

ruflin:
We have endpoint package with a global elasticsearch directory: https://github.com/elastic/package-storage/tree/production/packages/endpoint/1.1.1/elasticsearch I wonder now how this fits in.

mtojek:
> We have endpoint package with a global elasticsearch directory: https://github.com/elastic/package-storage/tree/production/packages/endpoint/1.1.1/elasticsearch I wonder now how this fits in.

For the current moment this directory is considered "illegal" as it violates the package-spec. I pinged @pzl over slack to elaborate on this.

ruflin:
This comes out of the work on transform. Seems we sneaked this in without the proper spec changes. Maybe we should have a separate issue for this?

mtojek (commented):
Regarding the ML model lifecycle, I think we lack a description which covers the upgrade procedure (what's going with resources, rollback handling, increase in size, etc.).

versions/1/elasticsearch/spec.yml
alvarezmelissa87: Hi, @mtojek - the ML model is stored in Elasticsearch, yes. It's not a Kibana item. ML modules include job configs and such that are kept track of as saved objects and then run latere so that's why they're in the Kibana asset sectioin.

versions/1/elasticsearch/spec.yml
alvarezmelissa87: Yes - that's why I've added the top-level ingest pipeline here. Part of the changes to the Fleet code for this include installing pipelines that are not part of a data stream.

versions/1/changelog.yml
```diff
@@ -175,3 +175,6 @@
   - description: Introduce "value" property for constant_keyword fields
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/194
+  - description: Support `ML model` Elasticsearch type.
```
mtojek, line 178: You put this in a wrong place. Please consider the order or changes in the changelog. Next version is `1.1.1-next` and this entry should be long there.

versions/1/elasticsearch/spec.yml
```diff
@@ -0,0 +1,28 @@
+spec:
```
mtojek, line 1: You need to add reference to this file in the package's spec: https://github.com/elastic/package-spec/blob/master/versions/1/spec.yml

versions/1/changelog.yml
```diff
@@ -10,6 +10,9 @@
   - description: Add tag support for elastic export
     type: bugfix
     link: https://github.com/elastic/package-spec/pull/223
+  - description: Support `ML model` Elasticsearch type.
```
mtojek, line 13: nit: please either use ML model or `ml_model` 

`code tags` are reserved for YAML snippets or real manifest properties

test/packages/good/elasticsearch/ml_model/good-ml-model-abc-1.json
```diff
@@ -0,0 +1,99 @@
+{
+  "model_id": "default",
```
mtojek, line 2: Just to be sure: this file will be installed in Elasticsearch by Kibana, right? If so, could you please confirm with @jen-huang that this structure is valid?

versions/1/elasticsearch/spec.yml
```diff
@@ -0,0 +1,28 @@
...
+  additionalContents: false
+  contents:
+    - description: Folder containing ML model assets
+      type: folder
+      name: ml_model
+      required: false
+      contents:
```
mtojek, line 8: `additionalContents: false` ?

versions/1/elasticsearch/spec.yml
```diff
@@ -0,0 +1,28 @@
...
+        - description: An ML model file
+          type: file
+          contentMediaType: "application/json"
+          pattern: '^{PACKAGE_NAME}-.+\.json$'
+    - description: Folder containing Elasticsearch Ingest Node pipeline definitions
+      type: folder
+      name: ingest_pipeline
+      additionalContents: false
+      contents:
+        - description: Supporting ingest pipeline definitions in YAML
```
mtojek, line 18: I guess you can also add test file for such ingest pipeline in `elasticsearch` directory.

PR #207, opened 2021-08-10
Prepare for release 1.0.0

mtojek:
This PR prepares the package-spec for GA.


exekias (approved):
LGTM!

I know we don't atm, but would it make sense to tag releases? not sure if this is something good or an unnecessary overhead

mtojek:
> I know we don't atm, but would it make sense to tag releases? not sure if this is something good or an unnecessary overhead

I think we have to do this as it will make our `go.mod`s cleaner. See the one in elastic-package: [go.mod](https://github.com/elastic/elastic-package/blob/master/go.mod#L13).

mtojek:
@ycombinator Do you think that there is something else we should remember about?

ycombinator:
> Do you think that there is something else we should remember about?

Hmmm, not for the spec _per se_ but I'm wondering if it makes sense to also release a 1.0.0 of the Golang validation library that coexists with the spec? So then any downstream dependency could refer to the dependency version as 1.0.0, for example [in `elastic-package`](https://github.com/elastic/elastic-package/blob/e65051deb8fedff88d84cf616fe561fade0f27a8/go.mod#L13).

mtojek:
Yes, it perfectly makes sense. I'm planing to release/tag it once this PR is merged (as responded [here](https://github.com/elastic/package-spec/pull/207#issuecomment-896110480)).

jsoriano (approved):

PR #188, opened 2021-06-16
Introduce "dynamic_template" property for fields

axw:
## What does this PR do?

This PR introduces a new "dynamic_template" property for fields.

## Why is it important?

See https://github.com/elastic/package-spec/issues/171

## Checklist

- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.
- [x] I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml).

## Related issues

Closes #187

Issue #171, opened 2021-05-05
[Change Proposal] Add support for explicitly defining dynamic_templates


axw:
In https://github.com/elastic/beats/pull/25422 we are extending `fields.yml` in libbeat to support explicitly defining dynamic_templates, for use with the `dynamic_templates` bulk request parameter introduced in Elasticsearch 7.13.0: https://github.com/elastic/elasticsearch/pull/69948. The purpose of this is to support dynamically mapping complex metrics, such as histograms and summaries. Naturally we would like to support this in integration packages too.

The above mentioned Beats PR introduces a new boolean "dynamic_template" property to fields in `fields.yml`. If this property is true, then the field represents a dynamic template rather than a concrete field. For example:

```yaml
- name: histogram_template
  type: histogram
  dynamic_template: true
```

This will result in an entry in the `dynamic_templates` section of the index template like this:

```json
"dynamic_templates": {
  "histogram_template": {
    "mapping": {
      "type": "histogram"
    }
  }
}
```

Tasks:
- [ ] Update the Kibana Fleet plugin's [template creation logic](https://github.com/elastic/kibana/blob/master/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts) to understand the `dynamic_template` field property, creating dynamic templates (https://github.com/elastic/kibana/issues/102289)
- [x] Add `dynamic_template` to the [field schema](https://github.com/elastic/package-spec/blob/master/versions/1/data_stream/fields/fields.spec.yml) in package-spec (https://github.com/elastic/package-spec/issues/187)

ruflin:
++ on adding support for this. @exekias Do we already have packages where this could become useful?

exekias:
Yes! Thanks for opening @axw, the obvious one will be Prometheus, where we can use this to get rid of the suffixes to match types. @ChrsMark FYI.

I'm +1 to getting this in as it is, as it matches what we have in Beats after elastic/beats#25422

axw:
Seems we have consensus so I've removed the discuss label, created implementation issues, and added checkboxes.

simitt:
cc @jen-huang for visibility.

axw:
I'm going to close this and revert the change to package-spec. It occurred to me that we an define the dynamic templates directly in the data stream manifest. See https://github.com/elastic/kibana/issues/102289#issuecomment-904640242

PR #212, opened 2021-08-24
Revert "Introduce "dynamic_template" property for fields (#188)"

axw:
## What does this PR do?

This reverts commit d5f8de36cdb4098f174026bbb264cbd7e4dbf249 (https://github.com/elastic/package-spec/pull/188).

## Why is it important?

We introduced `dynamic_template` as a field property in beats as a means of defining explicitly named dynamic templates. This is not needed in integration packages, as we can directly define the `dynamic_templates` in the elasticsearch index template mapping. See https://github.com/elastic/kibana/issues/102289#issuecomment-904640242

## Checklist

~- [ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.~

## Related issues

https://github.com/elastic/package-spec/issues/171


versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -86,11 +86,8 @@ spec:
         type: string
         enum:
           - ecs
-      dynamic_template:
-        description: Indication that this field represents an explicitly named dynamic template
-        type: boolean
```
jsoriano, line 91: `dynamic_template` is part of a GA version of the spec, included in version 1.0.0. I am not sure if we can remove it at this point. @ruflin @exekias thoughts?

You can also mark it as deprecated by now.
```
      dynamic_template:
        description: Indication that this field represents an explicitly named dynamic template
        type: boolean
        deprecated: true # Define the dynamic templates in the Elasticsearch index template mapping. 
```

ruflin: I'm torn on this one. I don't think we should break GA so deprecating is the better option. At the same time the only packages built are by us at the moment and as long as no packages uses it (also in the past) and Kibana does not have an implementation for it, we are not required to keep it. It could also be documented as a breaking change for the 1.1 version.

jsoriano: Yep, no packages seem to be using this so far, and as Kibana doesn't support it, it is not useful. I am fine with including it as a breaking change, and prepare a 1.1 version (that would also include the go.mod at the root level https://github.com/elastic/package-spec/pull/213, what will allow `elastic-package` to use tagged versions of the spec).

axw: The field spec has `additionalProperties: true`, so the field will just be ignored anyway:

https://github.com/elastic/package-spec/blob/dc22ddb861596dc5a5f1c10e2d5fc381de3fc7c5/versions/1/data_stream/fields/fields.spec.yml#L9

jsoriano: Good point. SGTM then.

jsoriano (approved):

Issue #99, opened 2020-12-20
Replace github.com/rakyll/statik with go:embed


ycombinator:
With Go 1.16, we get support for static file embedding in Golang itself. I don't have authoritative resources to link to yet, but here are some others that might help for now:

- https://www.reddit.com/r/golang/comments/jkdjgw/complete_goembed_support_has_just_landed/gaiy7na/
- https://www.reddit.com/r/golang/comments/kfm4ly/embed_files_in_go_using_new_embed_package_in_go/

ycombinator:
Video tutorial showing the usage of `go:embed`: https://www.youtube.com/watch?v=YoDGTwwQTso

ruflin:
Nice, thanks for sharing. I think there are other places where we can use this feature.

PR #213, opened 2021-08-25
Use go:embed

axw:
## What does this PR do?

- Move go.mod up to top level, and change module name accordingly; we need to have a Go package at the same or higher level as the   embedded files, as go:embed does not permit ".." in patterns
- Update go.mod to specify Go version 1.16 (needed for go:embed)
- Add a minimal top-level package with go:embed, exposing the contents as an io/fs.FS
- Remove all use of rakyll/statik. Makefile update/check just update and check license headers respectively

Note that this is a breaking change to the Go module, due to the module name changing.

## Why is it important?

Simplifies development.

## Checklist

~- [ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/master/test/packages) that prove my change is effective.~
~- [ ] I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml).~

## Related issues

Closes https://github.com/elastic/package-spec/issues/99


jsoriano (approved):
Thanks!

PR #223, opened 2021-09-30
Add support for Kibana tags

jlind23:
Missing tag entry which cause elastic-package lint to fail.
Tag files are created when you add tag to a dashboard and then try to export it with elastic-package export

### DIFF
--- a/versions/1/kibana/spec.yml
+++ b/versions/1/kibana/spec.yml
@@ -83,7 +83,7 @@
           type: file
           contentMediaType: "application/json"
           pattern: '^{PACKAGE_NAME}-.+\.json$'
-    - description: Folder containing Kibana dashboard assets
+    - description: Folder containing Kibana tags
       type: folder
       name: tag
       required: false
@@ -92,5 +92,3 @@
         type: file
         contentMediaType: "application/json"
         pattern: '^{PACKAGE_NAME}-.+\.json$'
-        forbiddenPatterns:
-          - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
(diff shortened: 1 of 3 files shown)


versions/1/kibana/spec.yml
```diff
@@ -83,3 +83,14 @@
...
+      type: folder
+      name: tag
+      required: false
+      contents:
+      - description: A dashboard tag file
+        type: file
+        contentMediaType: "application/json"
+        pattern: '^{PACKAGE_NAME}-.+\.json$'
+        forbiddenPatterns:
+          - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
```
mtojek, line 96: I think you can drop these properties. The `ecs` prefix was present for resources migrated from Beats.

versions/1/kibana/spec.yml
```diff
@@ -83,3 +83,14 @@
           type: file
           contentMediaType: "application/json"
           pattern: '^{PACKAGE_NAME}-.+\.json$'
+    - description: Folder containing Kibana dashboard assets
```
mtojek, line 86: containing Kibana tags

mtojek (commented):
Please also add a changelog entry here: https://github.com/elastic/package-spec/blob/master/versions/1/changelog.yml

mtojek (commented):
One thing I forgot: could you please copy your "tag" file to https://github.com/elastic/package-spec/tree/master/test/packages/good/kibana ? So that we have a representative sample, which should be good. Please mind file naming rules  "good-..."

mtojek (approved):

Issue #224, opened 2021-10-06
[Change Proposal] Add support for specifying cluster privileges


stuartnelson3:
This proposal is a continuation of work that was initially begun with https://github.com/elastic/package-spec/issues/203.

In that issue it was decided to add `elasticsearch.privileges.indices`, but not specifically cluster privileges.

As was recently surfaced in https://github.com/elastic/beats/issues/28164, access to specific cluster privileges are required for apm-server's stack monitoring ui to correctly while while being managed by elastic-agent.

The specific problem is that the `cluster:monitor/main` privilege is required.

I could use some additional insight from others on the specifics of how to achieve this. Assuming this is approved, I have a few questions:

- Where would this should be added to package spec? The current implementation seems datastream-focused, while this would be specifically so that the apm-server is able to request data from elasticsearch's root path.
- Would this require additional changes to kibana? Again, the [linked code](https://github.com/elastic/kibana/blob/e771748d96ba098e2e35cf98a010f4f2a23c9c39/x-pack/plugins/fleet/common/types/models/epm.ts#L300) from the previous issue makes me believe it is data-stream focused

However it ends up being structured, I think the original proposal's core looks like a good start:

```yaml
elasticsearch:
  privileges:
    cluster: ['monitor/main']
```

stuartnelson3:
@jen-huang assuming this goes through, would you have bandwidth to get the necessary changes in on the kibana side by FF on oct 19?

ruflin:
Package manifest sounds like a good place for me as it seems it does not tie into a specific data stream.

stuartnelson3:
I'm going to go ahead and call this "accepted". I'll create implementation issues and then move forward from there.

jsoriano:
Closing this as it seems it was already implemented in https://github.com/elastic/package-spec/pull/226. Please reopen if there was anything pending to do.

PR #226, opened 2021-10-13
add elasticsearch.privileges.cluster support

stuartnelson3:
## What does this PR do?

add an optional `elasticsearch.cluster` property to the package manifest to define cluster privileges to be assigned to an api key

## Why is it important?

APM server needs additional cluster privileges to access `/` on its elasticsearch output

## Checklist

## Related issues

proposed in https://github.com/elastic/package-spec/issues/224

### DIFF
--- a/test/packages/good/manifest.yml
+++ b/test/packages/good/manifest.yml
@@ -40,4 +40,7 @@ icons:
     title: system
     size: 1000x1000
     type: image/svg+xml
+# /main is a specific action underneath the monitor privilege. Declaring
+# "monitor/main" limits the provided privilege, "monitor", to only the "main"
+# action.
 elasticsearch.privileges.cluster: [monitor/main]
--- a/test/packages/good/manifest.yml
+++ b/test/packages/good/manifest.yml
@@ -40,4 +40,4 @@ icons:
     title: system
     size: 1000x1000
     type: image/svg+xml
-elasticsearch.cluster: [monitor/main]
+elasticsearch.privileges.cluster: [monitor/main]
--- a/test/packages/good/manifest.yml
+++ b/test/packages/good/manifest.yml
@@ -40,3 +40,4 @@ icons:
     title: system
     size: 1000x1000
     type: image/svg+xml
+elasticsearch.cluster: [monitor/main]
--- a/versions/1/manifest.spec.yml
+++ b/versions/1/manifest.spec.yml
@@ -107,7 +107,7 @@ spec:
         - "1.0.0-dev.1"
   properties:
     elasticsearch:
-      description: Elasticsearch cluster requirements
+      description: Elasticsearch requirements
       type: object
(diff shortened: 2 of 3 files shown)


stuartnelson3:
@mtojek I interpreted this sentence:
>Maybe it would be good enough to keep it in the package manifest?

to mean that I should make my changes in the top-level manifest.yml. If this is wrong let me know!

versions/1/manifest.spec.yml
```diff
@@ -106,6 +106,16 @@ spec:
         - "1.0.0-SNAPSHOT"
         - "1.0.0-dev.1"
   properties:
+    elasticsearch:
+      description: Elasticsearch cluster requirements
```
mtojek, line 110: L110 and L115 have the same comment, but I guess that they have/should have different meaning?

test/packages/good/manifest.yml
```diff
@@ -40,3 +40,4 @@ icons:
     title: system
     size: 1000x1000
     type: image/svg+xml
+elasticsearch.cluster: [monitor/main]
```
ruflin, line 43: My understanding is that this is only about cluster privileges and not cluster settings in general. I wonder if we should prefix it.

ruflin: Looking at your initial proposal and what have here, it seems we should prefix it: https://github.com/elastic/package-spec/pull/211

stuartnelson3: something like `elasticsearch.cluster.privileges`?

joshdover:
Thanks. I'm going to delegate this to @nchaulet since he's working on https://github.com/elastic/kibana/issues/114686 already.

versions/1/manifest.spec.yml
```diff
@@ -106,6 +106,21 @@ spec:
         - "1.0.0-SNAPSHOT"
         - "1.0.0-dev.1"
   properties:
+    elasticsearch:
```
mtojek, line 109: I noticed that these properties are more or less sorted in prioritized order, which means that "elasticsearch" as extension should be at the bottom.

versions/1/manifest.spec.yml
```diff
@@ -106,6 +106,21 @@ spec:
...
+      type: object
+      additionalProperties: false
+      properties:
+        privileges:
+          description: Elasticsearch privilege requirements
+          type: object
+          additionalProperties: false
+          properties:
+            cluster:
+              description: Elasticsearch cluster privilege requirements
```
mtojek, line 120: Do you know if there is any doc/page describing all correct values for this property? You used `monitor/main`, but it might be a bit mysterious for people unfamiliar with this topic.

Do you think we should introduce regex pattern here?

mtojek:
This PR hasn't received all approvals and got merged. There is also an unaddressed comment.

PR #236, opened 2021-10-19
Add support for dimension fields

### LINKED ISSUE #215: Add support for dimension fields
Dimensions in time series databases are used to identify different time series. Metrics in documents with the same set of dimensions are considered the same time serie. Metrics with different set of dimensions are different time series.
This is specially relevant relevant when querying time series.

There is an effort in Elasticsearch to include dimensions as first-class citizens (see https://github.com/elastic/elasticsearch/issues/74014, https://github.com/elastic/elasticsearch/pull/74450), what will have advantages when storing and consuming time series data. To leverage these features in integrations, their dimensions should be properly defined.

For that we could add a new boolean property "`dimension`". Fields with this property set to true will be considered dimensions, any other field will be considered fields, or tags in the case of other metadata.

Not all types of metrics can be dimensions (to confirm what types can be dimensions).

jsoriano:
## What does this PR do?

* Add a new `dimension` parameter to fields, so they can be defined as dimensions for time series data. Dimensions are used to identify time series. For a data stream, events with the same set of dimensions are considered the same time serie. Only keywords, ips and numeric values can be defined as dimensions, this is enforced by the spec.
* Errors returned by the spec semantic validator for fields include now the file containing the invalid field. This was needed to leverage the current tests ([here](https://github.com/elastic/package-spec/blob/c531c9ed9668ce4afc73bf0ed75b077066aef380/code/go/pkg/validator/validator_test.go#L67-L72)) but is helpful in any case.

## Why is it important?

There is an effort in Elasticsearch to improve the support of time series data (see elastic/elasticsearch#74014, elastic/elasticsearch#74450, https://github.com/elastic/elasticsearch/issues/74660). As part of this, dimensions are going to be first-class citizens, what will have advantages when storing and consuming time series data. To leverage these features in integrations, their dimensions should be properly defined.

## Checklist

## Related issues

- Fixes #215

### DIFF
--- a/versions/1/changelog.yml
+++ b/versions/1/changelog.yml
@@ -4,12 +4,12 @@
 ##
 - version: 1.2.1-next
   changes:
-  - description: Add support for dimension fields.
-    type: enhancement
-    link: https://github.com/elastic/package-spec/pull/236
   - description: Support wait-for-data timeout
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/237
+  - description: Add support for dimension fields.
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/236
 - version: 1.2.0
   changes:
   - description: Add tag support for elastic export


code/go/internal/validator/semantic/validate_field_groups.go
```diff
@@ -31,18 +33,18 @@ type field struct {
 func ValidateFieldGroups(pkgRoot string) ve.ValidationErrors {
```
mtojek, line 33: Originally this function (and whole file) was designed to check field group properties. See [comment](https://github.com/elastic/package-spec/blob/cadd81674dc98da1aea3220df313b6085905114f/code/go/internal/validator/semantic/validate_field_groups.go#L32).

Does validation logic of dimensions apply to ordinary, single fields? Maybe it's worth extracting it to a separate .go file?

PR #237, opened 2021-10-19
Support wait-for-data timeout

mtojek:
## Checklist

## Related issues

- https://github.com/elastic/elastic-package/issues/540

### DIFF
--- a/versions/1/data_stream/_dev/test/system/config.spec.yml
+++ b/versions/1/data_stream/_dev/test/system/config.spec.yml
@@ -9,6 +9,6 @@ spec:
     skip:
       $ref: "./../skip.spec.yml#/definitions/skip"
     wait_for_data_timeout:
-      description: Short explanation for why test has been skipped.
+      description: Timeout for waiting for metrics data during a system test.
       type: string
       example: 10m


versions/1/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -8,3 +8,7 @@ spec:
   properties:
     skip:
       $ref: "./../skip.spec.yml#/definitions/skip"
+    wait_for_data_timeout:
+      description: Short explanation for why test has been skipped.
```
jsoriano, line 12: Umm, this description is unrelated? :thinking:

versions/1/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -8,3 +8,7 @@ spec:
...
+      type: string
```
jsoriano, line 13: Q: Are types defined somewhere? Would it be possible to define a duration type for this strings?

versions/1/data_stream/_dev/test/system/config.spec.yml
mtojek: My bad, copied this structure from the "skip" definition.

jsoriano (approved):

PR #241, opened 2021-11-01
Add email, timezone and url config types

jsoriano:
## What does this PR do?

Add new config variable types for emails, timezones and urls so Fleet UI can do additional validation.

## Why is it important?

To allow Fleet UI to implement additional validations on configuration fields.

## Checklist

## Related issues

- Required for for https://github.com/elastic/kibana/issues/116914


versions/1/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Support wait-for-data timeout
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/237
+  - description:  Add email, timezone and url config types
```
mtojek, line 10: nit: double space before "Add"

versions/1/data_stream/manifest.spec.yml
```diff
@@ -48,10 +48,13 @@ spec:
             description: Data type of variable.
             type: string
             enum:
-              - integer
               - bool
+              - email
+              - integer
               - password
               - text
+              - timezone
```
mtojek, line 56: time_zone or timezone?

mtojek (commented):
LGTM.

Should it be ready for review? It looks like a good idea, just need a confirmation from Kibana that they will implement it.

versions/1/data_stream/manifest.spec.yml
jsoriano: Not a strong opinion, what do you prefer? :)

jsoriano:
> Should it be ready for review? It looks like a good idea, just need a confirmation from Kibana that they will implement it.

I was waiting for confirmation of additional settings that could be needed, and how/if we would support them, see https://github.com/elastic/kibana/issues/116914#issuecomment-956208104.

versions/1/data_stream/manifest.spec.yml
mtojek: I checked the Elasticsearch docs and there are references to time_zone: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html

On the other hand I'm not a native speaker :)

jsoriano: Let's go with `time_zone` then :)

jsoriano:
I am adding the `url_allowed_schemes` to cover the needing of limiting the allowed schemes.

versions/1/data_stream/manifest.spec.yml
```diff
@@ -83,6 +86,14 @@ spec:
             default: true
             examples:
               - false
+          url_allowed_schemes:
```
jen-huang, line 89: I wonder if we should have a "generic" way to validate these kind of text fields. Maybe something like a regex pattern that Fleet policy editor can test the user input against (i.e. `regexPattern: /^https?/`). Although only URL scheme validation is introduced here, I can see other kinds of arbitrary validation being necessary. @elastic/fleet what do you think? It might be overengineered and could have security implications.

jsoriano: Good point. I don't have a strong opinion, but I think that for common use cases as this one I would prefer to have explicit settings, using regexes is sometimes a bit error prone, for example `/^https?/` would check that the string starts with the correct scheme, but it could still be an invalid URL. We would still need to mark this as url with `type: url` to indicate Fleet to use an URL parser to validate it.

In any case, adding a `regexPattern` setting can make sense for other use cases too, so we could add it anyway.

This is also a bit related to [this comment](https://github.com/elastic/kibana/issues/116914#issuecomment-956208104). If we are going to have many validation options I would prefer something as the second option with "validators". If not I would prefer the simpler approach implemented by now here.

joshdover: I think we probably want to some "easy button" built-in validators for package developers so they don't have to think too much and we have consistent behavior across inputs. We can also ensure these are very well tested by Fleet so we don't have to put that burden on package developers.

The biggest benefit I can see to allowing packages specify a custom regex for validation would be that we can update it out-of-band from Stack releases. So if there's a bug in the regex, we can fix it quickly for users without requiring a Stack upgrade. That said, I don't think this is going to be a huge problem in practice and would prefer we keep things simple for now. As you mentioned, I believe there is some security risk as well with exposing a generic regex feature. We'd likely need some level of sanitization to avoid expensive, complex regexes that could be used to DDoS Kibana.

If we have other use cases for this in the future, we could reconsider.

versions/1/data_stream/manifest.spec.yml
```diff
@@ -83,6 +86,14 @@ spec:
...
+            description: Allowed URL schemes for the url type. If not set, any scheme is allowed.
```
joshdover, line 90: I think we should call this `Allowed URL protocols` and `url_allowed_protocols` to match the URL spec's terminology

jsoriano: The truth is that I took the name from the Go implementation ([url.URL](https://pkg.go.dev/net/url#URL)), but checking the specs, they actually refer to this part as "scheme" (see [here](https://en.wikipedia.org/wiki/URL#Syntax), [here](https://url.spec.whatwg.org/#url-representation), or [here](https://www.w3.org/Addressing/URL/url-spec.txt)).

In any case, as schemes use to refer to protocols, I am happy to change this if it is more clear this way.

versions/1/data_stream/manifest.spec.yml
jsoriano: @joshdover so you are fine with the current approach with `type: url` and `url_allowed_scheme/protocol`?

hop-dev: Just to add to this, I think the regex would be useful, I have noticed that for the custom logs integration we tell users not to use `-` in the dataset name but there is nothing to stop them from doing so at the moment.

jsoriano: Good example. I am going to add this option too. I would still keep the option for allowed schemas because they cover a pretty common use case.

jsoriano: I have started thinking a bit about the support for regular expressions, and there are some open questions. Apart of the mentioned risks, to make it useful in many cases we may need to negate regular expressions, what would require additional settings (for example in the mentioned example we would want to accept values that don't contain `-`).
Also, if we want to validate the regular expressions, we may find problems given that the package tooling is written in Go, but the regular expressions are going to be executed in Kibana.

So I am moving this to a separate PR for further discussion (https://github.com/elastic/package-spec/pull/245), and I think we can continue here with the original options.

versions/1/data_stream/manifest.spec.yml
joshdover: I stand corrected, thanks!

jsoriano:
Comments addressed, discussion about adding support for more advanced format checking based on regular expressions moved to https://github.com/elastic/package-spec/pull/245.

This would be ready for a final review. Thanks!

test/packages/good/manifest.yml
```diff
@@ -17,7 +17,8 @@ policy_templates:
         multi: false
         vars:
           - name: hosts
-            type: text
+            type: url
+            url_allowed_schemes: ['http', 'https']
```
mtojek, line 21: nit: sometimes we set values as "localhost:9876". Will such form be accepted/forbidden?

mtojek (dismissed):

test/packages/good/manifest.yml
jsoriano: Not by this config, but it could be accepted if configured as `['http', 'https', '']`. I have added a clarification about this in the spec.

mtojek (approved):

PR #243, opened 2021-11-02
Support configuration of dimension fields limit

jsoriano:
## What does this PR do?

Adds support to increase the limit of dimension fields [`index.mapping.dimension_fields.limit`](https://www.elastic.co/guide/en/elasticsearch/reference/7.16/index-modules.html#dynamic-index-settings).

At the end it only adds a test case because index template settings are already supported.

## Why is it important?

An index can have by default a maximum of 16 dimension fields. For some cases it may be needed to increase this limit.

## Checklist

## Related issues

- Related to https://github.com/elastic/package-spec/issues/215

### DIFF
--- a/test/packages/time_series/data_stream/example/manifest.yml
+++ b/test/packages/time_series/data_stream/example/manifest.yml
@@ -9,3 +9,9 @@ streams:
         type: text
         title: Period
         default: 10s
+
+elasticsearch:
+  index_template:
+    settings:
+      # Defaults to 16
+      index.mapping.dimension_fields.limit: 32


test/packages/time_series/data_stream/example/manifest.yml
```diff
@@ -9,3 +9,9 @@ streams:
         type: text
         title: Period
         default: 10s
+
+elasticsearch:
```
mtojek, line 13: Hm.. just a test file? double-checking :)

jsoriano: Well, I was going to TDD this and found that nothing was needed :D

mtojek: Ah ok, now it makes sense :)

mtojek (approved):
LGTM

PR #245, opened 2021-11-03
Add format validation using regular expressions

jsoriano:
## What does this PR do?

Add settings for format validation using regular expressions. 

## Why is it important?

There are some cases where validation using regular expressions could be useful to avoid introducing errors in the generated configurations and give better feedback to users. See https://github.com/elastic/package-spec/pull/241#discussion_r741989214 for example.

## Open questions

- [ ] It is frequent to need things like a set of characters not allowed, this is tricky to do with regular expressions, for that we would start needing options, such as `negate`. This makes this feature to require a configuration object instead of a single value.
- [ ] Should the validator validate the format of the regular expression? and if done, what specification to follow? Packages are mainly consumed by Kibana, so ECMAScript regular expressions should be used, whose implementation has some differences with the Go `regexp` package.
  This PR tries to add a semantic validator for these fields following the approach for regular expressions of an ECMAScript implementation in Go (https://github.com/dop251/goja).
- [ ] As mentioned in https://github.com/elastic/package-spec/pull/241#discussion_r741453291, this feature can be a bit overengineering and have security implications, do we want to support this? Should we add additional restrictions to the supported regular expressions?

## Checklist

## Related issues

- Discussion started in https://github.com/elastic/package-spec/pull/241


jsoriano:
Closing it by now, to be reopened if there is more interest on this feature.

Issue #225, opened 2021-10-07
[Change Proposal] Remove release tag from packages


jsoriano:
There is currently a partial overlap in different features regarding the stability level of a package, that uses to bring confusion to integration developers and users:
* `release` label in manifests can be experimental/beta/ga.
* Semantic versioning, a package with a version >= 1.0.0 is (could be) considered "stable".
* There are three registries: snapshot/staging/production.

Apart of the overlap, there is no enforcement between these features, so it can be possible to find in the production repository a 1.0.0 package marked as experimental.

The proposal would be:
* Remove the `release` labels from the main manifest. To avoid breaking changes they could be marked as optional in the spec, and eventually be ignored everywhere, package registry API could do best-effort interpretation for old experimental-related queries.
* Pre-releases semver fields can be defined to release unstable versions. Packages stability (and supportability) would be determined only by their version, not from its origin. This implies the need of a change in a package to release it as stable, but this is ok, as this allow to introduce an approval step in the process.
* "Unstable" or experimental features could be still added to GA packages, but must be disabled by default. An experimental feature, causing an issue to a user of a GA package not interested in this feature should be considered a bug in the package. Release tag can still be used at the datastream level to mark datastreams or integration tiles as beta or experimental.
* Users can configure the source of packages, a single source can be configured now, but this may change in the future, the source doesn't determine the stability of a package.
* Users can optionally enable the use of pre-releases to use non-GA packages. Limited or no support level would be granted for these packages.
* Eventually snapshot and staging registries may be removed (https://github.com/elastic/package-storage/discussions/3666).

Some examples with this model:
* Package foo 0.5.0 wouldn't be considered stable, wherever it is published.
* Package bar 2.0.0 would be considered stable, and should be supported by their maintainers/owners.
* Package bar 2.1.0-beta wouldn't be considered stable, but a release candidate.

Discarded ideas:
* The registry that serves the package is relevant to its stability. In principle we are going to continue using the promotion model, and it can be safely assumed that the production registry contains more stable packages than the other ones. But consumers of packages (as Kibana/Fleet) shouldn't need to interpret the stability of packages from its origin ([comment](https://github.com/elastic/package-spec/issues/225#issuecomment-947479939)).
* Rejection of "unstable" packages in the production repository should be enforced (consequence of previous comment).
* Removal of the `release` tag at the datastream level delayed as it belongs to a broader discussion about how to tag experimental/beta features in GA packages.

## Required changes

[...]

PR #256, opened 2022-01-11
Deprecate the release tag

jsoriano:
## What does this PR do?

Deprecate the release attribute from package manifests.

Make it optional, and considered GA when not set.

Semver pre-release tags should be used after this change to publish non-ga versions. See https://github.com/elastic/package-spec/issues/225

## Why is it important?

To simplify the declaration of packages in different stages, see https://github.com/elastic/package-spec/issues/225.

## Checklist

## Related issues

- Relates https://github.com/elastic/package-spec/issues/225


jsoriano:
Mergify issue to be solved in https://github.com/elastic/package-spec/pull/257.

mtojek (approved):

test/packages/good/data_stream/foo/manifest.yml
```diff
@@ -1,6 +1,5 @@
 title: Nginx access logs
 type: logs
-release: experimental
```
ruflin, line 3: Is the release tag is removed everywhere now, is there a way to test the "deprecation" part?

jsoriano: The release tag is still present in other packages being tested, I only removed it from the "good" one.

versions/1/manifest.spec.yml
```diff
@@ -295,5 +298,4 @@ spec:
   - title
   - description
   - version
-  - release
```
jsoriano, line 298: Tested some packages without `release` tag on current versions of Kibana, and they work, but the package appears as experimental.

So I am thinking that we have to keep this as mandatory, to properly display packages in current and older versions, and maybe add semantic validation in the future to make it optional in future versions of kibana where https://github.com/elastic/kibana/issues/122973 is implemented.

jsoriano: @jen-huang @joshdover @ruflin would it be an option to change the default in Kibana from "experimental" to "ga", for packages that don't include the `release` field in their manifest? This would need to be done for 7.17.0 and 8.0.0.

This wouldn't change the behaviour for existing packages, because currently the `release` tag is mandatory in the spec. It would only change the behaviour for future packages, that more likely won't support Kibana versions older than 7.17.

jsoriano: I was checking what changes would be needed in Kibana for this, and I have found that Kibana takes the release as it comes from the registry. The "experimental" default is there, so we can do the change from our side.

https://github.com/elastic/package-registry/blob/ae68c7ac07afe1180a80a25f7dc2615a68b6b332/packages/releases.go#L13

mtojek: Nice finding :)

ruflin: Can you make sure it is the case also for older release of Kibana? I think it is but just to double check that we don't break 7.14 or similar.

test/packages/good/data_stream/k8s_data_stream/manifest.yml
```diff
@@ -1,6 +1,5 @@
 title: Data stream using Kubernetes service deployer for tests
 type: logs
-release: experimental
```
mtojek, line 3: I might be lost in the other email thread, but I assume we agreed on keeping the data stream release tag, right?

jsoriano: Yes, I removed the deprecation (in https://github.com/elastic/package-spec/pull/256/commits/366c8d63d04b2a55a2d63d55ffec340d090ca292), but it is still optional. I wanted to be sure that we have test cases with good packages without the tag.

versions/1/manifest.spec.yml
jsoriano: > Can you make sure it is the case also for older release of Kibana? I think it is but just to double check that we don't break 7.14 or similar.

Tried today, 7.14 behaves the same.

jsoriano:
@mtojek I have implemented the recent decision of limiting the allowed prerelease tags as a semantic validator, please take a look.

I considered modifying the [regular expression](https://github.com/elastic/package-spec/blob/a444c765a37258004dcb9ddad1b11275a03aeeec/versions/1/manifest.spec.yml#L104) instead, but I would prefer to leave this regular expression as is to validate semver according to the spec, and our conventions are easier to maintain as a semantic validator. We can also provide better feedback this way.

mtojek (approved):
LGTM!

versions/1/changelog.yml
```diff
@@ -10,6 +10,12 @@
   - description: Report error when a directory name inside a package contains hyphens (-).
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/259
+  - description: Deprecate the release attribute, make it optional and set to ga by default.
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/256
+  - description: Restrict the allowed prerelease labels in sematic versions.
+    type: breaking-change
```
mtojek, line 17: 👍

mtojek (approved):

PR #259, opened 2022-01-20
Flag error to indicate bad naming convention

r00tu53r:
## What does this PR do?

Package registry LoadAssets API errors out if the additional content for the top level files/folders (where its currently allowed) have dashes in their names. The change here flags the error here allowing it to be caught early during `elastic-package lint` time rather than failing during deployment to package registry.

## Why is it important?

Catches error during build/lint time rather than failing at runtime.

## Checklist

## Related issues

None


jsoriano:
@ruflin @mtojek do you know the reason why the package registry rejects directory names with `-`? Maybe we should remove this check in the registry instead?

This is the related code in the registry: https://github.com/elastic/package-registry/blob/master/packages/package.go#L381-L386

mtojek:
If I remember correctly, it was a logic processing issue on the Kibana Fleet side, as it assumed the type based on the name. For example:

```
nginx-logs
apache-logs
```

if you have `-` in it, it was a problem:

```
winnie-the-pooh-logs
```

Kibana had trouble finding the right type.

code/go/internal/validator/folder_spec.go
```diff
@@ -84,6 +84,13 @@ func (s *folderSpec) validate(packageName string, folderPath string) ve.Validati
 
 		if itemSpec == nil && s.AdditionalContents {
 			// No spec found for current folder item but we do allow additional contents in folder.
+			// Package registry LoadAssets API expects the name not have "-" in it. Check that here
+			// so the issue is caught at validation / lint time of the package.
+			if strings.Contains(fileName, "-") {
+				errs = append(errs,
+					fmt.Errorf(`file "%s/%s" is invalid: directory name inside package %s contains -: %s`,
+						folderPath, fileName, packageName, fileName))
+			}
```
jsoriano, line 93: As this check is only expected to be for directories, should this be placed under the `if file.IsDir()` block?

And probably this should be checked for all directories, so we avoid adding directories to the spec with hyphens, something that according to https://github.com/elastic/package-registry/issues/517 we don't want.

code/go/internal/validator/folder_spec.go
```diff
@@ -84,6 +84,13 @@ func (s *folderSpec) validate(packageName string, folderPath string) ve.Validati
 
 		if itemSpec == nil && s.AdditionalContents {
 			// No spec found for current folder item but we do allow additional contents in folder.
+			// Package registry LoadAssets API expects the name not have "-" in it. Check that here
+			// so the issue is caught at validation / lint time of the package.
```
jsoriano, line 88: No need to mention this here, the spec is the source of truth, it doesn't need to be justified with what the registry does :slightly_smiling_face:

You can though mention that this is done to avoid hyphens in directory names, as by https://github.com/elastic/package-registry/issues/517.

jsoriano (commented):
Thanks for bringing this check here!

code/go/internal/validator/folder_spec.go
r00tu53r: Thanks for the comment. Fixed.

code/go/internal/validator/folder_spec.go
r00tu53r: Fixed :)

jsoriano:
@r00tu53r could you please add a changelog entry to `versions/1/changelog.yml`?

versions/1/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Prepare for 1.3.1
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/250
+  - description: Flag error to indicate bad naming convention
```
jsoriano, line 10: Nit.
```suggestion
  - description: Report error when a directory name inside a package contains hyphens (-).
```

jsoriano (dismissed):

jsoriano (approved):
Thanks!

jsoriano:
> I think we have to do this as `_dev` doesn't play an important role in the interaction between Elasticsearch/Kibana/Fleet/Registry.

PR for this https://github.com/elastic/package-spec/pull/270

> Is there any breaking PR open?

Not yet, but there will be if we release current elastic-package main branch. I found this when testing the change for the release labels in https://github.com/elastic/integrations/pull/2651.

PR #267, opened 2022-02-05
Add support for platinum license

alvarezmelissa87:
## What does this PR do?

Adds support for platinum license type - used in the upcoming [ProblemChild](https://github.com/elastic/integrations/pull/2115) and [DGA](https://github.com/elastic/integrations/pull/2352) model packages.

## Why is it important?

The ML model packages mentioned above require:
- platinum licenses (support for this license type has already been added to Fleet)

## Checklist


mtojek (changes requested):
Could you please split this PR into 2 different pieces (new license and ML model adjustment)?

versions/1/changelog.yml
```diff
@@ -9,6 +9,9 @@
     link: https://github.com/elastic/package-spec/pull/265
 - version: 1.4.0
   changes:
+  - description: Support platinum license type.
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/267
```
jsoriano, line 14: Please move the changelog entry to the version `1.4.1-next` above.

test/packages/good/manifest.yml
```diff
@@ -3,6 +3,7 @@ name: good
 title: Good package
 description: This package is good.
 version: 1.0.0
+license: platinum
```
mtojek, line 6: I suggest leaving the license of the `good` package as is, as many packages are using basic (default) and creating a new one for `platinum` (`elastic-package create`).

mtojek (commented):
This change will open the gate for the first package with a platinum license. I'm concerned that our tools don't support it [out of the box](https://github.com/elastic/elastic-package/blob/a77bd34f4057ebcbb924c7715d45be7a91c1399d/internal/profile/_static/elasticsearch_config_8x.yml#L7), so it might cause troubles with serving the first platinum integration.

FYI we may need to adjust it before you push the first one.

mtojek:
@jen-huang @akshay-saraswat @joshdover Do you see any concerns around this from your perspective?

joshdover:
> * support for this license type has already been added to Fleet

We do have license checking during package install that will refuse to install if the cluster does not have a valid minimum license level. One question I have is how should this behave if the cluster's license expires after the package is installed? Will anything break if these assets are still installed after the license expires?

I'm guessing this is more of a thing that the ML code in ES needs to handle, but I'd like to know what the behavior is and whether or not we should consider any enhancements in the Fleet UI for this scenario.

> This change will open the gate for the first package with a platinum license. I'm concerned that our tools don't support it [out of the box](https://github.com/elastic/elastic-package/blob/a77bd34f4057ebcbb924c7715d45be7a91c1399d/internal/profile/_static/elasticsearch_config_8x.yml#L7), so it might cause troubles with serving the first platinum integration.

Doesn't a `trial` license give you access to platinum features too? Maybe it will just work for now.

mtojek:
> We do have license checking during package install that will refuse to install if the cluster does not have a valid minimum license level. One question I have is how should this behave if the cluster's license expires after the package is installed? Will anything break if these assets are still installed after the license expires?

I bet you have just discovered a scenario that nobody has explored yet :)

> Doesn't a trial license give you access to platinum features too? Maybe it will just work for now.

I checked it here: [docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/license-settings.html) and it says "basic", but check double-check.

mtojek:
I checked how the Elastic Package stack would behave with `platinum` package and it seems that it plays well, which is good news.

@joshdover Do you think we should open another issue for license expiration risks? If so, do you see any blockers for introducing another license? I assume that Kibana should check the license field before installing a package?

alvarezmelissa87:
> This change will open the gate for the first package with a platinum license. I'm concerned that our tools don't support it [out of the box](https://github.com/elastic/elastic-package/blob/a77bd34f4057ebcbb924c7715d45be7a91c1399d/internal/profile/_static/elasticsearch_config_8x.yml#L7), so it might cause troubles with serving the first platinum integration.
> 
> FYI we may need to adjust it before you push the first one.

Support for this license type was added to Fleet along with the model installation support in https://github.com/elastic/kibana/pull/107710/files#diff-cf9cec44de2ad0a6a3b74cca05e5308231d57d5c3e180ae4bc5114c2bf9af4ebR268

This will disable the card when the license type in the package manifest is not what it needs to be.

alvarezmelissa87:
> > We do have license checking during package install that will refuse to install if the cluster does not have a valid minimum license level. One question I have is how should this behave if the cluster's license expires after the package is installed? Will anything break if these assets are still installed after the license expires?
> 
> I bet you have just discovered a scenario that nobody has explored yet :)
> 
> > Doesn't a trial license give you access to platinum features too? Maybe it will just work for now.
> 
> ~I checked it here: [docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/license-settings.html) and it says "basic", but I can check double-check.~
> 
> Oh, I misread it, maybe it includes the options we need. Let's see.

Good question! If the assets are already installed and the license changes - the user will still need the ML level license for accessing the model and jobs - that's all handled on ML's end. It wouldn't break anything AFAIK. cc @peteharverson, @mtojek

akshay-saraswat:
I hope we have checked with the legal team that it's okay to distribute platinum licensed packages through EPR that operates with a basic license.

versions/1/changelog.yml
alvarezmelissa87: Updated in 0aa6454b67bd7d760b7dcc644e91050de50edaa2

test/packages/good/manifest.yml
alvarezmelissa87: Created a separate test package for platinum license in 0aa6454b67bd7d760b7dcc644e91050de50edaa2

mtojek:
> I hope we have checked with the legal team that it's okay to distribute platinum licensed packages through EPR that operates with a basic license.

I haven't checked that, it is precedence now. It's a good idea to double-check it. I'm wondering who should contact the legal team and lean towards @alvarezmelissa87 and @peteharverson, as you know ML internals and its license model. All we need is a confirmation that the "basic" Package Registry can serve "platinum" ML integration.

akshay-saraswat:
> > I hope we have checked with the legal team that it's okay to distribute platinum licensed packages through EPR that operates with a basic license.
> 
> I haven't checked that, it is precedence now. It's a good idea to double-check it. I'm wondering who should contact the legal team and lean towards @alvarezmelissa87 and @peteharverson, as you know ML internals and its license model. All we need is a confirmation that the "basic" Package Registry can serve "platinum" ML integration.

I can start an email chain with the legal team. I'll keep you all in cc so that you could provide more details to them. My understanding of the ML world in Elastic is limited. your help would be greatly appreciated.

peteharverson:
Many thanks @akshay-saraswat for emailing the legal team to check the license situation.

Some details on the ML internals:

The anomaly detection jobs will get closed when the license checker notices that the platinum license expired, and will be impossible to reopen until a platinum license has been installed. 

The inference model has a license level set that indicates the operation mode with which it is valid. If the ingest processor which uses the model has previously been used with a valid license, but the license goes out of date, it will continue to process, but it will write an audit message indicating that the license expired. So we don’t instantly stop people using models in an ingest pipeline when their license downgrades.

joshdover:
> Do you think we should open another issue for license expiration risks? If so, do you see any blockers for introducing another license? I assume that Kibana should check the license field before installing a package?

I think @peteharverson's answer above is sufficient. I also expect this to be a concern of the asset owners rather than Fleet itself.

One question I still have is whether or not a user would be able to uninstall this package and it's assets if they decide to downgrade. @alvarezmelissa87 when we disable the package card due to an mismatched license, how can a user uninstall the package?

ruflin:
I have a few additional questions on this:

* On the integrations page, will these platinum packages show up if someone runs Kibana under basic license? 
* Trying to better understand, why the package itself has to be under platinum. My assumption is, the assets itself could be under the basic license but during installation time, a platinum license is required because otherwise the feature would not work. To give an example: The apache package (basic) could contain model, but the model would only be enabled / installed when the user has a platinum license. Still everything would be under basic.

peteharverson:
To answer the second question raised by @ruflin 

> why the package itself has to be under platinum?

The key asset in the DGA and ProblemChild packages that are currently being prepared is the ML model, which can only be used with a platinum (or trial) license. Without access to the model, the package will provide no value to users, as the other assets (anomaly detection jobs, detection rules) all depend on the model. This is different to packages such as the Apache and Nginx ones, for which ML anomaly detection jobs were added for 7.13, as they do contain a range of other assets such as dashboards and ingest pipelines which do add value to users who are not on platinum licenses even without the ML jobs.

And on the first question

>  will these platinum packages show up if someone runs Kibana under basic license?

@alvarezmelissa87 @joshdover looking at the [comment](https://github.com/elastic/package-spec/pull/267#issuecomment-1032909236) above, am I right in thinking that for users on a basic license, the card will still be shown on the Integrations page, but the card will be disabled, preventing the user from trying to install it? This is the ideal behavior, as it will advertise a feature that will become available if they upgrade to a platinum license.

jlind23:
@ruflin @joshdover what if a license expire and we already have a couple of integrations installed? Should we keep them for a given period of time or should we remove those even in the agent policy?

alvarezmelissa87:
> > Do you think we should open another issue for license expiration risks? If so, do you see any blockers for introducing another license? I assume that Kibana should check the license field before installing a package?
> 
> I think @peteharverson's answer above is sufficient. I also expect this to be a concern of the asset owners rather than Fleet itself.
> 
> One question I still have is whether or not a user would be able to uninstall this package and it's assets if they decide to downgrade. @alvarezmelissa87 when we disable the package card due to an mismatched license, how can a user uninstall the package?

@joshdover - the card will never be disabled so the user will always be able to uninstall the package 👍  Good question!

alvarezmelissa87:
Closing as this change won't be going in anymore.

PR #268, opened 2022-02-07
Update ml_model file format pattern

alvarezmelissa87:
## What does this PR do?

This PR also updates the ml_model file format pattern for validation - used in the upcoming [ProblemChild](https://github.com/elastic/integrations/pull/2115) and [DGA](https://github.com/elastic/integrations/pull/2352) model packages.

## Why is it important?

The ML model packages mentioned above require:
- a ml model file format (in `elasticsearch/ml_model`) that matches the id of the model to be installed - this is why the format pattern is being updated.

## Checklist


versions/1/changelog.yml
```diff
@@ -9,6 +9,9 @@
     link: https://github.com/elastic/package-spec/pull/265
 - version: 1.4.0
   changes:
+  - description: Updates the ml_model file format pattern for validation.
+    type: enhancement
```
jsoriano, line 13: I'd suggest to consider this a breaking change, as files matching the previous pattern won't match the new one.

I'd also rephrase the changelog entry being more explicit about the update in the pattern. Maybe something like this:
```suggestion
  - description: ML model file name matches now the id of the model.
    type: breaking-change
```

versions/1/changelog.yml
```diff
@@ -9,6 +9,9 @@
...
+    link: https://github.com/elastic/package-spec/pull/268
```
jsoriano, line 14: Please add this changelog entry in the "next" version block above.

versions/1/elasticsearch/spec.yml
```diff
@@ -9,7 +9,7 @@ spec:
         - description: An ML model file
           type: file
           contentMediaType: "application/json"
-          pattern: '^{PACKAGE_NAME}-.+\.json$'
+          pattern: '^{PACKAGE_NAME}_.+\.json$'
```
jsoriano, line 12: Or if you don't want to make this a breaking change, allow both `-` and `_`.
```suggestion
          pattern: '^{PACKAGE_NAME}[-_].+\.json$'
```

versions/1/changelog.yml
```diff
@@ -9,6 +9,9 @@
     link: https://github.com/elastic/package-spec/pull/265
 - version: 1.4.0
   changes:
+  - description: Updates the ml_model file format pattern for validation.
```
mtojek, line 12: Please move it to the `1.4.1-next`.

jsoriano (changes requested):

versions/1/elasticsearch/spec.yml
mtojek, line 12: It looks like a breaking change for existing resources if we have already pushed anything to the integrations repo. Have we already done it? 

If so, then we need to modify the regexp to accept both entries.

mtojek (commented):
Could you please add a sample file to the `good` package?

versions/1/elasticsearch/spec.yml
mtojek: Cool, thanks for confirmation!

versions/1/changelog.yml
alvarezmelissa87: This isn't a breaking change as the only packages that will now use this have not been merged into integrations yet. No packages have used it so far.

versions/1/elasticsearch/spec.yml
alvarezmelissa87: See comment above about how nothing is using this pattern yet so it's not a breaking change.

alvarezmelissa87:
> Could you please add a sample file to the `good` package?

This already existed - just renamed it in this PR so it passed  the new validation:
`test/packages/good/elasticsearch/ml_model/good_ml_model_abc_1.json`

versions/1/changelog.yml
alvarezmelissa87: Made the suggested changelog changes in 9b30841d1905134ebfb81f5d24833e6e53446732

versions/1/changelog.yml
alvarezmelissa87: Updated in 9b30841d1905134ebfb81f5d24833e6e53446732

versions/1/changelog.yml
alvarezmelissa87: Moved in 9b30841d1905134ebfb81f5d24833e6e53446732

mtojek (approved):
LGTM, but let's wait for 👍 from @jsoriano

jsoriano (approved):

PR #270, opened 2022-02-08
Allow directory names with hyphens in _dev directories

jsoriano:
Add new flag `developmentFolder` to indicate if a directory contains development files. Inside these directories it is allowed to have files with hyphens.

Follow up to #259.

Solves issue mentioned in https://github.com/elastic/package-spec/pull/259#issuecomment-1032767888:
```
[2022-02-08T13:53:46.969Z] Error: checking package failed: linting package failed: found 1 validation error:
[2022-02-08T13:53:46.969Z]    1. file "/var/lib/jenkins/workspace/est-manager_integrations_PR-2651/src/github.com/elastic/integrations/packages/postgresql/_dev/deploy/docker/docker-entrypoint-initdb.d" is invalid: directory name inside package postgresql contains -: docker-entrypoint-initdb.d
```


code/go/internal/validator/folder_spec.go
```diff
@@ -85,7 +85,7 @@ func (s *folderSpec) validate(packageName string, folderPath string) ve.Validati
 		if itemSpec == nil && s.AdditionalContents {
 			// No spec found for current folder item but we do allow additional contents in folder.
 			if file.IsDir() {
-				if strings.Contains(fileName, "-") {
+				if !s.DevelopmentFolder && strings.Contains(fileName, "-") {
```
mtojek, line 88: A short challenge:

Do you agree that it is unusual to forbid `-` in dirnames? If so, maybe we should forbid it only for specific Kibana/Elasticsearch items. For example, I don't think that `-` should be forbidden for subdirectories in docs or images, or any other extra content, the developer would like to put there.

Long story short - instead of introducing `DevelopmentFolder` flag, I would consider a `StackInstalledFlag`, which forbids dashes.

Let me know what you think.

jsoriano: > Do you agree that it is unusual to forbid `-` in dirnames?

Yeah, I [fully agree](https://github.com/elastic/package-spec/pull/259#issuecomment-1017221604) :)

But if we forbid them I think it'd be less confusing to forbid them everywhere (as proposed in https://github.com/elastic/package-registry/issues/517) instead of on specific directories.
The exception in development directories (or at least in deploy directories) can make sense because there can be files there that have hyphens by convention in other systems (as the hyphens to separate words in some files in docker images).

Btw, if we allow hyphens everywhere except for directories containing stack files, we have to reconsider the removal of the [check](https://github.com/elastic/package-registry/blob/4a270e45cb76e4482f3d214512679c073abe10ed/packages/package.go#L393) in the package registry that originated https://github.com/elastic/package-spec/pull/259.

mtojek: > But if we forbid them I think it'd be less confusing to forbid them everywhere (as proposed in https://github.com/elastic/package-registry/issues/517) instead of on specific directories.

I admit I follow the intuition here and it would be confusing no matter which option we follow. I'm rather thinking about what's better for future use cases. I can imagine extra directories with docs, resources, extra scripts for testing.

> Btw, if we allow hyphens everywhere except for directories containing stack files, we have to reconsider the removal of the [check](https://github.com/elastic/package-registry/blob/4a270e45cb76e4482f3d214512679c073abe10ed/packages/package.go#L393) in the package registry that originated https://github.com/elastic/package-spec/pull/259.

Yeah, I'm thinking about this. Maybe this is the right timing to adjust it to match only relevant directories.

Anyway, back to this PR. We have only development resources, which are affected by this, so I would say it's fine to push this one and come back to the discussion when somebody else tries to a folder with a hyphen.

versions/1/_dev/spec.yml
```diff
@@ -1,5 +1,6 @@
 spec:
   additionalContents: false
+  developmentFolder: true
```
mtojek, line 3: nit: `_dev` is also in data_stream. You may add a test case that verifies that one too.

jsoriano: Good catch, for some reason I thought they shared the same spec. Fixed.

code/go/internal/validator/folder_spec.go
jsoriano: > Maybe this is the right timing to adjust it to match only relevant directories.

I would remove instead of adding validation logic from the registry :slightly_smiling_face:

mtojek: Well, we have spec now, so maybe we can make the Package Registry a bit more stupid :)

PR #272, opened 2022-02-09
Support auxiliary files in terraform deploy spec

### LINKED ISSUE #269: [Change Proposal] Support Terraform related files in deploy/tf
While [working at adding](https://github.com/elastic/elastic-package/pull/662) extended Terraform support to `elastic-package` Terraform Deployer (to enable GCP billing testing) package linting fails with this error:

```
Error: checking package failed: linting package failed: found 4 validation errors:
   1. item [.gitignore] is not allowed in folder [.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]
   2. item [.terraform.lock.hcl] is not allowed in folder [.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]
   3. item [billing-schema.json] is not allowed in folder [.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]
   4. item [test-data.ndjson.tftpl] is not allowed in folder **[.../test/packages/parallel/gcp/data_stream/billing/_dev/deploy/tf]**
```

This change proposal aims to add support `_dev/deploy/tf` folder. for:
1. `*.tftpl`; `tftpl` [is the recommended naming pattern to use for Terraform template files](https://www.terraform.io/language/functions/templatefile).
2. `*.json`; JSON is a standard storage format for accessory data. For the specific use case JSON is used to store GCP BigQuery Table schema.
3. `.terraform.lock.hcl`; this file is Terraform Dependency Lock Files. Terraform CLI itself [suggest](https://www.terraform.io/language/files/dependency-lock#dependency-installation-behavior) adding this file to version control for reproducible dependency installation.
4. `.gitignore`. It seems `gitignore` is not supported even at package level, and this makes everything a bit more complex. Adding support for this would also enabled ignoring the same files in related tools (`elastic-package build` for example).

endorama:
## What does this PR do?

- add support for Terraform dependency lock file
- add support for Terraform template files
- add support for JSON files

## Why is it important?

- Terraform suggest to add it's dependency lock file to VCS, as it helps with reproducible installations
- Terraform support template files, with `.tftpl` suggested extension
- JSON files may be used for ausiliary content required by Terraform code.

Further discussion available in #269.

## Checklist

## Related issues

- Closes #269


endorama:
@mtojek I have a question about the changelog. The file header says 
```
[...]
[...]
```

but it seems the opposite is true? Am I expected to append to `v1.4.1` as is still unreleased or create a new version?

jsoriano:
> but it seems the opposite is true? Am I expected to append to `v1.4.1` as is still unreleased or create a new version?

Put it at the bottom of the 1.4.1 section, yes :)

versions/1/changelog.yml
```diff
@@ -10,6 +10,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/265
+  - description: Support ausiliary files in terraform deploy spec
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/272
```
jsoriano, line 15: :+1:

endorama:
I added a `deploy_terraform` test package, but I'm not sure is being used :thinking: given that I also added the same files to package `good`, should I remove it?

jsoriano:
> I added a `deploy_terraform` test package, but I'm not sure is being used thinking given that I also added the same files to package `good`, should I remove it?

Test packages are used in `code/go/pkg/validator/validator_test.go`, you will need to add the package there.

test/packages/deploy_terraform/_dev/deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,3 @@
+version: '2.3'
+services:
+  hello_world:
```
jsoriano, line 3: Nit. Is this file needed here? I guess it doesn't do any harm though.

test/packages/good/data_stream/foo/_dev/deploy/tf/env.yml
```diff
@@ -2,8 +2,5 @@ version: '2.3'
 services:
   terraform:
     environment:
-      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
-      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
-      - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
-      - AWS_PROFILE=${AWS_PROFILE}
-      - AWS_REGION=${AWS_REGION:-us-east-1}
\ No newline at end of file
+      - FOO=${BAR:-default-value}
```
jsoriano, line 5: Unrelated changes?

versions/1/changelog.yml
```diff
@@ -10,6 +10,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/265
+  - description: Support ausiliary files in terraform deploy spec
```
jsoriano, line 13: Nit. Should it be "auxiliary"?
```suggestion
  - description: Support auxiliary files in terraform deploy spec
```

versions/1/_dev/deploy/tf/spec.yml
```diff
@@ -5,9 +5,22 @@ spec:
...
       type: file
       name: "env.yml"
       contentMediaType: "application/x-yaml"
       required: false
       $ref: "./env.spec.yml"
+    - description: Fixture JSON files
+      pattern: '^*.json$'
+      type: file
+      contentMediaType: "application/json"
+      required: false
```
jsoriano, line 26: Perhaps we end-up needing `additionalContents: false` if we find scenarios that would need arbitrary files. But we can reduce restrictions in the future if we find these cases.

jsoriano (dismissed):
Some nitpicking, but LGTM in general.

versions/1/_dev/deploy/tf/spec.yml
```diff
@@ -5,9 +5,22 @@ spec:
       pattern: '^*.tf$'
       type: file
       required: true
+    - description: Terraform dependency lock file
+      pattern: ".terraform.lock.hcl"
+      type: file
+      required: true
```
jsoriano, line 11: Btw, this may be breaking for current packages. At least we will need to add this file to the AWS package when updating the spec there.
https://github.com/elastic/integrations/tree/8e600daead4b89e97271e4976b984722c8db7c74/packages/aws/data_stream/ec2_metrics/_dev/deploy/tf

test/packages/deploy_terraform/_dev/deploy/docker/docker-compose.yml
endorama: leftover from copying over the `deploy_docker` module, thank you for catching this.

test/packages/good/data_stream/foo/_dev/deploy/tf/env.yml
endorama: No but given the file is not required I'll remove it

versions/1/_dev/deploy/tf/spec.yml
endorama: Nice find, actually I don't want to introduce a breaking change here. Do we have a process here? I will make this not required, but I think it make sense to have it everywhere. Is there a way to notify users of this change? For example outputting a message: "this field is going to default to true from version x.y.z"

jsoriano: Maybe we can add this as optional by now, add after some time make it required. At the moment I think that only the AWS package uses it, so not many users affected :)

mtojek: Yeah, let's mark it optional for now.

mtojek (approved):
LGTM

jsoriano (approved):

Issue #263, opened 2022-01-29
[Change Proposal] csp-rule asset type


eyalkraft:

orouz:
has a general asset type - `saved_object` been considered? 

i imagine it comes with considerations of its own, but it would solve our problem and perhaps others too.

brokensound77:
Thanks for tagging me. I would be in favor of option one in order to avoid mixing rule sets which are completely unrelated. We still push regularly to that integration to support our out of band releases for the detection engine.

joshdover:
Please see my comment here on a path forward that requires less architectural change: https://github.com/elastic/kibana/issues/124149#issuecomment-1030013986

joshdover:
> One thing I'd like to note is that these objects should probably not be [Kibana Space](https://www.elastic.co/guide/en/kibana/master/xpack-spaces.html)-specific since they serve as templates for the entire cluster. We do not have a field in the spec for this today, but I'd recommend we add a `namespaceType` field which defaults to `single` and also allows `agnostic` which would match Kibana's naming of this feature.

Thinking about this more, I don't think we need a spec change for this. It's actually specified by the Saved Object type definition in the Kibana code itself and I don't think we need any special properties on the actual objects in the package in order to accomplish this. I think we can defer on this and revisit if I'm indeed wrong.

kfirpeled:
Hey @joshdover, if the integrations are kibana space specific I think it would be best to keep the csp templates kibana-space specific as well. 

Otherwise, updating the integration version on one space would effect also on the other space. which might not be expected by the user.

if integrations are not kibana-space specific, I'm less worried about it

eyalkraft:
Following the decisions made on the related [issue](https://github.com/elastic/kibana/issues/124149#issuecomment-1032762790), Renaming this issue and opening a PR to close it.

PR #276, opened 2022-02-10
Add kibana/csp-rule-remplate asset type

eyalkraft:
## What does this PR do?

Add a `csp-rule-template` asset type for rules that will be delivered to Security » Cloud security posture » Rules part of Kibana.

## Why is it important?

This change is crucial for the cloud security posture solution, As integration packages for different benchmarks must contain the related `csp-rule-template`s for concrete rules creation in the cloud-security-posture plugin in Kibana.  

## Checklist

- [x] Passed `make update && make check && make test` locally
- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective (Added to the good package example).

## Related issues

- Closes https://github.com/elastic/package-spec/issues/263
- Relates https://github.com/elastic/kibana/issues/124149

cc @joshdover @mtojek @ruflin @kfirpeled

### DIFF
--- a/test/packages/good/kibana/csp_rule_template/good-csp-rule-template-abc-1.json
+++ b/test/packages/good/kibana/csp_rule_template/good-csp-rule-template-abc-1.json
@@ -1,0 +1,22 @@
+{
+  "attributes": {
+    "benchmark_rule_id": "1.1.1",
+    "rego_rule_id": "cis_k8s.cis_1_1_1",
+    "name": "Ensure that the API server pod specification file permissions are set to 644 or more restrictive (Automated)",
+    "description": "'Disable anonymous requests to the API server",
+    "rationale": "When enabled, requests that are not rejected by other configured authentication methods\nare treated as anonymous requests. These requests are then served by the API server. You\nshould rely on authentication to authorize access and disallow anonymous requests.\nIf you are using RBAC authorization, it is generally considered reasonable to allow\nanonymous access to the API Server for health checks and discovery purposes, and hence\nthis recommendation is not scored. However, you should consider whether anonymous\ndiscovery is an acceptable risk for your purposes.",
+    "impact": "Anonymous requests will be rejected.",
+    "default_value": "By default, anonymous access is enabled.",
+    "remediation": "Edit the API server pod specification file /etc/kubernetes/manifests/kubeapiserver.yaml on the master node and set the below parameter.\n--anonymous-auth=false",
+    "enabled": true,
+    "muted": false,
+    "tags": [
+      "Kubernetes",
+      "Containers"
+    ],
+    "benchmark": { "name": "CIS Kubernetes", "version": "1.4.1" },
+    "severity": "low"
+  },
+  "id": "good-csp-rule-template-abc-1",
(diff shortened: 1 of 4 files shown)


test/packages/good/kibana/csp_rule_template/good-csp-rule-template-abc-1.json
```diff
@@ -0,0 +1,22 @@
...
+    "rego_rule_id": "cis_k8s.cis_1_1_1",
+    "name": "Ensure that the API server pod specification file permissions are set to 644 or more restrictive (Automated)",
+    "description": "'Disable anonymous requests to the API server",
+    "rationale": "When enabled, requests that are not rejected by other configured authentication methods\nare treated as anonymous requests. These requests are then served by the API server. You\nshould rely on authentication to authorize access and disallow anonymous requests.\nIf you are using RBAC authorization, it is generally considered reasonable to allow\nanonymous access to the API Server for health checks and discovery purposes, and hence\nthis recommendation is not scored. However, you should consider whether anonymous\ndiscovery is an acceptable risk for your purposes.",
+    "impact": "Anonymous requests will be rejected.",
+    "default_value": "By default, anonymous access is enabled.",
+    "remediation": "Edit the API server pod specification file /etc/kubernetes/manifests/kubeapiserver.yaml on the master node and set the below parameter.\n--anonymous-auth=false",
+    "enabled": true,
+    "muted": false,
```
eyalkraft, line 12: @joshdover these fields are here as default values for the concrete rules in Kibana.

versions/1/kibana/spec.yml
```diff
@@ -74,6 +74,15 @@
...
         pattern: '^.+\.json$'
+    - description: Folder containing CSP rule templates
+      type: folder
+      name: "csp_rule_template"
+      required: false
+      contents:
+      - description: An individual CSP rule template file for the cloud security posture management solution
+        type: file
+        contentMediaType: "application/json"
+        pattern: '^.+\.json$'
```
jsoriano, line 85: A question From my lack of knowledge on these features :slightly_smiling_face:

I see there is already a directory for `security_rule`s, how are these csp rule templates different to the security rules above? Are they installed using the same APIs?

I am wondering if we could use the same `security_rule` directory for both kinds of files if they are handled the same way on installation, or use a subdirectory in `security_rule` for these CSP templates if they are different assets, but belong to the same feature.

eyalkraft: Hey @jsoriano, Thanks for reviewing!
This option (using the existing `security-rule` asset) did came up in the [issue](https://github.com/elastic/package-spec/issues/263), but since they are completely unrelated in Kibana (all `security-rule`s are part of the detection engine, `csp-rule-template`s are relates to the could security posture plugin) and require different installation process (usage of [hooks](https://github.com/elastic/kibana/issues/124149#issuecomment-1032762790)) it seemed like everyone agree they should be separate types of assets.

jsoriano: Ok, makes sense, thanks!

jsoriano (approved):

Issue #162, opened 2021-04-23
Define what the limits of a package are


ruflin:
Over the past month, quite a few asset types were added to the package-spec. Recently a discussion was raised around what is allowed in a package and what not. Can API Keys be shipped as part of a package, user definitions etc.? 

This issue is to discuss on how we could find and define a general concept on what fits and what doesn't.

jlind23:
Linked to https://github.com/elastic/package-registry/issues/670

mtojek:
@jsoriano Do you think we can close this one?

jsoriano:
I think this issue is more broad, for example we are not answering to "Can API Keys be shipped as part of a package, user definitions etc.?" Though not sure how we can check that thinks like API keys are not shipped by mistake.

@ruflin, wdyt, is it enough with the limits implemented in https://github.com/elastic/package-spec/pull/278, or you think that we are still missing something as part of this issue?

ruflin:
The issue is broader indeed. The goal is to find some logic / guidelines on which assets fit into the package-spec and which ones not. Example: Any asset needed for the Elastic Stack to ingest and process very likely fits into package-spec. Any assets that reconfigures the Elasticsearch cluster likely does not fit. There will always be grey area, but it would be good to have some logic that can be applied

mtojek:
I see now. I was confused as the issue title is nearly the same as the other issue ("Package Storage v2 - Define limits for a package").

Let's rename this issue to be more meaningful and note questions we should answer. Regarding the original question:

> Can API Keys be shipped as part of a package, user definitions etc.

Did somebody have concerns about this? I admit that I have never seen that request, but maybe I'm missing the context. If this is mostly a question, what fits in, what doesn't, I suppose it's more a "write guidelines" issue?

ruflin:
This question came out of the conversations around making packages more "native" in Elasticsearch and API Keys was the obvious example where we thought: This should never be delivered in a package. The reason this issue exists as we could come up with a quick way to describe the boundaries of a package. So here we are. And yes, it is more about guideline / vision on what a package should be and become.

jsoriano:
> Let's rename this issue to be more meaningful and note questions we should answer.

Renamed.

PR #277, opened 2022-02-16
Remove unused content attribute from the spec

jsoriano:
## What does this PR do?

Remove unused attribute from common folder items spec. 

## Why is it important?

Cleanup.

To implement limits checking (https://github.com/elastic/package-spec/issues/162) I am considering to create a set of structures, separated to the spec ones, that keep information about the files found, such as size, and maybe their content. Such an attribute would belong there.

## Checklist

- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective. -> Tests pass after removing this code.
- [x] I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/main/versions/1/changelog.yml). -> No user-facing change.

### DIFF
--- a/code/go/internal/validator/folder_item_spec.go
+++ b/code/go/internal/validator/folder_item_spec.go
@@ -79,8 +79,6 @@ func (s *folderItemSpec) validate(fs fs.FS, folderSpecPath string, itemPath stri
 	if s.Ref != "" {
 		schemaPath := filepath.Join(filepath.Dir(folderSpecPath), s.Ref)
 		schemaLoader = yamlschema.NewReferenceLoaderFileSystem("file:///"+schemaPath, fs)
-	} else if s.Content != nil {
-		schemaLoader = yamlschema.NewRawLoaderFileSystem(s.Content, fs)
 	} else {
 		return nil // item's schema is not defined
 	}
--- a/code/go/internal/validator/common_spec.go
+++ b/code/go/internal/validator/common_spec.go
@@ -10,10 +10,9 @@ import (
 )
 
 type commonSpec struct {
-	AdditionalContents bool                   `yaml:"additionalContents"`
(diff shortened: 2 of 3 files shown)


code/go/internal/validator/folder_item_spec.go
```diff
@@ -79,8 +79,6 @@ func (s *folderItemSpec) validate(fs fs.FS, folderSpecPath string, itemPath stri
 	if s.Ref != "" {
 		schemaPath := filepath.Join(filepath.Dir(folderSpecPath), s.Ref)
 		schemaLoader = yamlschema.NewReferenceLoaderFileSystem("file:///"+schemaPath, fs)
-	} else if s.Content != nil {
```
mtojek, line 82: I'm trying to remind why did we need this attribute and IIRC it's to embed the spec instead of referencing. For example:

```
  - description: The package's CHANGELOG file
    type: file
    name: "changelog.yml"
    required: true
    $ref: "./changelog.spec.yml"
```

could be replaced with:

```
  - description: The package's CHANGELOG file
    type: file
    name: "changelog.yml"
    required: true
    content:
        <content of changelog.spec.yml, but embedded>
```

Thanks for checking if we're using this property in spec and if we don't, I guess it's safe to get rid of it.

mtojek (approved):
LGTM

code/go/internal/validator/folder_item_spec.go
jsoriano: Umm, well, we can always revert this change if we find a situation where we need embedded specs :slightly_smiling_face:

mtojek: I might be wrong, but it was intended for really short spec files. I guess it's safe to remove it :)

PR #278, opened 2022-02-17
Implement package limits

jsoriano:
## What does this PR do?

* Implement package limits:
  - [x] Global - Package size - 250MB
  - [x] Global - Total filesize - 150MB
  - [x] Global - Total number of files in a package - 65535
  - [x] Global - Total number of files in a directory - 65535
  - [x] Local - Number of data streams - 500
  - [x] Local - Number of fields in data stream - 1024
  - [x] Local - Size of graphic resources - 3MB
  - [x] Local - Size of configuration files (yml files, manifests, fields) - 5MB
  - [x] Local - Ingest pipelines - 3MB
  - [x] Local - Config templates (hbs files) - 2MB 
* Add helper types for content media types and file sizes.
* Refactor `folderItemSpec.validate` and `loadItemContent` to split its responsibilities.

## Why is it important?

Ensure that packages and the resources they contain fit under controlled limits.

## Checklist

## Related issues

- Relates to https://github.com/elastic/package-spec/issues/162


code/go/internal/validator/semantic/format_checkers.go
```diff
@@ -37,10 +42,18 @@ func (r relativePathChecker) IsFormat(input interface{}) bool {
...
 	path := filepath.Join(r.currentPath, asString)
-	_, err := os.Stat(path)
+	info, err := os.Stat(path)
 	if err != nil {
 		return false
 	}
+
+	// TODO: It happens that we want the same max size for all the files we reference with
+	// relative paths, but it'd be better if we could find a way to parameterize format
+	// checkers so we can configure specific max sizes, and we can provide better feedback.
```
jsoriano, line 52: Library we use for jsonschema can only be customized with format checkers, what is not enough to add further attributes that could be passed as parameters to formatters and allow schemas like this one:
```
          src:
            description: Relative path to the screenshot's image file.
            type: string
            format: relative-path
            maxSize: 3MB
            examples:
              - "/img/apache_httpd_server_status.png"
```
Or even better, define new types, so we could define schemas like this one:
```
          src:
            description: Relative path to the screenshot's image file.
            type: relative-path
            maxSize: 3MB
            examples:
              - "/img/apache_httpd_server_status.png"
```

I wonder if we should consider a more customizable library, such as these ones:
* https://pkg.go.dev/github.com/qri-io/jsonschema
* https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema

code/go/internal/validator/folder_item_spec_errors.go
```diff
@@ -4,11 +4,15 @@
...
+	"github.com/elastic/package-spec/code/go/internal/validator/semantic"
+)
 
 func adjustErrorDescription(description string) string {
 	if description == "Does not match format '"+semantic.RelativePathFormat+"'" {
-		return "relative path is invalid or target doesn't exist"
+		return fmt.Sprintf("relative path is invalid, target doesn't exist or is bigger than %s", semantic.RelativePathFileMaxSize)
 	} else if description == "Does not match format '"+semantic.DataStreamNameFormat+"'" {
 		return "data stream doesn't exist"
 	}
```
jsoriano, line 18: Avoiding the need of these adjustments would be another reason to look for another jsonschema library with better customization options.

jsoriano:
@mtojek ideas for testing this? I wouldn't like to commit files that exceed the limits :slightly_smiling_face:

An idea could be to add a generator of packages for testing, and generate the packages that exceed the limits on test time.

code/go/internal/validator/semantic/types.go
```diff
@@ -43,14 +44,32 @@ func validateFields(pkgRoot string, validate validateFunc) ve.ValidationErrors {
 			vErrs = append(vErrs, errors.Wrapf(err, `file "%s" is invalid: can't unmarshal fields`, fieldsFile))
 		}
 
-		for _, u := range unmarshaled {
-			errs := validate(fieldsFile, u)
+		errs := validateNestedFields("", fieldsFile, unmarshaled, validate)
```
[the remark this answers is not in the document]
jsoriano: https://github.com/elastic/package-spec/pull/279

code/go/internal/spectypes/contenttype.go
```diff
@@ -0,0 +1,66 @@
...
+	return t.unmarshalString(value.Value)
+}
+
+func (t *ContentType) unmarshalString(text string) error {
+	mediatype, params, err := mime.ParseMediaType(text)
+	if err != nil {
+		return err
+	}
+	if mime.FormatMediaType(mediatype, params) == "" {
+		// Bug in mime library? Happens when parsing something like "0;*0=0"
```
mtojek, line 54: I don't have any deeper knowledge of this. Did you find it by accident with unit tests?

code/go/internal/validator/semantic/format_checkers.go
mtojek: If so, then let's treat it as a separate issue. I remember that the initial idea was to have an option to deliver a spec parser for different languages (MVP: Go, JS), but we failed to build it.

code/go/internal/spectypes/filesize.go
```diff
@@ -0,0 +1,121 @@
...
+
+// MarshalJSON implements the json.Marshaler interface for FileSize.
+func (s FileSize) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + s.String() + `"`), nil
+}
+
+// MarshalYAML implements the yaml.Marshaler interface for FileSize.
+func (s FileSize) MarshalYAML() (interface{}, error) {
+	return s.String(), nil
+}
```
mtojek, line 46: I admit I'm a bit lost here. Do we need to implement marshallers to check the file size?

code/go/internal/validator/common_spec.go
```diff
@@ -7,12 +7,16 @@ package validator
...
 
 type commonSpec struct {
-	AdditionalContents bool             `yaml:"additionalContents"`
-	Contents           []folderItemSpec `yaml:"contents"`
-	DevelopmentFolder  bool             `yaml:"developmentFolder"`
+	AdditionalContents bool               `yaml:"additionalContents"`
+	Contents           []folderItemSpec   `yaml:"contents"`
+	DevelopmentFolder  bool               `yaml:"developmentFolder"`
+	MaxSize            spectypes.FileSize `yaml:"maxSize"`
+	MaxContents        int                `yaml:"maxContents"`
```
mtojek, line 19: It looks like inconsistent: maxSize is spectype, but maxContents is int.

code/go/internal/validator/folder_item_content.go
```diff
@@ -9,59 +9,122 @@ import (
...
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v3"
+
+	ve "github.com/elastic/package-spec/code/go/internal/errors"
+	"github.com/elastic/package-spec/code/go/internal/spectypes"
+)
+
+const (
+	maxFileSize              = 150 * spectypes.MegaByte
+	maxConfigurationFileSize = 5 * spectypes.MegaByte
```
mtojek, line 23: I'm wondering if there shouldn't be a general YAML file for default settings like this.

code/go/internal/validator/folder_item_spec.go
```diff
@@ -16,20 +16,21 @@ import (
...
-	ContentMediaType  string   `yaml:"contentMediaType"`
-	ForbiddenPatterns []string `yaml:"forbiddenPatterns"`
-	Name              string   `yaml:"name"`
-	Pattern           string   `yaml:"pattern"`
-	Required          bool     `yaml:"required"`
-	Ref               string   `yaml:"$ref"`
-	Visibility        string   `yaml:"visibility" default:"public"`
+	Description       string                 `yaml:"description"`
+	ItemType          string                 `yaml:"type"`
+	ContentMediaType  *spectypes.ContentType `yaml:"contentMediaType"`
```
mtojek, line 27: Why exactly do we need the `spectypes.ContentType`?

mtojek (commented):
> @mtojek ideas for testing this? I wouldn't like to commit files that exceed the limits 🙂
An idea could be to add a generator of packages for testing, and generate the packages that exceed the limits on test time.

My first thought would be mocking, so you don't have to generate big files even temporarily. You check if you can mimic the FS interface and return some big `size()`.

Another alternative would be setting really low limits just for testing.

---

First round of the code review done.

I admit that I got scared with this PR... It was supposed to be an easy size check to implement, but we ended up with complex "spectyping". I'm wondering if we shouldn't rethink the idea or adjust the existing design.

What we can consider here is a "file limits" file, which compacts all global limitations we have for folders and their content, so that it's easy accessible and not hidden in the Go code. Keep in mind about the original (but dead today) portability idea :)

code/go/internal/spectypes/contenttype.go
jsoriano: Gave a try to 1.18's fuzzy testing in https://github.com/elastic/package-spec/pull/278/files#diff-2a7131ad04e8aa8fcd479f359de4357a6dbe6316b61c5f6966079c4c486611f2 and found this.

code/go/internal/spectypes/filesize.go
jsoriano: The marshalers are not really needed here, I added them for completion, they come almost for free after implementing `String()`. `String()` is helpful to show the file sizes in error messages.

Well, thinking twice they may be needed in the places where we convert from JSON to YAML by unmarshaling and marshaling.

code/go/internal/validator/common_spec.go
jsoriano: `maxSize` is `spectypes.FileSize`, a kind of int.

code/go/internal/validator/folder_item_content.go
jsoriano: Yep, it can make sense to make them part of the folder spec, this would allow to define these settings per directory too. I will give it a try as part of the implementation of the global limits.

code/go/internal/validator/folder_item_spec.go
jsoriano: I found the need of having the content type parsed in different places, then I decided to parse it directly when unmarshaling.

code/go/internal/validator/semantic/format_checkers.go
jsoriano: Yes, this is not needed by now, but only because it happens that all the relative paths we need to check have the same limits.

versions/1/data_stream/spec.yml
```diff
@@ -1,5 +1,6 @@
 spec:
   additionalContents: false
+  maxContents: 500
```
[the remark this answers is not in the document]
jsoriano: I used `Contents` as is the same thing referenced by `additionalContents`, but not a strong opinion.

versions/1/data_stream/spec.yml
```diff
@@ -10,6 +11,7 @@ spec:
     - description: A data stream's manifest file
       type: file
       contentMediaType: "application/x-yaml"
+      maxSize: 5MB
```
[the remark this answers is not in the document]
jsoriano: `sizeLimit`? :slightly_smiling_face:  Or to what would refer `totalSize` in files?

code/go/internal/spectypes/filesize.go
mtojek: Maybe add a comment there (why we need marshallers), otherwise every next person will be lost while reading it :)

code/go/internal/validator/common_spec.go
mtojek: Yes, I'm wondering if we want to switch to spectypes, maybe MaxContents should be also a spectype.

code/go/internal/validator/folder_item_spec.go
mtojek: ok, so the only thing we need to remember is to not load the file content if we expect to have packages weighing a few hundred of megabytes.

code/go/internal/validator/common_spec.go
jsoriano: Well, I am adding spectypes for values that can benefit of special formatting or parsing, but a count of elements is ok as a plain int.

versions/1/data_stream/spec.yml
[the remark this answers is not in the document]
jsoriano: I am renaming all limits in the spec to `somethingLimit`.

code/go/internal/validator/folder_item_spec.go
jsoriano: Already done, this was one of my main motivations to refactor [`loadItemContent`](https://github.com/elastic/package-spec/pull/278/files#diff-533ad86ff7ed72649d73fc50169d7666aa1635da1376499b8cfbbf496be9ae03L73) and split its responsibilities.

Now the content of files is only load [when needed](https://github.com/elastic/package-spec/pull/278/files#diff-533ad86ff7ed72649d73fc50169d7666aa1635da1376499b8cfbbf496be9ae03R105), and after [checking the limits](https://github.com/elastic/package-spec/pull/278/files#diff-533ad86ff7ed72649d73fc50169d7666aa1635da1376499b8cfbbf496be9ae03R73-R86).

The only problem the new approach is that I am doing redundant `Stats`, but this should be much better that loading the content of all files, and file info could be reused in a further refactor. In principle I would leave further refactors there for a follow up.

jsoriano:
I will move https://github.com/elastic/package-spec/pull/278/commits/b524c927a69e3d86dd7eaa185b528d42393f9806 to a different PR.

code/go/internal/validator/semantic/validate_fields_limits.go
```diff
@@ -0,0 +1,65 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	ve "github.com/elastic/package-spec/code/go/internal/errors"
+	"github.com/elastic/package-spec/code/go/internal/fspath"
+)
+
+const maxFieldsPerDataStream = 1024
```
jsoriano, line 18: Not sure if moving this too to somewhere in the spec :thinking:

mtojek: I would consider every number/count/etc. as a constant value, so - in the spec :)

jsoriano: Moved https://github.com/elastic/package-spec/pull/278/commits/9453025cf021e269e8aa36a1fd40af7d73a35b5f

jsoriano:
Tests and changelog added, opening for review.

mtojek:
> Tests and changelog added, opening for review.

Do you think you can mark it as ready for review or there is a potential blocker?

jsoriano:
> > Tests and changelog added, opening for review.
> 
> Do you think you can mark it as ready for review or there is a potential blocker?

Ah yes, I wanted to do that after my previous comment :facepalm:

code/go/internal/spectypes/filesize.go
```diff
@@ -0,0 +1,125 @@
...
+
+// MarshalJSON implements the json.Marshaler interface for FileSize, it returns
+// the string representation in a format that can be unmarshaled back to an
+// equivalent value.
+func (s FileSize) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + s.String() + `"`), nil
+}
+
+// MarshalYAML implements the json.Marshaler interface for FileSize, it returns
```
mtojek, line 45: yaml.Marshaller?

code/go/internal/spectypes/filesize.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+const (
+	byteString     = "B"
+	kiloByteString = "KB"
+	megaByteString = "MB"
+)
+
+// FileSize represents the size of a file.
+type FileSize uint64
```
mtojek, line 30: You can also define its interfaces:

```
var _ json.Mashaller = new(FileSize)
```

same for YAML?

code/go/internal/spectypes/filesize.go
```diff
@@ -0,0 +1,125 @@
...
+	if len(match) < 3 {
+		return fmt.Errorf("invalid format for file size (%s)", text)
+	}
+
+	q, err := parseFileSizeInt(match[1])
+	if err != nil {
+		return fmt.Errorf("invalid format for file size (%s): %w", text, err)
+	}
+
+	unit := match[2]
```
mtojek, line 95: Maybe we should use a library like: https://github.com/dustin/go-humanize

code/go/internal/validator/common_spec.go
```diff
@@ -5,14 +5,51 @@
...
+	ConfigurationSizeLimit spectypes.FileSize `yaml:"configurationSizeLimit"`
+
+	// Limit to files referenced as relative paths (images).
+	RelativePathSizeLimit spectypes.FileSize `yaml:"relativePathSizeLimit"`
+
+	// Maximum number of fields per data stream, can only be set at the root level spec.
+	FieldsPerDataStreamLimit int `yaml:"fieldsPerDataStreamLimit"`
+}
+
+func (l *commonSpecLimits) update(o commonSpecLimits) {
```
mtojek, line 44: In general, I'm hesitant to use reflection unless we can't achieve something any other way round. I'm wondering if we can go with a standard copy operation between fields.

code/go/internal/validator/folder_item_content.go
```diff
@@ -9,59 +9,115 @@ import (
...
-	// Usually it can be solved by adding the .empty file. There is no media type defined for this file.
-	// It's expected of the file with media type defined not to be empty - the file can be marked as non-required.
-	if len(itemData) == 0 && mediaType != "" {
-		return nil, errors.New("file is empty, but media type is defined")
+func validateContentType(fsys fs.FS, path string, contentType spectypes.ContentType) error {
+	switch contentType.MediaType {
+	case "application/x-yaml":
+		v, _ := contentType.Params["require-document-dashes"]
+		requireDashes := (v == "true")
+		if requireDashes {
```
mtojek, line 36: FYI I remember there was a problem with Kibana validating YML ingest pipelines. It failed when a document didn't have `---`. It shouldn't affect validation if moved here, right?

code/go/internal/validator/folder_item_content.go
```diff
@@ -9,59 +9,115 @@ import (
...
+	var sizeLimit spectypes.FileSize
+	switch contentType.MediaType {
 	case "application/x-yaml":
-		// TODO Determine if special handling of `---` is required (issue: https://github.com/elastic/package-spec/pull/54)
-		if v, _ := params["require-document-dashes"]; v == "true" && !bytes.HasPrefix(itemData, []byte("---\n")) {
-			return nil, errors.New("document dashes are required (start the document with '---')")
-		}
+		sizeLimit = limits.ConfigurationSizeLimit
+	}
+	if sizeLimit > 0 && size > sizeLimit {
```
mtojek, line 86: Question: empty YAML file. Is it a valid file :) ?

code/go/pkg/validator/limits_test.go
```diff
@@ -0,0 +1,382 @@
...
+			} else {
+				assert.Error(t, err)
+			}
+		})
+	}
+}
+
+var _ fs.FS = &mockFS{}
+
+type mockFS struct {
```
mtojek, line 121: 👍

code/go/pkg/validator/limits_test.go
```diff
@@ -0,0 +1,382 @@
...
+	root *mockFile
+}
+
+func newMockFS() *mockFS {
```
mtojek, line 125: I know there is this [project](https://github.com/spf13/afero), but not sure if it isn't an overkill. WDYT?

code/go/internal/validator/semantic/format_checkers.go
```diff
@@ -38,19 +41,28 @@ func (r relativePathChecker) IsFormat(input interface{}) bool {
 	}
 
 	path := filepath.Join(r.currentPath, asString)
-	_, err := fs.Stat(r.fsys, path)
+	info, err := fs.Stat(r.fsys, path)
 	if err != nil {
 		return false
 	}
+
+	// TODO: It happens that we want the same max size for all the files we reference with
```
mtojek, line 49: That's true as format checkers are a hackish solution, adapted to our needs. It's a gap in the original parser.

mtojek (commented):
I left a few minor comments to be addressed/responded to, but nothing serious. You did a pretty good job adapting the spec.

What makes me worried is the fact, that the implementation will become more tangled than it was before. I think we should open a new issue to research alternative spec parsers to slim down the current codebase.

code/go/internal/spectypes/filesize.go
jsoriano: It could be a good idea to have less code here.

But there are a couple of small concerns I have with humanize:
* It is strict with the interpretation of unit prefixes (Kilo means 1000, KiByte needs to be used for 1024), we would need to change from MB to MiB for example.
* Related to previous point, given that the parser supports units with both prefixes, we may end up having a mix in the spec (250MB and 250MiB would be correctly parsed, but they don't mean the same).
* It rounds-up when giving the string representation ("1024B" and "1025B" are both converted to "1.0 KiB"). We need to be able to convert back and forth because we parse yaml and convert it to json, and we would be losing precision. We could workaround this problem by marshalling always to the number of bytes, and using `humanize` to parse and for `String()`. Or we could ignore this lose of precission.

Wdyt, should I give a try to humanize?

code/go/internal/validator/folder_item_content.go
jsoriano: Yes, this is still checked. We have a test (`missing_pipeline_dashes`) that would fail otherwise.

code/go/internal/validator/folder_item_content.go
jsoriano: There is a check some lines above that makes validation to fail on empty files that have a content type defined, I brought this from previous implementation, though I think that there are no tests for this.

The truth is that if a file is empty, it cannot satisfy the condition of having a content type :)

code/go/pkg/validator/limits_test.go
jsoriano: I considered it, but this provides a full in-memory filesystem. I didn't find it so suited for mocking responses. For example I think that with aferofs I would need to create files of a given size, and not only mock the result of `info.Size()`. Should I give it another try?

code/go/internal/validator/common_spec.go
jsoriano: I agree, but for this case I didn't find any other way around:
* Preinitializing the structs with default values and a standard copy is not possible, because these objects are created when unmarshaling the objects. And these "default" values are obtained in the same unmarshal operation.
* I could do some kind of unmarshaling in two passes, one to read the limits on the parent and preinitialize the children contents, and another one to unmarshal the children, bit this can be worse than using reflection 
* I could do the good old set field by field, but I consider this error prone, we may forget of adding a field here after adding it to the struct.

jsoriano:
> What makes me worried is the fact, that the implementation will become more tangled than it was before. I think we should open a new issue to research alternative spec parsers to slim down the current codebase.

I guess that with any "validation" feature we add we are adding more code to the mix, not sure what we can do about this...

code/go/pkg/validator/limits_test.go
mtojek: No, it's a good justification then. It looks you found the niche :) Let's leave it as is.

code/go/internal/validator/folder_item_content.go
mtojek: It seems to be more of a life-death discussion if the empty file is valid JSON/YAML :) Never mind, wanted to double-check if there isn't any bug around.

code/go/internal/validator/common_spec.go
mtojek: > I could do some kind of unmarshaling in two passes, one to read the limits on the parent and preinitialize the children contents, and another one to unmarshal the children, bit this can be worse than using reflection

Yes, I used to follow this option when interacting with complex structures, ignoring the performance issues. Regarding reflection, I'm not sure if it still works when we add slices or pointers. 

Let's stick to reflection, but maybe make sure if it's supported if we decide to build a WebAssembly out of the package-spec. It might be a blocker then.

jsoriano: > Yes, I used to follow this option when interacting with complex structures, ignoring the performance issues. Regarding reflection, I'm not sure if it still works when we add slices or pointers.

Yep, with more complex types this may not work, but this object is intended to store limits, that are most likely going to be numeric values.

> Let's stick to reflection, but maybe make sure if it's supported if we decide to build a WebAssembly out of the package-spec. It might be a blocker then.

I wouldn't expect compatibility issues because of using reflection, this is also used by core packages such as `encode/json`.

(In any case, I have tried, and tests pass with WebAssembly - `GOOS=js GOARCH=wasm go test -exec="$(go env GOROOT)/misc/wasm/go_js_wasm_exec" ./...`)

code/go/internal/spectypes/filesize.go
jsoriano: > > Related to previous point, given that the parser supports units with both prefixes, we may end up having a mix in the spec (250MB and 250MiB would be correctly parsed, but they don't mean the same).
> 
> they would be parsed correctly, but would mean different amount of units: 250_1024_1024 vs 250_1000_1000), so I would say it isn't a bug.

Well, 250MB are parsed as 250000000 bytes, while 250MiB are parsed as 262144000 bytes, a difference of 11MiB, or 12MB.

Let's revisit this in follow ups if needed.

jsoriano:
> What do you think about adding an issue to the backlog to write down all non-standard features we're using that are not present in the jsonschema library? It will help us understand what exactly are we looking for to make the spec clear.

Issue created for this https://github.com/elastic/package-spec/pull/278.

PR #279, opened 2022-02-17
Fix validation of dimension fields inside objects

jsoriano:
## What does this PR do?

There is a helper to visit a function with all the fields in a package, this helper was not visiting fields nested inside objects. Modify it to visit also nested fields, and "flatten" their names to give better information in error messages.

## Why is it important?

* Fixes validation of dimension fields.
* Error messages on fields validation provide now the full name of the field.
* Reduces the risk of incorrectly using this helper again (as was happening to me in https://github.com/elastic/package-spec/pull/278).

## Checklist

## Related issues

- Discovered in https://github.com/elastic/package-spec/pull/278.


mtojek (approved):
LGTM

PR #281, opened 2022-02-21
Disallow unknown properties on fields definitions

jsoriano:
## What does this PR do?

Disallow unknown properties on fields definitions.

## Why is it important?

Avoid typos as the ones reported in https://github.com/elastic/integrations/issues/2643.

## Checklist

## Related issues

- Related to https://github.com/elastic/integrations/issues/2643.


mtojek (commented):
I like this idea in general, but I remember we kept it open because we couldn't satisfy all features (multi-fields, example, etc.). Did you check how does this change behaves with existing packages?

jsoriano:
I have created https://github.com/elastic/package-spec/issues/282 to follow on this.

jsoriano:
> The only exception would be `example`s, for that we would need some kind of `type: any`, or we should change all of them to their string representation.

If we change all of them to the string representation, we also need to change the ones imported from ECS, there the examples can contain at least numbers.

mtojek:
I'm wondering if there aren't any problems with multi fields (see https://github.com/elastic/elastic-package/issues/678).

jsoriano:
> I'm wondering if there aren't any problems with multi fields (see [elastic/elastic-package#678](https://github.com/elastic/elastic-package/issues/678)).

We can validate them (see https://github.com/elastic/package-spec/pull/281/commits/ee17cfa684d7f52302b7a28e232d22fb16dbc98c), this other issue seems to be about importing external fields.

jsoriano:
Closing this, after https://github.com/elastic/package-spec/pull/314 we are in a better situation.

PR #283, opened 2022-02-23
Make all file operations through the filesystem interface

jsoriano:
## What does this PR do?

Migrate all the pending file operations to be done through the filesystem interface.

## Why is it important?

This will allow to validate the spec on packages that are stored on any resource for what an implementation of the filesystem interface is available.

This will be also useful to mock filesystems that can be used for testing some scenarios, like the ones needed for https://github.com/elastic/package-spec/pull/278.

## Checklist

- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective. _No test added, but as a refactor I focused on making the current ones pass_.
- [x] I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/main/versions/1/changelog.yml). _No user facing change_.

## Related issues

- Useful to test https://github.com/elastic/package-spec/pull/278.


code/go/internal/validator/folder_item_spec_format.go
```diff
@@ -1,30 +0,0 @@
...
-		return false
-	}
-
-	path := filepath.Join(r.currentPath, asString)
-	_, err := os.Stat(path)
-	if err != nil {
-		return false
-	}
-	return true
-}
```
jsoriano, line 30: This file is not used, this struct is also defined in `format_checkers.go`.

mtojek (dismissed):

jsoriano:
@mtojek sorry, I have just pushed a function to validate zips, happy to move this to a different PR, I only wanted to try this with a different implementation of `fs.FS`.

mtojek (approved):

Issue #23, opened 2020-08-13
Add Facility for deploying ElasticSearch Transform


nnamdifrankie:
**Background**

And the ability to deploy a defined ElasticSearch transform to be be deployed when a package is applied or upgraded. 

**Acceptance Criteria**

- [x] As a user I should be able to define an ElasticSearch transform as part of a package. 
- [x] The ElasticSearch transform should be started after it is added to the search database. 
- [x]  As a user I should be able to update the attributes of the transform including the name possibly. This should not result in two Transform running.
- [x] As a user we should be able to delete a transform through an update to the package. 
- [x] As a user I should be able to view statistics and information about a transform after deployment using the ElasticSearch API or Kibana if available. 

**Sample Transform Creation Statements Captured From Kibana Devtools**

```
PUT _transform/endpoint_host_metadata_transform
{
  "source": {
    "index": "metrics-endpoint.metadata-default"
  },
  "dest": {
    "index": "metrics-endpoint.metadata_current-default"
  },
  "pivot": {
    "group_by": {
      "agent.id": {
        "terms": {
          "field": "agent.id"
        }
      }
    },
    "aggregations": {
      "HostDetails": {
        "scripted_metric": {
          "init_script": "state.timestamp_latest = 0L; state.last_doc=''",
          "map_script": "def current_date = doc['@timestamp'].getValue().toInstant().toEpochMilli(); if (current_date > state.timestamp_latest) {state.timestamp_latest = current_date;state.last_doc = new HashMap(params['_source']);}",
          "combine_script": "return state",
          "reduce_script": "def last_doc = '';def timestamp_latest = 0L; for (s in states) {if (s.timestamp_latest > (timestamp_latest)) {timestamp_latest = s.timestamp_latest; last_doc = s.last_doc;}} return last_doc"
        }
      }
    }
  },
  "description": "collapse and update the latest document for each host",
  "frequency": "1m",
  "sync": {
    "time": {
      "field": "event.created",
      "delay": "60s"
    }
  }
}

POST _transform/endpoint_host_metadata_transform/_start
DELETE _transform/endpoint_host_metadata_transform
```

ycombinator:
@ruflin could use your input here.

>  As a user I should be able to define an ElasticSearch transform as part of a package.

[...]

PR #285, opened 2022-02-27
[POC] Validation of zip using javascript

jsoriano:
## What does this PR do?

Validates packages in zip format from JavaScript, using WebAssembly.

## Why is it important?

To use the same validation code in Go-based tooling and JavaScript components as Fleet.

## Checklist

## Related issues

-


joshdover:
Thanks for experimenting with this! One question I have is whether or not it would be possible to validate a package without having to save a zip file on disk. Ideally we could pass in a `Buffer` of bytes for the zip file contents. We can then stream these directly from the EPR response or from an uploaded package.

This could also be used as the basis for supporting development of bundled packages directly in the Kibana repo if we could easily access the ability to build a package from a directory. This could be done by using the CLI directly as well in build steps, but it'd be nice to only have way of interacting with elastic-package from our JS code in Kibana. cc @kpollich 

In terms of packaging and consuming from the Kibana repository, it'd be ideal if elastic-package distributed a `@elastic/package` NPM module that contained the built wasm and JS interface for accessing it in Node.js code. Would this be feasible to accommodate within the CI pipeline for elastic-package?

joshdover:
> Do you expect some use case where this would be needed apart of in build time? In build and development time I guess that it is ok to use the `elastic-package` binary.

Right this would be a build-time step, so using the elastic-package binary directly might be fine. Thought it might be simpler from a maintenance perspective if we had a single WASM target/NPM package that would support both use cases, but happy to explore using the binary directly.

.ci/Jenkinsfile
```diff
@@ -52,6 +52,7 @@ pipeline {
     stage('Test') {
       steps {
         cleanup()
+        installTools([ [tool: 'nodejs', version: '10.0' ] ])
```
jsoriano, line 55: This doesn't work.

joshdover:
> @joshdover I have added to the POC package validation using a `Uint8Array` as input. This allows to do client-side validation, I have also added a simple web server to show this.

Very nice, that should make consuming this from Kibana very easy. Note that we won't be able to start using this validation until https://github.com/elastic/package-spec/issues/23 is resolved since the Endpoint package has a transform and index template that aren't valid in the spec currently.

joshdover:
I got this working in Kibana, but it required a small adjustment to the interface to be able to use this from Node.js.

Here's the adjustments to this PR: https://github.com/jsoriano/package-spec/pull/1
Here's the POC PR for using this build from Kibana: https://github.com/elastic/kibana/pull/128076

jsoriano:
Great, thanks @joshdover for giving it a try :)

We would have to think about how to package this functionality for node, and what functionality to expose. Do you think it can be a problem to include wasm in Kibana?

joshdover:
> We would have to think about how to package this functionality for node, and what functionality to expose. Do you think it can be a problem to include wasm in Kibana?

Got this preliminary answer, and expecting @azasypkin to chime in as well. No immediate concerns for server-side usage, but client-side may need some research, but I don't think it's necessary: https://github.com/elastic/kibana/pull/128076#issuecomment-1072653591

jsoriano:
We had a discussion about this topic in an internal meeting of the Fleet team, some points we discussed:
* As result of the package validation, it would be nice to have structured errors, or a json, instead of the current string.
* Current wasm is about 12MB in size, check if there are ways to reduce it (there are some tips in the [docs](https://github.com/golang/go/wiki/WebAssembly#reducing-the-size-of-wasm-files)). 
* Exporting through globals may not be such a problem, but we need confirmation from Kibana team. We may give another though to explore options to expose the functionality.
* Package as npm module.
* Keeping the go instance running instead of starting and stopping it on every call may be ok.
* Expose information from packages, like for example the data stream information.

jsoriano:
If we manage to build with tinygo we may solve some of the problems mentioned above: smaller artifact, local exports, no need to keep the go instance running.

PR #307, opened 2022-03-30
Add transform to spec

eyalkraft:
## What does this PR do?

This is the first iteration change to add Elasticsearch [transform](https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html) to the spec.

This PR allows adding a transform asset to packages, with an optional destination index template to be installed with it. It also allows installing a transform without starting it.

Upon installation of a package that contains a transform asset, the **create** behavior is as follows:
1. **If** a destination index template is included:
    1. Create the destination index template.
        The name of the created index template will be derived from the data type, name of the package and the name of the containing folder,
        For example: `logs-example_package.example_name`.
    2. Add the appropriate `_meta` fields to mark the matching indices as managed by Fleet.
    3. **If** a fields directory is included:
        1. Use the defined fields as the destination index template mappings.
    4. **if** the destination index doesn't already exist:
        1. Initialize the index.
2. Create the transform in the `default` namespace using `kibana_system` privileges.
     The name of the created transform will be derived from the data type, the name of the package and the name of the containing folder,
     For example: `logs-example_package.example_name-default-0.0.1`.
     Namespace and version are added by fleet.
3. **If** the transform is configured to be started upon installation (no configuration defaults to true):
    1. Start the transform.

Upon upgrading a package that contains a transform asset, the **upgrade** behavior is as follows:
1. **If** a destination index template is included:
    1. Delete the destination index template.
    2. Matching indices are not deleted.
2. **If** the transform is started:
    1. Stop the transform.
3. Delete the transform.
4. From here, follow the creation behavior.

The recommended naming convention for the source and destination indices is the data stream naming scheme, or patterns matching the data stream naming scheme.
Data streams are named in the following way: `{type}-{dataset}-{namespace}`,
For example: `logs-example_package.example_name-default`.
For more information see the [introduction to the Elastic data stream naming scheme](https://www.elastic.co/blog/an-introduction-to-the-elastic-data-stream-naming-scheme).

## What doesn't this PR allow?

This list of possible changes will be further discussed in follow up issues:
1. Transform installations to different namespaces other than `default`.
2. ~~Installing a transform without starting it.~~ Included.
3. Installing a transform using any other user but `kibana_system` (such as the installing user/ some service user). This limits the source indices possible to use by package-installed transforms.
4. Smart upgrading trasforms (opposing to destroy and re-create).
5. Chained transform (or other assets) dependencies and controlling their installation order.
6. Installing stored scripts.

## Why is it important?

1. Required for the Cloud Security Posture integration
9. Already documented as supported

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/pull/2930
- Follow up https://github.com/elastic/kibana/issues/134321
- Relates https://github.com/elastic/security-team/issues/2857


versions/1/changelog.yml
```diff
@@ -249,3 +249,6 @@
   - description: Introduce "value" property for constant_keyword fields
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/194
+  - description: Add Elasticsearch transform support
```
mtojek, line 252: Please move it to the newest version. We won't release it as part of the oldest :)

ruflin:
I'm happy to see that we complete the package spec with the things we already support / have implemented. We should have a follow up discussion on how transforms should work in the future. Even thought it works today I would argue it is less then ideal. Many conversations happened in the past in https://github.com/elastic/package-spec/issues/23 and linked issues. There were also discussions happening with the team behind transforms on how it could be improved but not sure where we landed there. @kevinlog you might know?

versions/1/changelog.yml
```diff
@@ -248,4 +251,4 @@
     link: https://github.com/elastic/package-spec/pull/193
   - description: Introduce "value" property for constant_keyword fields
     type: enhancement
-    link: https://github.com/elastic/package-spec/pull/194
+    link: https://github.com/elastic/package-spec/pull/194
```
mtojek, line 254: Hm.. it looks like you removed a new line here.

versions/1/elasticsearch/spec.yml
```diff
@@ -28,3 +28,13 @@ spec:
...
+    - description: Folder containing Elasticsearch Transforms
+      type: folder
+      name: transform
+      required: false
+      contents:
+        - description: An Elasticsearch transform file
+          # https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html
+          type: file
+          contentMediaType: "application/json"
+          pattern: '^.+\.json$'
```
mtojek, line 40: Do you think we should apply any naming requirements? For example, to prepend a package name to the file. Is it possible that another package can override this transform?

versions/1/changelog.yml
eyalkraft: oops. Fixed.

versions/1/elasticsearch/spec.yml
eyalkraft: I'm not sure but I think the filename doesn't effect the created Transform at all.
As you can see in our integration (works locally) we have [two different transforms](https://github.com/elastic/integrations/pull/2930/files) in different directories with the same file name (`default.json`) and they don't override each other.
Probably @joshdover could explain better the exact meaning and implications the filenames and the containing directory names have.

mtojek: > endpoint security integration [transforms](https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform) look the same.

From the spec perspective, all Endpoint security integration [transforms](https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform) are illegal as they haven't been documented. This is the reason why you need to deal with it now :)

> I'm not sure but I think the filename doesn't effect the created Transform at all.
> As you can see in our integration (works locally) we have [two different transforms](https://github.com/elastic/integrations/pull/2930/files) in different directories with the same file name (default.json) and they don't override each other.

Let me elaborate here, what will happen if we have another package `foobar` with exact transforms and directories? When we install it, will it overwrite your transforms?

mtojek: cc @pzl

eyalkraft: > Let me elaborate here, what will happen if we have another package foobar with exact transforms and directories? When we install it, will it overwrite your transforms?

To my understanding the answer is no.
Seems like the id of the transform is a result of both the integration package name and the transform directory name.

Screen Shot 2022-03-31 at 15 04 25

Additionally after having a conversation with @CohenIdo about the naming of the transform files I understood he was instructed to name them `default.json` so I'll change the spec to reflect this requirement.

Anyways I'm sure Josh will shed more light over this topic.

kevinlog: @eyalkraft thanks for doing this, it's something I should have done some time back when my team originally added transforms.

@mtojek 

> I'm not sure but I think the filename doesn't effect the created Transform at all.
> Let me elaborate here, what will happen if we have another package `foobar` with exact transforms and directories? When we install it, will it overwrite your transforms?

Currently, when transforms are installed, the package name will be prepended (along with the directory path) to actual name of the transform.  Thus, it doesn't matter what the definition `.json` file is named.  So, as long as the packages themselves are named differently, I don't think there will be any collisions.  Here is what our transforms are named when installed.  (@pzl let me know if I have anything wrong here)

image

That being said, I am OK with any naming convention we want to enforce and we can update the names of files to match.

> We will enable the strict package validation soon and it will block the publishing of packages with undocumented (= illegal) features.

For clarity, adding this PR to the package spec will "document" the transform feature and will no longer make it illegal, correct?

Further, let us know when we can test against strict package enforcement so we can fix any other problems on our end and be prepared for when it is live.  I would ask that we can publish packages as is for the next couple weeks so that we can push our package with changes for the `8.2` release and in case we need to push anything additional to fix bugs.

mtojek: > That being said, I am OK with any naming convention we want to enforce and we can update the names of files to match.

That's the killer argument for this issue. I guess we can resolve the conversation. Thanks for jumping in, Kevin.

> For clarity, adding this PR to the package spec will "document" the transform feature and will no longer make it illegal, correct?

Yes, that's the idea.

> Further, let us know when we can test against strict package enforcement so we can fix any other problems on our end and be prepared for when it is live. I would ask that we can publish packages as is for the next couple weeks so that we can push our package with changes for the 8.2 release and in case we need to push anything additional to fix bugs.

@eyalkraft It would be great if we can copy Endpoint's transforms here as test cases to make sure they are aligned with new spec requirements. @kevinlog What would be the best source to copy from?

kevinlog: @eyalkraft @mtojek 

Either of the definitions here are good: https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform

joshdover (commented):
I've been meaning to sit down and write some thoughts on the open issue for this #23, but now is a good as time as any. I see two main problems with the existing implementation of transforms as used by the Endpoint package:

### An index template is required, but there is no explicit link in the package (or the package install implementation) between the template and the transform

Transforms require an index template as well (which is another 'illegal' feature), yet in Endpoint's current transform implementation there is no explicit link between the index template and transform. This creates problems when a transform's source index is another transform's destination index, as @CohenIdo noted in https://github.com/elastic/package-spec/issues/23#issuecomment-1062634206. In such a scenario, it'd be ideal for our package install code to be able to be aware of both the dependencies between transforms and between a transform and its index template. This would allow us to install the index templates and create the destination indices before setting up the transforms.

The CSP plugin is currently hacking around this limitation by manually creating the index template in their Kibana plugin. This means the index template is currently duplicated between the Kibana codebase and the CSP package source code.

### Adding support for namespace-specific transforms

We need to keep an eye towards this future requirement with an idea of how we will support this in the future (see private issue linked below). Currently the transforms contain the `default` namespace directly in the transform's name and source/dest indices. We should instead provide some sort of naming convention/pattern that does not include the namespace and rely on the package install code to add the namespace at package install time. In the future, we'll be providing an API to add namespaces and the associated resources, see https://github.com/elastic/kibana/issues/121118

---

I think we should redesign how this works a bit, what about supporting a yaml file like this?

```yaml
description: Latest Endpoint metadata document per host
frequency: 10s
source:
  # namespace would be appended as `metrics-endpoint.metadata-default*`
  index_prefix: metrics-endpoint.metadata-
  query: |
    {
      "range": {
        "@timestamp": {
          "gt": "now-90d/d"
        }
      }
    }
dest:
  # namespace would be appended as `metrics-endpoint.metadata_current-default`
  index_prefix: metrics-endpoint.metadata_current-
latest: |
  {
    "unique_key": [
      "elastic.agent.id"
    ],
    "sort": "@timestamp"
  }
sync: |
  {
    "time": {
      "field": "event.ingested",
      "delay": "10s"
    }
  }
```

One challenge here is that to solve the above problems we may also need to support the existing pattern that the Endpoint plugin uses. Alternatively, we could switch over to a new pattern in a given stack version and older Endpoint packages wouldn't be installable in those newer Stack versions. @kevinlog would this be acceptable?

kevinlog:
@joshdover 

> Alternatively, we could switch over to a new pattern in a given stack version and older Endpoint packages wouldn't be installable in those newer Stack versions. @kevinlog would this be acceptable?

This would be OK from our perspective since we tear down the old transform and install the new one on each package upgrade.  We always tie a new package release to a new stack release.  One thing to keep in mind is that the new package doesn't get installed until a user with proper permissions visits Fleet or Security in Kibana.  I know there is talk of installing packages in ES at upgrade time which would solve this drift.  I _still_ think we would be OK since the older package/assets would already be installed and we would only try to install the new package with the new patterns.

Let me know if the above makes sense.

kevinlog:
@joshdover 

> Is this still true? Fleet changed to auto upgrade packages on boot using the kibana_system user in 8.0 without requiring any users to visit the UI. Is there custom code in the Endpoint side that still needs to run separate from Fleet's package install/upgrade process?

You are correct - there is no other custom code on the Endpoint side.  I simply forgot about this change, apologies for the confusion.

test/packages/good/elasticsearch/transform/good_transform_abc_1/default.json
```diff
@@ -0,0 +1,46 @@
+{
```
mtojek, line 1: Could you please add also transforms referenced by Kevin in the other conversation?

eyalkraft: Done

versions/1/elasticsearch/spec.yml
```diff
@@ -28,3 +28,17 @@ spec:
...
           required: false
+    - description: Folder containing Elasticsearch Transforms
+      # https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html
+      type: folder
+      name: transform
+      required: false
+      contents:
+        - description: An Elasticsearch transform folder
+          type: folder
+          pattern: '^[a-z|_]+$'
```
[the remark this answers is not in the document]
eyalkraft: Done (and removed `|`)

versions/1/elasticsearch/spec.yml
eyalkraft: Added both

eyalkraft:
> @sophiec20: What naming conventions are being enforced with this PR? Can we review to ensure there is enough flexibility to cover other transform use cases.

As can be seen in the spec, this PR allows:

- Optionally adding a directory called `transform` under the `elasticsearch` directory to represent transform assets.
- The `transform` directory may contain subdirectories. Each one of these directories represent a transform asset to be installed upon package installation. The names of these subdirectories are allowed to contain lowercase letters, numbers and underscores. The directory name ends up being part of the transform id.
- Each on of these subdirectories (for example `elasticsearch/transform/example_transformation_01/`) should contain a `default.json` file representing the transformation itself.
- The spec says nothing about the source/destination indices of the transforms.

Regarding the naming of the created transform, as can be seen here:
Screen Shot 2022-03-31 at 15 04 25
The package assets that were installed:
```
endpoint/elasticsearch/transform/metadata_current/default.json
endpoint/elasticsearch/transform/metadata_united/default.json
cis_kubernetes_benchmark/elasticsearch/transform/latest/default.json
cis_kubernetes_benchmark/elasticsearch/transform/score/default.json
```

The ids of the created transforms:
```
endpoint.metadata_current-default-1.6.0-dev.0
endpoint.metadata_united-default-1.6.0-dev.0
cis_kubernetes_benchmark.latest-default-0.0.1
cis_kubernetes_benchmark.score-default-0.0.1
```

Note that it's not the spec that determines these ids but the package installation code in Kibana.

> @joshdover: I think we should redesign how this works a bit, what about supporting a yaml file like this?

Not going into the complicated details of your comment but just noting that mixing yaml and raw json inside as strings can result with a pretty ugly parsing logic, and might lead to mistakes while creating these assets. One thing that really helped us during the integration development is that the assets formats are (almost?) identical to the exported saved objects jsons. This makes sense since under the hood Kibana actually uses the "import saved object" API for the package assets installation, And this let us first create whatever asset we wanted our integration to have in Kibana using the UI and only then export it as Json and add it to our integration (no manual asset creation). 

<details>
<summary>This is your yaml parsed to json</summary>

```json
{
  "description": "Latest Endpoint metadata document per host",
  "frequency": "10s",
  "source": {
    "index_prefix": "metrics-endpoint.metadata-",
    "query": "{\n  \"range\": {\n    \"@timestamp\": {\n      \"gt\": \"now-90d/d\"\n    }\n  }\n}\n"
  },
  "dest": {
    "index_prefix": "metrics-endpoint.metadata_current-",
    "index_template": ""
  },
  "latest": "{\n  \"unique_key\": [\n    \"elastic.agent.id\"\n  ],\n  \"sort\": \"@timestamp\"\n}\n",
  "sync": "{\n  \"time\": {\n    \"field\": \"event.ingested\",\n    \"delay\": \"10s\"\n  }\n}"
}
```

</details>

I'm probably missing what is the value of having the raw index template json inside the yaml as string?

Alternatively, what's wrong with going for yaml alone?

<details>
<summary>Alternative yaml suggestion</summary>

```yaml
description: Latest Endpoint metadata document per host
frequency: 10s
source:
  # namespace would be appended as `metrics-endpoint.metadata-default*`
  index_prefix: metrics-endpoint.metadata-
  query:
      range:
        "@timestamp":
          gt: "now-90d/d"
dest:
  # namespace would be appended as `metrics-endpoint.metadata_current-default`
  index_prefix: metrics-endpoint.metadata_current-
  index_template:
latest:
    unique_key:
      - elastic.agent.id
    sort: "@timestamp"
sync:
    time:
      field: event.ingested
      delay: 10s
```

And as a json this looks like:
```json
{
  "description": "Latest Endpoint metadata document per host",
  "frequency": "10s",
  "source": {
    "index_prefix": "metrics-endpoint.metadata-",
    "query": {
      "range": {
        "@timestamp": {
          "gt": "now-90d/d"
        }
      }
    }
  },
  "dest": {
    "index_prefix": "metrics-endpoint.metadata_current-",
    "index_template": null
  },
  "latest": {
    "unique_key": [
      "elastic.agent.id"
    ],
    "sort": "@timestamp"
  },
  "sync": {
    "time": {
      "field": "event.ingested",
      "delay": "10s"
    }
  }
}
```

</details>

eyalkraft:
Also another suggestion, since:
1. This spec change is a blocker for the @elastic/cloud-posture-security team (specifically for https://github.com/elastic/integrations/pull/2930)
2. Transformations as integration-package assets already exist de-facto, and this PR changes the spec to reflect it

What do you think about merging this PR to align the spec with Kibana & the endpoint integration package, and opening another issue to continue discussing the problems and open ends with today's implementation of transformation assets?

I'll happily open the follow-up issue with all the information from this thread if you all agree.

@mtojek @ruflin @joshdover @kevinlog @sophiec20

joshdover:
> Controlling the name of the transform and its destination and source indices might not be applicable to all use cases. It might fit for the endpoint use case, but there are multiple future requests for packages which contain transforms (such as Host Risk Score and Beaconing, from a different part of Security). These do not currently conform to the Fleet naming conventions for data streams afaik.

@sophiec20 Is there a reason they can't conform to the data stream naming convention? It's a goal for all data ingested by Elastic integrations to conform to this convention as it provides many benefits across the entire architecture.

> Not going into the complicated details of your comment but just noting that mixing yaml and raw json inside as strings can result with a pretty ugly parsing logic, and might lead to mistakes while creating these assets.

> Alternatively, what's wrong with going for yaml alone?

@eyalkraft Yeah I'm not too particular on the mix of yaml and json in my example. Going pure YAML probably makes sense. The goal of my example was to demonstrate how accommodating my concerns around the associated index template and dynamic namespace could work.

> What do you think about merging this PR to align the spec with Kibana & the endpoint integration package, and opening another issue to continue discussing the problems and open ends with today's implementation of transformation assets?

If we're going to make breaking changes to how this works, I'd rather not add it to the spec now since then other packages may start to use this and we'd then need to support two ways of doing the same thing from the install side.

If we defer on adding it to the spec now, could CSP continue using the existing implementation in Kibana today and then accommodate a new updated pattern with a hard version requirement (`constraints.kibana.version`) in the future, similar to what Endpoint is able to accommodate? Can we make an exception in the Integrations repo CI to allow the CSP packages to merge with this violation for now?

CohenIdo:
Hey @joshdover, thanks for syncing everyone with the index-template issue. 
I want to add something else in the same context: 
When adding a new Transform and the source index does not exist we will also get an error, for the test, I created a Transform that uses `logs-foo` as a source index and when Installed the package I got the next error message: 
```
 Failure to install package [cis_kubernetes_benchmark]: [ResponseError: validation_exception: [validation_exception] Reason: Validation Failed: 1: no such index [logs-Ido];]
```

So what I think is in addition to adding a section of `index -template` in the Transform config we will also need a `create` boolean flag.

mtojek:
>  Can we make an exception in the Integrations repo CI to allow the CSP packages to merge with this violation for now?

@joshdover To be honest, I'm on the fence, as we used to obey spec rules. We don't have a mechanism to disable spec validation now, but even if we have that would be a backdoor many developers could use :) So rather than disabling the spec, I would recommend merging this PR, as it just legalizes a "transform directory":

```
    - description: Folder containing Elasticsearch Transforms
      # https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html
      type: folder
      name: transform
      required: false
      contents:
        - description: An Elasticsearch transform folder
          type: folder
          pattern: '^[a-z0-9_]+$'
          contents:
            - description: An Elasticsearch transform file
              type: file
              pattern: '^default\.json$'
```

> Transformations as integration-package assets already exist de-facto, and this PR changes the spec to reflect it

^ this, to be honest. We already "committed" a crime, let's just confirm it :)

joshdover:
> > Transformations as integration-package assets already exist de-facto, and this PR changes the spec to reflect it
> 
> ^ this, to be honest. We already "committed" a crime, let's just confirm it :)

This will mean we have to support this for quite a while until we do a v2 of the spec, which is unfortunate since it's only used by 1 package today, one that we have very tight control over. If we add this way of configuring transforms to the spec any package can start using it and we already know this way has problems.

This current way of configuring transforms even poses problems for the very use case we're trying to unblock here and we're already having to hack around it in the CSP Kibana plugin. If we're already having to hack around it, why don't we do this:
1. Don't add this 'broken' way of configuring transforms to the spec
2. Don't add transforms to the CSP package at all until we have fleshed out the long-term support for this in the spec
3. Extend the hack in the CSP Kibana plugin to handle all the transform installation logic until we flesh out the long-term support for transforms in the spec

I just don't see the benefit of adding a broken feature to the spec that we're going to have to keep supporting. We've lived with this 'illegal' feature for some time, why not leave it as-is until we have the time to really fix it rather than committing to supporting it's problems for potentially years when there's not much incentive to do so.

In terms of the long-term approach, how much do we still need to decide? Could we iron out the details in the next week or so?

eyalkraft:
Hey @joshdover, Thanks for your comment. Your proposal sounds reasonable to me and @kfirpeled.
The following steps on my side (will update with links):
1. Remove the transforms from the first version of our @elastic/cloud-security-posture integration package, and merge it. [transform removal commit](https://github.com/elastic/integrations/pull/2930/commits/4674dd6116cc3748e73068617fc6ee5a4655e75c).
2. Update @elastic/cloud-security-posture Kibana plugin to create the required transforms, as first phase.
3. Open an Issue to be handled upon the spec update - "Move transforms creation from the Kibana CSP plugin to the CSP integration package"

sophiec20:
> Is there a reason they can't conform to the data stream naming convention? It's a goal for all data ingested by Elastic integrations to conform to this convention as it provides many benefits across the entire architecture.

Having the transform id follow a `<type>-<dataset>-<namespace>` naming conventions is likely fine (although I can't quite work out the versioning, see later comment about upgrades). Forcing the source and dest indices to have strict naming will limit flexibility of usage, but it sounds like this is not happening according to this comment https://github.com/elastic/package-spec/pull/307#issuecomment-1086844065

> it's only used by 1 package today

@joshdover Transforms in packages is a short term requirement and they will be used in multiple packages soon. It is very difficult to tell what's going on without deep Package knowledge (which not everyone has, especially me). I suggest we get a meeting arranged with more internal stakeholders to figure out the best way forward.

For example, earlier in the thread I see the comment where it is assumed the transform will always be torn down on each package upgrade. We might need to think more about this, because if the package only contains updates to visualizations (say), then the load and flux incurred by retransforming all data should be avoided.

ruflin:
Thanks everyone for the detailed discussion. The issues with transforms have been lingering around for some time and we never fully solved the problem. We rushed an initial version of transforms into Fleet to get the endpoint package out of the doors but it caused us quite a bit of headache and we applied quite a few changes along the way. Lets find a proper solution this time in collaboration with the team from @sophiec20 . Lets not optimise for all possible ways we can think of transforms going to be used but the ones we already know of will come into packages soonish.

If we go down this path, we create a problem for @eyalkraft and team. Even though I hate to have a temporary fix in Kibana as proposed in https://github.com/elastic/package-spec/pull/307#issuecomment-1091716548 I think it is a viable solution to go short term. It will allow to the team to release a first version and later move transforms to the package.  We need to be aware that there will likely be some migration code needed in Kibana at one point.

@eyalkraft I really appreciate that you helped push this conversation forward. My ask to you and the team is to stay proactive on getting this into the package-spec eventually even though you might have a version in Kibana that works. Getting transforms properly into package spec will help many other package devs.

joshdover:
I've sent out a meeting invite for this Thursday with known stakeholders that have a need to ship a transform in their package(s). We'll use this time to quickly get on the same page of what requirements we need to meet and any open questions before we move forward with this spec change. If anyone who is not on the invite would like to be included, please let me know.

eyalkraft:
>  @eyalkraft / @CohenIdo to investigate wildcard workaround mentioned above

I can confirm this will work for us.
This [PR](https://github.com/elastic/kibana/pull/130543) fixes exactly that, but on our first transform (to allow creating the transform before any data was shipped from our beat).

joshdover:
> I can confirm this will work for us.
> This [PR](https://github.com/elastic/kibana/pull/130543) fixes exactly that, but on our first transform (to allow creating the transform before any data was shipped from our beat).

That's great. With that as a workaround, what if we take this course of action:
1. Document this workaround / best practice somewhere
2. Defer on adding the transform dependency feature to the spec that was discussed above
3. Defer on adding namespace-specific transforms in the spec
4. Move forward with moving the index template definition into the same `yml` file as the transform

I believe (3) would be the only required change to this PR and would allow the CSP packages and Endpoint package to pass spec validation. It would require some minor changes to Kibana to implement this and updates to the future Endpoint packages.

Deferring on namespace-specific transforms (2) may be controversial, but I think we can tackle this separately. My understanding is that it is a future requirement, but not an MVP one. There's also a non-trivial amount of effort that needs to happen on the Kibana/Fleet side to support namespace-specific assets of any kind (https://github.com/elastic/kibana/issues/121118), and the Fleet team may not have capacity to take this on in the near-term.

@sophiec20 @eyalkraft please review this and confirm whether or not this fits with your requirements in the near-term.

eyalkraft:
I've updated the spec change suggestion to comply with what we decided.
Under `elasticsearch/transform/example_name/` there would be:
1. A required `transform.json` file with format identical to the [create transform API](https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html#put-transform) (which Kibana effectively supports already). The name of the created transform will be derived from the name of the package and the name of the containing folder (`example_package.example_name-default-0.0.1`) as it is of now (namespace and version are added by fleet).
2. An optional `dest_index_template.json` file with format identical to the [create index template API](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html#create-index-templates), to be installed **before** the in installation of the transform itself. The name of the created index template will be derived from the name of the package and the name of the containing folder (`example_package.example_name`).

Some decisions that are reflected in this suggestion and their reasoning:
1. **Separate files for the transform and the destination index template** - To me it seemed cleaner to have two separate files representing two different objects, instead of having one of them contain the other. I think this could also result with simpler code in Kibana implementing this spec change, as the formats of these files match exactly those of the create APIs. Additionally, I think this design is closer to what I imagine as the longer term design goal, which could be having transforms and index templates as separate assets that could have some relations between them (asset installation order defined in the package?).
2. **Renaming the previous transform file `default.json` to `transform.json`** - This change will require minor changes in the packages that already utilize the existing implementation that supports transforms as package assets (Endpoint security integration package). I wanted to rename the file since following the change I'm suggesting the `elasticsearch/transform/example_name/` directory could contain two files - one for the transform and the other one for the index template. It makes since to name the file explicitly as what it represents.

Let me know what you think @joshdover @sophiec20 @kevinlog @ruflin @mtojek @SourinPaul

test/packages/good/elasticsearch/transform/metadata_united/transform.json
```diff
@@ -0,0 +1,41 @@
+{
+    "source": {
+        "index": [
+            "metrics-endpoint.metadata_current_default*",
+            ".fleet-agents*"
+        ]
+    },
+    "dest": {
+        "index": ".metrics-endpoint.metadata_united_default"
```
jsoriano, line 9: Will referenced indexes be managed by Fleet when installing the package?

test/packages/good/elasticsearch/transform/good_example_abc_1/dest_index_template.json
```diff
@@ -0,0 +1,11 @@
+{
+  "template": {
+    "mappings": {
+      "properties": {
+        "@timestamp": {
```
jsoriano, line 5: Would it be an option to build the mapping from a `fields` directory? Mappings definition would be consistent with other fields definitions in packages, and we could reuse other features we have for static validation.

test/packages/good/elasticsearch/transform/metadata_united/transform.json
```diff
@@ -0,0 +1,41 @@
...
+            }
+        },
+        "group_by": {
+            "agent.id": {
+                "terms": {
+                    "field": "agent.id"
+                }
+            }
+        }
+    },
```
jsoriano, line 36: As a follow up it'd be nice to add in `elastic-package` a way to test transforms.

versions/1/integration/elasticsearch/spec.yml
```diff
@@ -28,3 +28,24 @@ spec:
...
+        - description: An Elasticsearch transform folder
+          type: folder
+          pattern: '^[a-z0-9_]+$'
+          contents:
+            - description: An Elasticsearch transform file
+              type: file
+              required: true
+              contentMediaType: "application/json"
+              pattern: '^transform\.json$'
+            - description: An optional Elasticsearch index template for the tranform's destination index
```
jsoriano, line 46: Nit. 
```suggestion
            - description: An optional Elasticsearch index template for the transform's destination index
```

jsoriano (commented):
Added some questions on the proposed spec.

test/packages/good/elasticsearch/transform/metadata_united/transform.json
jsoriano: We could mention this in the description of the file in the spec.

jsoriano: I have also created https://github.com/elastic/package-spec/issues/354, it may be good to have some reference docs about how the different resources in a package should be interpreted.

ajosh0504:
Hello! Super excited about this work. I did see y'all mention that there's currently no support for chained transforms. For something like Host Risk Score which has a latest transform to get the latest risk score, is there a suggested workaround?

We've tried doing this using aggs in the past but haven't been successful.

test/packages/good/elasticsearch/transform/good_example_abc_1/dest_index_template.json
joshdover: +1 on defining fields the same way as we do for data streams. Also, do we have any use cases where we wouldn't want to the fields defined already? If not, I think we should make it required anyways.

test/packages/good/elasticsearch/transform/metadata_united/transform.json
joshdover: @jsoriano I think your link for "description of the file" may be wrong. Could you clarify where this should be documented for now?

joshdover:
@eyalkraft Some feedback on the description, let's also find a place in the spec itself to include this information, see https://github.com/elastic/package-spec/pull/307#discussion_r901595866.

> This PR allows adding a transform asset to packages, with an optional destination index template to be installed with it. Upon installation of a package that contains a transform asset, the behavior is as follows:
> 
> 1. **If** a destination index template is included:
>    
>    1. Create the destination index template.
>    2. **if** the destination index doesn't already exist:
>       
>       1. Initialize the index, including adding appropriate _meta fields to mark the index as managed by Fleet

I mislead you a bit here, these meta fields should actually be included in the index template itself, so any time an index is created from the template it has the appropriate meta fields.

Let's also add a description of the upgrade behavior, which I believe if we match what we have today would be:
1. Delete index template (but not the index)
2. Stop the transform
3. Delete the transform
4. Create the index template + index
5. Create the transform
6. Start the transform

test/packages/good/elasticsearch/transform/metadata_united/transform.json
jsoriano: Oops, multitasking is bad :sweat_smile: This is the correct link I wanted to use: https://github.com/elastic/package-spec/pull/307/files#diff-a60684d31c8050dde697084c6f224112dad180edd8df8fe820c8deea2747b09eR41

test/packages/good/elasticsearch/transform/metadata_united/transform.json
eyalkraft: opened
- https://github.com/elastic/elastic-package/issues/859

versions/1/integration/elasticsearch/spec.yml
eyalkraft: fixed

eyalkraft:
@jsoriano @joshdover Thanks for your reviews! I've went over your comments and fixed the PR description and the code. Let me know if something is wrong.

@ajosh0504 Thanks for your feedback!

> For something like Host Risk Score which has a latest transform after a pivot to get the latest risk score, is there a suggested workaround?

I successfully used a workaround suggested by @joshdover for a similar issue - try using wildcards in the 2nd transform's source index name.
Please share if this works.

test/packages/good/elasticsearch/transform/good_example_abc_1/dest_index_template.json
jsoriano: After adding the `fields` directory, is this file needed now?

versions/1/integration/elasticsearch/spec.yml
```diff
@@ -28,3 +28,48 @@ spec:
...
+              type: file
+              required: true
+              contentMediaType: "application/json"
+              pattern: '^transform\.json$'
+            - description: Elasticsearch index template for the transform's destination index
+              #  https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html
+              type: file
+              required: false
+              contentMediaType: "application/json"
+              pattern: '^dest_index_template\.json$'
```
jsoriano, line 70: This wouldn't be needed now after adding the `fields` directory?

eyalkraft: I understand how the fields directory replaces the `mappings` part of the index template but I don't understand how the fields directory can replace some other parts of the destination index template, such as `aliases` for example. I'm probably missing something. Can you explain?
For example, for a user to create [this](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-template.html#_index_template_with_index_aliases) index template (from the docs):
```
PUT _index_template/template_1
{
  "index_patterns" : ["te*"],
  "template": {
    "settings" : {
        "number_of_shards" : 1
    },
    "aliases" : {
        "alias1" : {},
        "alias2" : {
            "filter" : {
                "term" : {"user.id" : "kimchy" }
            },
            "routing" : "shard-1"
        },
        "{index}-alias" : {} 
    }
  }
}
```
How would the fields directory look like?

We can decide these parts of the index template are irrelevant for packages and not let them be included in the spec, but otherwise I don't understand how the fields directory is sufficient. 

Packages already can contain data stream assets which implicitly include index templates so I think the gap in my understanding here is how would a user be able to control these extra fields that aren't included in the fields directory (from what I understand).

kevinlog:
@eyalkraft apologies, I missed [this comment](https://github.com/elastic/package-spec/pull/307#issuecomment-1150754249) - I was on PTO for a bit and just getting caught up.

Pulling in @pzl and @joeypoon 

I think we need to make changes in two areas to comply:

> 1. A required `transform.json` file with format identical to the [create transform API](https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html#put-transform) (which Kibana effectively supports already). The name of the created transform will be derived from the name of the package and the name of the containing folder (`example_package.example_name-default-0.0.1`) as it is of now (namespace and version are added by fleet).

we need to rename our files under the following to be `transform.json`. 
- https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform/metadata_current
- https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform/metadata_united

It looks like the naming convention would still apply to our our transforms are named.  For example, `endpoint.metadata_current-default-8.3.0` complies with `example_package.metadata_current-default-version`.  So as far as I can tell, we shouldn't have any issues installing new transforms and referencing them in our code, such as our in app failed transform checks.

> 2. An optional `dest_index_template.json` file with format identical to the [create index template API](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html#create-index-templates), to be installed **before** the installation of the transform itself. The name of the created index template will be derived from the name of the package and the name of the containing folder (`example_package.example_name`).

This is where I'm a little fuzzy.  We do currently have separate index template files here for each of our transforms:
https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/index_template

Based on the spec as written now, it looks like we'd place them next to the `transform.json` files here:
- https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform/metadata_current
- https://github.com/elastic/endpoint-package/tree/master/package/endpoint/elasticsearch/transform/metadata_united

The naming convention will differ from what we currently have will be different than the derived names.

For instance, we have `metrics-endpoint.metadata_current_default` which contains an extra `metrics` qualifier before the package name.

I agree that the new naming convention makes sense, but the first package upgrade which complies with this would require some additional cleanup to delete the other, existing index templates.

@joshdover I'm basing this off of the upgrade behavior in [your comment here](https://github.com/elastic/package-spec/pull/307#issuecomment-1160380258).  Let me know if I'm missing something.

Also, we would need to make some code small Kibana code changes where we reference the old destination index names.

@pzl @joeypoon let me know if the above makes sense.

versions/1/integration/elasticsearch/spec.yml
jsoriano: We would like to limit the features offered to what is needed. If we allow raw index templates, we would need to support all the combination of features available in index templates, we'd prefer to limit this to fields definitions if this is enough. Fields definition seem enough for the sample template included in this PR.

But as mentioned in an offline discussion, and looking to the real templates linked in https://github.com/elastic/package-spec/pull/307#issuecomment-1163003994, apart of the fields mappings definitions, we would also need to support some additional settings:
```
        "settings": {
            "index": {
                "codec": "best_compression",
                "refresh_interval": "5s",
                "number_of_shards": "1",
                "number_of_routing_shards": "30",
                "hidden": true
            }
        },
``` 
```
        "settings": {
            "index": {
                "codec": "best_compression",
                "refresh_interval": "5s",
                "number_of_shards": "1",
                "number_of_routing_shards": "30",
                "sort.field": [
                    "@timestamp",
                    "agent.id"
                ],
                "sort.order": [
                    "desc",
                    "asc"
                ]
            }
        },
```
This kind of settings are supported in data stream index templates, by using the [`elasticsearch.index_template.settings`](https://github.com/elastic/package-spec/blob/10b32dff0212c3c09970171f7702a77d4c3fd55c/versions/1/integration/data_stream/manifest.spec.yml#L198) setting in the data stream manifest.

Maybe an option is to add a manifest in the transform directory to include these settings, in a way that would be consistent with data streams.

@joshdover @mtojek what do you think is the best way to support these settings?

eyalkraft: @jsoriano I've updated the PR with your suggestion, and updated the examples based on the actual transforms and index templates used by the endpoint package

versions/1/integration/elasticsearch/transform/manifest.spec.yml
```diff
@@ -0,0 +1,14 @@
...
+## Describes the specification for a transform's manifest.yml file
+##
+spec:
+  type: object
+  additionalProperties: false
+  definitions:
+    index_template:
+      $ref: "../../data_stream/manifest.spec.yml#/properties/elasticsearch/properties/index_template"
+  properties:
+    dest_index_template:
```
jsoriano, line 11: Nit. No need to use abbreviations.
```suggestion
    destination_index_template:
```

test/packages/good/elasticsearch/transform/metadata_current/transform.json
```diff
@@ -0,0 +1,32 @@
...
+        "managed": true
+    },
+    "frequency": "1s",
+    "sync": {
+        "time": {
+            "field": "event.ingested",
+            "delay": "1s"
+        }
+    }
+}
```
jsoriano, line 32: Nit. We support defining pipelines as YAML, that are converted to JSON on install time. This makes them easier to maintain.
Would it make sense to do the same here?

eyalkraft: Yeah I think it could work here as well

eyalkraft: changed to yml.

joshdover:
> I agree that the new naming convention makes sense, but the first package upgrade which complies with this would require some additional cleanup to delete the other, existing index templates.
> 
> @joshdover I'm basing this off of the upgrade behavior in [your comment here](https://github.com/elastic/package-spec/pull/307#issuecomment-1160380258). Let me know if I'm missing something.

@kevinlog The uninstall logic in Fleet is very basic right now. We keep track of references of all assets we create during install, regardless of which part of the install process created the asset, and then just delete those assets.

So I think since we will have a reference to the index template we created in the legacy package, when we do the upgrade, the old template should get removed regardless. Note that we don't delete indices, would that be required?

versions/1/integration/elasticsearch/transform/manifest.spec.yml
```diff
@@ -0,0 +1,14 @@
+##
+## Describes the specification for a transform's manifest.yml file
+##
+spec:
+  type: object
+  additionalProperties: false
+  definitions:
+    index_template:
+      $ref: "../../data_stream/manifest.spec.yml#/properties/elasticsearch/properties/index_template"
+  properties:
```
joshdover, line 10: Didn't we have a requirement to have the option to disable starting a transform by default? I'd be ok doing it in an immediate follow up PR to just get this one in first.

versions/1/integration/elasticsearch/spec.yml
```diff
@@ -28,3 +28,49 @@ spec:
...
+            3. Delete the transform.
+            4. From here, follow the creation behavior.
+          type: folder
+          pattern: '^[a-z0-9_]+$'
+          contents:
+            - description: Elasticsearch transform file
+              type: file
+              required: true
+              contentMediaType: "application/x-yaml"
+              pattern: '^transform\.yml$'
```
joshdover, line 64: Should we have validation on the content of this file? For instance, I'd like there to be constraints on the source / destination index names to match the data stream naming scheme.

joshdover (commented):
Looking very close! A couple last suggestions

versions/1/integration/elasticsearch/spec.yml
sophiec20: I am worried that validation on source/dest index names will make this less applicable to use of transforms in other packages. Security team are not using datastream naming scheme in their host risk score package. Can we avoid strict naming scheme checking?

kevinlog:
@joshdover 

> @kevinlog The uninstall logic in Fleet is very basic right now. We keep track of references of all assets we create during install, regardless of which part of the install process created the asset, and then just delete those assets.

> So I think since we will have a reference to the index template we created in the legacy package, when we do the upgrade, the old template should get removed regardless. Note that we don't delete indices, would that be required?

If we conform to the proposed index template naming, it seems it would imply that we should rename the destination indices for our transforms.

For instance: https://github.com/elastic/endpoint-package/blob/master/package/endpoint/elasticsearch/index_template/metrics-metadata-current.json

Our index is coded to be:  `metrics-endpoint.metadata_current_default` which is different than the proposed index template naming convention `package-name.index-name`, note the `metrics` qualifier.  It's OK to create new indices and index our data into those new indices from scratch, however we should clean up the old indices.  Maybe we have a one-time cleanup to account for this which should be a special case?

Let me know if the above makes sense.  I may be missing something.

cc @eyalkraft @pzl @joeypoon

versions/1/integration/elasticsearch/transform/manifest.spec.yml
eyalkraft: Updated the spec to include `start`, added an example, updated the description on the spec & issue.

eyalkraft:
@kevinlog 

> If we conform to the proposed index template naming, it seems it would imply that we should rename the destination indices for our transforms.

I don't think that would be necessary.

> The name of the created index template will be derived from the name of the package and the name of the containing folder, For example: `example_package.example_name`.

Note that this talks about the name of the index template itself, and not about the index pattern(s).
Your index template could have the same index pattern as of now.

See for example one of the index templates we use:
Screen Shot 2022-06-26 at 16 35 40
It doesn't have `logs` in the name, but only in the index pattern.

Your index template 
Screen Shot 2022-06-26 at 16 38 05
Could be renamed to `endpoint.metadata_current` and keep the index pattern as is.

---
I'm updating the spec and issue descriptions with the naming information as it was a bit hidden in one of the previous comments.

versions/1/integration/elasticsearch/transform/transform.spec.yml
```diff
@@ -0,0 +1,226 @@
...
+      type: object
+      required:
+      - index
+      additionalProperties: false
+      properties:
+        index:
+          "$id": "#root/dest/index"
+          title: Index
+          type: string
+          pattern: "^.*$" # TODO: Enforce the data stream naming scheme.
```
eyalkraft, line 108: Destination index naming constraints

versions/1/integration/elasticsearch/transform/transform.spec.yml
```diff
@@ -0,0 +1,226 @@
...
+          "$id": "#root/source/index"
+          title: Index
+          oneOf:
+          - type: array
+            items:
+              type: string
+          - type: string
+          examples:
+          - kibana_sample_data_ecommerce
+          pattern: "^.*$" # TODO: Enforce the data stream naming scheme.
```
eyalkraft, line 40: Source index naming constraints

eyalkraft (commented):
TODO: Index naming constrains.

Could be done on the following iteration.

versions/1/integration/elasticsearch/spec.yml
eyalkraft: I've added validations on the transform based on the [schema from Elasticsearch's repo](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/transform/src/test/resources/rest-api-spec/schema/transform_config.schema.json).

Currently no constraints on the source / destination index names as following @sophiec20's comment I thought we might want and wait with that for the next iteration.

The PR includes the infrastructure for including such constraints on the next iteration:
[destination](https://github.com/elastic/package-spec/pull/307#discussion_r906832249), [source](https://github.com/elastic/package-spec/pull/307#discussion_r906832321).

kevinlog:
@eyalkraft 

> Note that this talks about the name of the index template itself, and not about the index pattern(s).
Your index template could have the same index pattern as of now.

You are correct here, thanks for showing your team's example as well!

@joshdover my prior concerns with upgrade and deleting old indices shouldn't be an issue.  We can change the name of our index pattern and/or move the template files in the directory, but we can leave the index patterns the same.

cc @pzl @joeypoon

joshdover:
> Note that this talks about the name of the index template itself, and not about the index pattern(s). Your index template could have the same index pattern as of now.
> 
> See for example one of the index templates we use: Screen Shot 2022-06-26 at 16 35 40 It doesn't have `logs` in the name, but only in the index pattern.

@eyalkraft Is there a reason not to include the type of the data in the name of the index template? I think it should match the existing pattern we have for all other index templates which does include the type prefix, for example the `logs-nginx.access` template matches `logs-nginx.access-*`.

You are right that the name of the index template shouldn't impact the name of the destination indices, but IMO, they should actually match so this isn't confusing or inconsistent.

eyalkraft:
@joshdover 
> @eyalkraft Is there a reason not to include the type of the data in the name of the index template? I think it should match the existing pattern we have for all other index templates which does include the type prefix, for example the `logs-nginx.access` template matches `logs-nginx.access-*`.

Actually there is no reason I know of we don't include the type of the data in the name of the index template.
I was under the impression we went through some verifications over the names we use but I guess I was wrong.
We will rename our index templates to match the suggested pattern.

I'm updating the PR and the description to match then.

cc @kevinlog sorry for the confusion

eyalkraft:
Following another quick sync  with @joshdover regarding the naming of the transform, we thought it makes sense for the transform to be named in a similar way to the index template (aka include the data type prefix).
This means both CSP and endpoint's transforms would be renamed.
image
So for example
`endpoint.metadata_current-default-8.3.0` -> `metrics-endpoint.metadata_current-default-8.3.0`.

Same with the index template names, the name of the transform doesn't affect the destination/source indices it runs on, but optimally is correlated with them to provide a clear user experience.

@kevinlog

versions/1/integration/elasticsearch/spec.yml
joshdover: Discussed with @sophiec20 on Slack. It seems we have some migration scenarios of existing transforms to packages, and this restriction won't be feasible for such cases. Let's move this to a recommendation in the spec's description and remove the TODOs for now. Thanks @eyalkraft

eyalkraft: Done

joshdover (approved):
LGTM thank you for locking down the transform fields.

jsoriano (approved):
:rocket:

joshdover:
Implementation for supporting this in Fleet's installation code will be discussed in https://github.com/elastic/kibana/issues/134321. This is not yet scheduled.

joshdover:
I've opened a follow up issue for the future improvements that still need to be made, many of which were discussed in this PR: https://github.com/elastic/package-spec/issues/370

szeitlin:
> I've opened a follow up issue for the future improvements that still need to be made, many of which were discussed in this PR: #370

Great! could you link to that follow-up issue please?

PR #309, opened 2022-03-30
Validate that fields are defined only once per data stream

jsoriano:
## What does this PR do?

Validates that fields are defined only once per data stream.

## Why is it important?

Fields defined multiple times can lead to undefined or unexpected results, for example in https://github.com/elastic/integrations/pull/2916#pullrequestreview-926504950 it was found that a field changed unexpectedly, and it was defined twice.

## Checklist

## Related issues

-


test/packages/bad_duplicated_fields/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.0.1"
+  changes:
+    - description: Initial draft of the package
+      type: enhancement
+      link: https://github.com/elastic/integrations/pull/0 # FIXME Replace with the real PR link
```
mtojek, line 6: nit: you can link it to this PR.

mtojek (approved):

Issue #282, opened 2022-02-21
[Change Proposal] Stricter validation on fields properties


jsoriano:
We have found that it is pretty common (https://github.com/elastic/integrations/issues/2643) to have typos in fields properties. This can be an important issue because this may create unexpected field mappings, and fixing mappings uses to bring breaking changes.

In https://github.com/elastic/package-spec/pull/281 I explore the enforcement of stricter validation, but there I found that at least for the `example` field we would need some trade offs.

andrewkroh:
There is a misalignment between the spec and what Fleet accepts for field definitions. I think these two should be sync'ed while enabling the stricter validation. 

https://github.com/elastic/kibana/blob/b4b3285643dd9105e4c34883e8c032e771ee5fcb/x-pack/plugins/fleet/server/services/epm/fields/field.ts#L14

endorama:
This bit us multiple times, would be great to have it! It also makes onboarding quite difficult: knowing and navigating the package spec is complex and any error slips through validation. 

Would be very useful for package devs to have this in the form of warnings (for example as part of package linting). Having a soft check that does not break the integration may also be helpful for adapting integrations before forcing this in the CI (as I expect multiple integrations would not pass this check at first).

jsoriano:
@endorama stricter validation is coming in package-spec v2 :) https://github.com/elastic/package-spec/pull/420
This will make CI fail though, but package developers can choose when to change the `format_version` to take advantage of these new validations. You will have more news about this soon.

Regarding docs, yes, this is another pending thing we have. I was thinking on generate more easily consumable docs from the spec (https://github.com/elastic/package-spec/issues/400), I hope to work on this soonish.

PR #314, opened 2022-04-05
Add missing properties to fields.yml spec

andrewkroh:
## What does this PR do?

I looked over the [code](https://github.com/elastic/kibana/blob/0695df649705c91ed28dbc73037aea878ce419f7/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L105) in Fleet EPM to see what properties are used when generating templates and then updated the specification to match.

~The current version is a draft because it contains commentary as "TODO"s.~

## Why is it important?

A detailed and accurate specification is required to ensure the developers of integration packages and the implementors of the specification agree on the behavior.

## Checklist

## Related issues

- Relates #282

## Author Notes

Based on feedback in this PR we need to follow up on these tasks.

- [x] Answer why Fleet has include_in_root, include_in_parent. _These are for nested type fields and are now in the spec._
- [x] Answer why Fleet has "group-nested" and "nested". _It's internal to Fleet. No integrations will use it._
- [ ] `null_value` should be supported by types other than `wildcard`.
- [ ] Add support from `dynamic: runtime`.
- [ ] Update schema to support [dynamic mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html) (e.g. `path_match`, `path_unmatch`, `match`, `unmatch`, etc). This blocks development of https://github.com/elastic/kibana/issues/129344. We'll want to address this before incorporating this into `elastic-package` so that there is a migration path for fields that "think" they are creating dynamic mappings today and are using unsupported properties.
- [ ] Fleet should support `multi_field` on `match_only_text` field types.
- [ ] Fleet should add support for `norms` on `text` fields.
- [ ] Need to determine how deal with what `format` does in Beats (e.g. applying a formatter to the data view).


versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -6,66 +6,82 @@ spec:
...
+        description: >
+          Name of field. Names containing dots are automatically split into
+          sub-fields.
         type: string
+        # TODO: Should we allow wildcards if they are not implemented in Fleet?
+        pattern: '^[_@A-Za-z0-9]+(\.[_@A-Za-z0-9]+)*$'
       type:
         description: Datatype of field
         type: string
         enum:
```
andrewkroh, line 23: No changes to the contents of this list were made. I just sorted it to make identifying missing types a little easier.

This is not a complete list, but I noticed `geo_shape` and `dense_vector` are supported in ES but not in packages.

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -6,66 +6,82 @@ spec:
...
-        - ip_range
-        - group
+        - double
+        - double_range
+        - flattened
+        - float
+        - float_range
         - geo_point
-        - object
+        - group
```
andrewkroh, line 39: Fleet EPM has a `group-nested` type. Does that need added to the list? Why would we have both `group-nested` and `nested`?

https://github.com/elastic/kibana/blob/0695df649705c91ed28dbc73037aea878ce419f7/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L120

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -6,66 +6,82 @@ spec:
...
-        description: Name of field
-        type: string
-      title:
-        description: Title of field
+        description: >
+          Name of field. Names containing dots are automatically split into
+          sub-fields.
         type: string
+        # TODO: Should we allow wildcards if they are not implemented in Fleet?
+        pattern: '^[_@A-Za-z0-9]+(\.[_@A-Za-z0-9]+)*$'
```
jsoriano, line 19: Yes, I think we need to allow wildcards.

The reason we don't see more wildcards now is because we are interpreting objects without wildcards as if the wildcard would be at the end by convention. For example for `labels`, the matching path pattern would be `labels.*`. See how the wildcard is added also in beats [here](https://github.com/elastic/beats/blob/ea207346d651448b8917b0791b2b117b9f9b9212/libbeat/template/processor.go#L471-L473).
There are cases where the `*` needs to be in the middle of a path, like in the [prometheus cases](https://github.com/elastic/integrations/blob/b0f94db6310604548ca98c322fa4dd5489106a0a/packages/prometheus/data_stream/collector/fields/fields.yml#L15), and they cannot benefit of the previous convention.

An alternative would be to have an explicit field to define the [`path_match`](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/dynamic-templates.html#path-match-unmatch), but this could be error prone (typos between the name of the field and its matching path), would need changes in fields validation, and it would break packages that currently use the wildcard.

versions/1/data_stream/fields/fields.spec.yml
jsoriano: If these fields are not used in current packages and is not supported by Fleet, I would say to add them later when needed.

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -6,66 +6,82 @@ spec:
...
-        - flattened
-        - wildcard
-        - version
+        - object
+        - scaled_float
+        - short
+        - text
         - unsigned_long
+        - version
+        - wildcard
```
jsoriano, line 57: I think there was some kind of "order" in this enum (numbers together, strings together...). I am ok though with using alphabetical order, but maybe we can leave the reordering of this list for a different PR, so it is clearer here if any type is being added or removed.

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -77,22 +93,178 @@ spec:
...
+        type: string
+
+      multi_fields:
+        description: >
+          It is often useful to index the same field in different ways for
+          different purposes. This is the purpose of multi-fields. For instance,
+          a string field could be mapped as a text field for full-text search,
+          and as a keyword field for sorting or aggregations.
+
+          Only honored for 'type: text/keyword/wildcard'.
```
jsoriano, line 207: After https://github.com/elastic/kibana/pull/128391 `match_only_text` will be also supported.
```suggestion
          Only honored for 'type: text/keyword/wildcard/match_only_text'.
```

Or do you mean that ES only supports multi fields on these types? Then maybe we have to reconsider the latest changes about this in the spec and fleet.

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -77,22 +93,178 @@ spec:
...
+        $ref: "#" # JSON-schema syntax for pointing to the root of the schema
+
+      # TODO: Fleet only honors this for 'type: wildcard'. That seems wrong.
```
jsoriano, line 210: Ping @joshdover

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -77,22 +93,178 @@ spec:
...
+      normalizer:
+        description: >
+          Specifies the name of a normalizer to apply to keyword fields. A
+          simple normalizer called lowercase ships with elasticsearch and can be
+          used. Custom normalizers can be defined as part of analysis index
+          settings.
+        type: string
+
+      # TODO: This two are read by Kibana, but AFAICT they have no use in
+      # Elasticsearch. And nothing in elastic/integrations uses it.
```
jsoriano, line 259: Then maybe we shouldn't add them by now?

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -6,66 +6,82 @@ spec:
   type: array
   items:
     type: object
+    # TODO: We should disable additionalProperties. There are properties used in
+    # elastic/integrations that have no purpose or are typos.
     additionalProperties: true
```
jsoriano, line 11: +1, specially after adding these missing properties. @mtojek wdyt?
```suggestion
    # TODO: We should disable additionalProperties. There are properties used in
    # elastic/integrations that have no purpose or are typos.
    additionalProperties: false
```

jsoriano (commented):
Thanks @andrewkroh! Awesome revamp of the fields spec.

versions/1/data_stream/fields/fields.spec.yml
mtojek: +1 on my side too.

I think that we left it open to allow for extra properties like examples, patterns, etc. It seems that we aren't using the actively, so let's close this option.

versions/1/data_stream/fields/fields.spec.yml
mtojek: I would be precautious about them too unless we're sure that Fleet fully supports them. Is there a requirement now to allow for and use all these properties?

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: This is documenting Fleet. For clarity I'll change the wording to "Fleet honors this for `keyword`, `match_only_text', `text`, and `wildcard` types."

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: I'll remove them and make a note to follow up to determine why these exist in Fleet. From there we can determine the next steps - either add them here or remove them from Fleet.

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: We can consider sorting and do that in another PR. I have reverted that change.

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: I amended the pattern to allow `*` and added a test case with a `foo.*`. I'll create a follow-up issue to continue discussion of the design for dynamic mappings.

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: I set `additionalProperties: false` and adjusted some of the test packages to remove level, group, and title.

jsoriano: There are many fields using level, group and/or title in the integrations repo, wdyt about adding these fields here by now? We could add them directly as deprecated, and eventually phase them out.

I have created this PR to check what would fail in the integrations repo after this change: https://github.com/elastic/integrations/pull/3024

joshdover:
@juliaElastic Do you have time to review this? I believe you're more up to date on our mapping field logic than I am.

juliaElastic:
> @juliaElastic Do you have time to review this? I believe you're more up to date on our mapping field logic than I am.

I can have a look later today.

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -52,20 +52,33 @@ spec:
...
+
       value:
         description: The value to associate with a constant_keyword field.
         type: string
+
       metric_type:
-        description: Metric type
+        description: >
+          The metric type of a numeric field. This is attached to the
+          field as metadata (via `meta`). A gauge is a single-value measurement
```
juliaElastic, line 67: `metric_type` is added as a `time_series_metric` mapping in a recent [change](https://github.com/elastic/kibana/pull/126322), should this be mentioned here?

versions/1/data_stream/fields/fields.spec.yml
juliaElastic: it was added 2 years ago in this pr, but doesn't say a lot of context about why: https://github.com/elastic/kibana/pull/64829

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: Thanks for letting me know. I've update the docs for this. I also updated the `dimension` parameter to mention that it uses `time_series_dimension`.

versions/1/data_stream/fields/fields.spec.yml
```diff
@@ -77,22 +90,167 @@ spec:
...
+
+      multi_fields:
+        description: >
+          It is often useful to index the same field in different ways for
+          different purposes. This is the purpose of multi-fields. For instance,
+          a string field could be mapped as a text field for full-text search,
+          and as a keyword field for sorting or aggregations.
+
+          Fleet honors this for `keyword`, `match_only_text`, `text`, and
+          `wildcard` types.
```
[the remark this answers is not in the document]
andrewkroh: Good catch. I'll remove `match_only_text`. (Fleet should support multi fields on `match_only_text` so I will make a note of that and create an issue.)

versions/1/data_stream/fields/fields.spec.yml
andrewkroh: I didn't try to follow the logic, but my take is that is seems like an internal-only value. So none of the integrations should use `group-nested` and it does not need to be in the spec.

juliaElastic (dismissed):
LGTM

jsoriano:
I have updated https://github.com/elastic/integrations/pull/3024, and there are still some missing fields (`level`, `group`, `title`, `example`...). It may be hard to support all fields on a first pass.

@andrewkroh @mtojek wdyt about setting `additionalProperties` back to `true` so we can merge all the fields and docs included here, and continue adding fields in future PRs?

test/packages/missing_required_fields/data_stream/foo/fields/some_fields.yml
```diff
@@ -1,24 +1,18 @@
...
-      level: core
       type: geo_point
       description: Longitude and latitude.
     - name: geo.region_iso_code
-      level: core
       type: keyword
       description: Region ISO code.
       ignore_above: 1024
     - name: geo.region_name
-      level: core
```
jsoriano, line 21: @andrewkroh can you also revert the removal of these fields in test packages? So we keep these as they are still used by many integrations. Thanks!

test/packages/good/data_stream/foo/fields/some_fields.yml
```diff
@@ -35,3 +35,15 @@
 - name: error.message
   description: Error message.
   type: match_only_text
+- name: metric.*_bytes
```
jsoriano, line 38: Nit. Is this an intentional test case? We don't use to have wildcards mixed with suffixes, not sure if everything works as expected with this.
```suggestion
- name: metric.*.bytes
```

jsoriano (commented):
LGTM, I have updated https://github.com/elastic/integrations/pull/3024 as a last check in case there is something unexpected.

test/packages/good/data_stream/foo/fields/some_fields.yml
andrewkroh: This was intentional to test the `pattern` specified for `name`. It really is an area of unspecified behavior at the moment. I was operating under the assumption that wildcards will be used to implement behavior similar to [`path_match`](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-templates.html#path-match-unmatch). But for now this is a test case just to check that the pattern matches fields like this:

https://github.com/elastic/integrations/blob/3c88555054fd24a461babebe624d8645b5feea6c/packages/aws/data_stream/cloudwatch_metrics/fields/package-fields.yml#L16

We should refine this area of the spec for dynamic fields.

andrewkroh: I could remove the `pattern` for now. Then when we get to dynamic fields add it back to match the definition there.

jsoriano: Oh ok, lets leave it. Thanks.

jsoriano (approved):
Pending failures in https://github.com/elastic/integrations/pull/3024 don't seem to be related to this, but to #309.

andrewkroh:
I've been looking over the issues in the integrations repo. After fixing many unused fields and typos (see https://github.com/elastic/integrations/pull/3239/commits) this is what's left:

- 763 instances of `format` - used to specify things like `byte`, `percent`
- 1162 instances of `norms` - used to specify https://www.elastic.co/guide/en/elasticsearch/reference/current/norms.html
- 164 instances of `object_type_mapping_type` - part of dynamic fields behavior that is TBD
- 61 issues with field `name`s that don't match `pattern`. We need to include `-` (dash) in the list of accepted characters.

andrewkroh:
After allowing a dash there are three issues with names, and they all seems like problems with the integration:

`DBClusterIdentifier,Role` - https://github.com/elastic/integrations/blob/4247dfe2b3fe14d5e3dd02eb2e9891814e9f92e0/packages/aws/data_stream/rds/fields/fields.yml#L13

`DbClusterIdentifier, EngineName` - https://github.com/elastic/integrations/blob/4247dfe2b3fe14d5e3dd02eb2e9891814e9f92e0/packages/aws/data_stream/rds/fields/fields.yml#L16

`NumberOfNotificationsFilteredOut-InvalidAttributes.sum` - https://github.com/elastic/integrations/blob/8770467808f03a38de9563d7052a22ea1e472eea/packages/aws/data_stream/sns/fields/fields.yml#L49

So I think this is good now.

jsoriano:
> `DBClusterIdentifier,Role` - https://github.com/elastic/integrations/blob/4247dfe2b3fe14d5e3dd02eb2e9891814e9f92e0/packages/aws/data_stream/rds/fields/fields.yml#L13
> 
> `DbClusterIdentifier, EngineName` - https://github.com/elastic/integrations/blob/4247dfe2b3fe14d5e3dd02eb2e9891814e9f92e0/packages/aws/data_stream/rds/fields/fields.yml#L16
> 
> `NumberOfNotificationsFilteredOut-InvalidAttributes.sum` - https://github.com/elastic/integrations/blob/8770467808f03a38de9563d7052a22ea1e472eea/packages/aws/data_stream/sns/fields/fields.yml#L49

@kaiyan-sheng do you expect any problem if we fix these fields to don't contain unexpected characters?

kaiyan-sheng:
> DBClusterIdentifier,Role

@jsoriano Hmmm good catch, I don't think we are actually storing dimensions like this. For example, this dimension is stored as `aws.dimensions.DBClusterIdentifier` and `aws.dimensions.Role` as two separate fields. These fields are documented wrong, I will fix them.

andrewkroh:
@kaiyan-sheng Thanks for addressing RDS in https://github.com/elastic/integrations/pull/3253. Is the SNS dimension a bug or are we expecting to have JSON key names containing commas?

kaiyan-sheng:
> Is the SNS dimension a bug or are we expecting to have JSON key names containing commas?

@andrewkroh Good catch! SNS dimension is also a bug and I just added the fix into the same PR. Thank you!

andrewkroh:
I think this is ready now. I will start creating follow-up issues for the things listed in the PR description.

jsoriano (approved):
:+1: thanks Andrew! @mtojek @juliaElastic any objection to merge this?

jsoriano:
Oh, one thing would be missing, @andrewkroh could you please add a changelog entry?

mtojek (approved):

jsoriano:
Ah, changelog is already here :facepalm: Merging.

PR #321, opened 2022-04-19
[Osquery] Add Kibana/osquery-saved-query

tomsonpl:
## What does this PR do?

Adds support for `kibana/osquery-saved-query`

## Why is it important?

We want to be able to ship `osquery_saved_query` with `osquery_manager` integration, so the user can start using Osquery with the set of prebuilt saved queries.

## Checklist

## Related issues

[Relates https://github.com/elastic/integrations/pull/2851](https://github.com/elastic/integrations/pull/2998)
[Relates https://github.com/elastic/kibana/pull/128109](https://github.com/elastic/kibana/pull/129461)

### DIFF
--- a/versions/1/changelog.yml
+++ b/versions/1/changelog.yml
@@ -2,16 +2,14 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom.
 ##
-- version: 1.8.0
-  changes:
-  - description: Add Kibana/osquery-saved-query
-    type: enhancement
-    link: https://github.com/elastic/package-spec/pull/321
 - version: 1.7.1-next
   changes:
   - description: Validate that fields are only defined once per data stream.
     type: enhancement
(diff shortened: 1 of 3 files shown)


versions/1/changelog.yml
```diff
@@ -2,6 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom.
 ##
+- version: 1.8.0
+  changes:
+  - description: Add Kibana/osquery-saved-query
```
mtojek, line 7: Let's add this entry to the 1.7.1-next. Eventually, it will become 1.8.0.

mtojek:
ping @jen-huang as Josh is OOTO.

versions/1/changelog.yml
tomsonpl: Ok!

jen-huang (approved):
New Kibana asset type LGTM

mtojek (approved):

Issue #319, opened 2022-04-14
Support “input” package type


mtojek:
_This is the meta issue for all actions required to introduce the new package type._

Background:

> The concept of the Elastic Package is GA for a while and helped us migrate Beats modules to Elastic Integrations and create new integrations from scratch. So far, we identified only one package type - the Elastic Integration, which meets most of the technical requirements for data streams, fields, and agent policies.
It has been proved that the format of the Elastic Integration, defined by the Package Spec, is flexible enough to wrap Beats inputs as separate integrations (for example Custom Logs or HTTP JSON), but not so flexible to implement custom Ingest configurations. On the other hand, the Integration type seems to be an overkill for the Input-only package type, as it doesn’t define multiple data streams or come with a specific ingest pipeline.

Design doc: [link](https://docs.google.com/document/d/1vzBh2frnlxEBBcab8-ist8pu_PDZUjOHdsV-QuUn_WA/edit#)

Let's review the doc, comment, and create subtasks.

Dependency:
- [ ] https://github.com/elastic/kibana/issues/133296

mtojek:
As agreed with @jlind23, I'm moving on with this topic:

- [ ] Distinguish two different package types in spec.
   - Implementation challenge: use two different directories
- [ ] Validate the new type of package
   - Add positive and negative tests.

mtojek:
Status update on input packages:

https://github.com/elastic/elastic-package/pull/840 - is the last PR enabling support for input package in elastic-package. Post-MVP: what is left is figuring out and adapting system tests logic, but to do this we need API support on the Fleet side.

To close this chapter of works we have to release a new version of package-spec, elastic-package, and include them in Integrations.

PR #323, opened 2022-04-25
Prepare for support of multiple, independent package types

mtojek:
Meta-issue: https://github.com/elastic/package-spec/issues/319

This PR modifies the package-spec logic to support multiple package types under `versions/1/`. I think that it's more like a technical PR to open implementation for extending with more package types.

Fun fact: it's really concerning that we haven't required `type: integration` to be present.


spec.go
```diff
@@ -9,7 +9,7 @@ import (
 	"io/fs"
 )
 
-//go:embed versions/1 versions/1/_dev versions/1/data_stream/_dev
+//go:embed versions/1 versions/1/integration/_dev versions/1/integration/data_stream/_dev
```
mtojek, line 12: @jsoriano Does Go ignore `_underscore-dirs` by design? I must have missed something here.

jsoriano: Yes, there are certain parts of Go that ignore dirs prefixed with underscores and the `testdata` directory. For example I think that `go get` doesn't get these files by default, and `go test ./...` probably doesn't look for tests there.
I think that files referenced by `embed` are also seen by go get and other commands that would ignore them otherwise.

mtojek: Ok, I guess it's a workaround then. The PR is ready for a review round.

jsoriano (approved):
:+1:

PR #324, opened 2022-04-26
Add validation to check top changelog link

r00tu53r:
## What does this PR do?

On occasion the PR link in the changelog is not updated with the correct number. It is left with `xxxx` or `0` or with no number. The package could get merged to `main` with no check to catch it. This PR adds a validation to check all the changelog entry's links. 

Since the link could technically be any issue tracking system, this change enforces that the base of the path (which is the number) be a valid number (not a `0`) for a github.com URL.

Initial discussion about this is here - https://github.com/elastic/elastic-package/pull/793

## Why is it important?

As explained in the section above.

## Checklist

- [x] I have added unit tests to test the validation.

### DIFF
--- a/code/go/internal/validator/semantic/version_integrity.go
+++ b/code/go/internal/validator/semantic/version_integrity.go
@@ -18,6 +18,24 @@ import (
 	"github.com/elastic/package-spec/code/go/internal/pkgpath"
 )
 
+var errGHIssue = errors.New("issue number in changelog link should be a positive number")
+
+// ChangelogLinkError records the link and the error
+type ChangelogLinkError struct {
+	link string
+	err  error
+}
+
+// Is checks if the target matches one of the allowed links errors.
+// Currently checks for github issue/pr links.
+func (e ChangelogLinkError) Is(target error) bool {
+	return target == errGHIssue
+}
+
+func (e ChangelogLinkError) Error() string {
+	return fmt.Sprintf("%s: %s", e.link, e.err.Error())
+}
+
 // ValidateVersionIntegrity returns validation errors if the version defined in manifest isn't referenced in the latest
 // entry of the changelog file.
 func ValidateVersionIntegrity(fsys fspath.FS) ve.ValidationErrors {
@@ -85,7 +103,6 @@ func readChangelogVersions(fsys fspath.FS) ([]string, error) {
 }
 
 func readChangelog(fsys fspath.FS, jsonpath string) ([]string, error) {
-
 	changelogPath := "changelog.yml"
 	f, err := pkgpath.Files(fsys, changelogPath)
 	if err != nil {
@@ -151,9 +168,7 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
 }
 
 func ensureLinksAreValid(links []string) ve.ValidationErrors {
-
 	type validateFn func(link *url.URL) error
-
 	var errs ve.ValidationErrors
 
 	validateLinks := []struct {
@@ -162,7 +177,7 @@ func ensureLinksAreValid(links []string) ve.ValidationErrors {
 	}{
 		{
 			"github.com",
-			validateGithub,
+			validateGithubLink,
 		},
 	}
 	for _, link := range links {
@@ -184,10 +199,13 @@ func ensureLinksAreValid(links []string) ve.ValidationErrors {
 	return errs
 }
 
-func validateGithub(ghLink *url.URL) error {
+func validateGithubLink(ghLink *url.URL) error {
 	prNum, err := strconv.Atoi(path.Base(ghLink.Path))
 	if err != nil || prNum <= 0 {
-		return fmt.Errorf("issue number in changelog link %v should be a positive number", ghLink)
+		return &ChangelogLinkError{
+			ghLink.String(),
+			errGHIssue,
+		}
 	}
 	return nil
 }
--- a/code/go/internal/validator/semantic/version_integrity.go
+++ b/code/go/internal/validator/semantic/version_integrity.go
@@ -186,10 +186,7 @@ func ensureLinksAreValid(links []string) ve.ValidationErrors {
 
 func validateGithub(ghLink *url.URL) error {
 	prNum, err := strconv.Atoi(path.Base(ghLink.Path))
-	if err != nil {
-		return err
-	}
-	if prNum <= 0 {
+	if err != nil || prNum <= 0 {
 		return fmt.Errorf("issue number in changelog link %v should be a positive number", ghLink)
 	}
 	return nil
--- a/code/go/internal/validator/semantic/version_integrity.go
+++ b/code/go/internal/validator/semantic/version_integrity.go
@@ -165,10 +165,12 @@ func ensureLinksAreValid(links []string) ve.ValidationErrors {
 			validateGithub,
 		},
 	}
-	for _, link := range links {
+	for i, link := range links {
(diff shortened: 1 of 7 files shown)


versions/1/changelog.yml
```diff
@@ -2,6 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom.
 ##
+- version: 1.7.2-next
```
mtojek, line 5: Let's "park" it in 1.7.1-next. I suppose that all these will hit 1.8.

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -40,6 +43,20 @@ func ValidateVersionIntegrity(fsys fspath.FS) ve.ValidationErrors {
 	return nil
 }
 
+// ValidateTopChangelogLink returns validation errors if the link is not a valid PR github.com link.
+// If the link is not a github.com link this validation is skipped and does not return an error.
+func ValidateTopChangelogLink(fsys fspath.FS) ve.ValidationErrors {
```
mtojek, line 48: Is there any reason why you validate only the top link? I think it's feasible to go over all entries.

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -127,3 +154,43 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
...
+	validateLinks := []struct {
+		domain       string
+		validateLink validateFn
+	}{
+		{
+			"github.com",
+			validateGithub,
+		},
+	}
+	for _, link := range links {
```
mtojek, line 171: This is the reason why we have `ValidationErrors` (plural). You can run the validation against all entries and combine errors into an array.

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -127,3 +154,43 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
...
+	return nil
+}
+
+func validateGithub(ghLink *url.URL) error {
+	prNum, err := strconv.Atoi(path.Base(ghLink.Path))
+	if err != nil {
+		return err
+	}
+	if prNum == 0 {
+		return errors.New("zero is not a valid pull request number")
```
jsoriano, line 193: Nit.
```suggestion
	if prNum >= 0 {
		return errors.New("issue number in changelog link should be a positive number")
```

jsoriano (commented):
As mentioned in the original issue, I would prefer not to couple the format link to github, but we can be this strict now and relax later when there is more broad usage.

code/go/internal/validator/semantic/version_integrity.go
r00tu53r: I wasn't sure if it was okay to validate all the older links as that could break builds for something that was in built a while ago and other changes were built on top of it.

r00tu53r: The code now checks all links.

code/go/internal/validator/semantic/version_integrity.go
r00tu53r: The individual errors are now being appended to `ValidationErrors`.

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -127,3 +149,45 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
 	}
 	return errors.New("current manifest version doesn't have changelog entry")
 }
+
+func ensureLinksAreValid(links []string) ve.ValidationErrors {
+
```
mtojek, line 154: nit: empty line

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -127,3 +149,45 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
...
+		}
+		for _, vl := range validateLinks {
+			if strings.Contains(linkURL.Host, vl.domain) {
+				if err = vl.validateLink(linkURL); err != nil {
+					errs.Append(ve.ValidationErrors{err})
+				}
+			}
+		}
+	}
+	return errs
```
mtojek, line 184: I might be wrong, but I thought that we set `errs` when there is at least one error in the array. Otherwise, it's nil.

code/go/internal/validator/semantic/version_integrity.go
```diff
@@ -127,3 +149,45 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
...
+		if err != nil {
+			errs.Append(ve.ValidationErrors{
+				fmt.Errorf("invalid URL %v", err),
+			})
+			continue
+		}
+		for _, vl := range validateLinks {
+			if strings.Contains(linkURL.Host, vl.domain) {
+				if err = vl.validateLink(linkURL); err != nil {
+					errs.Append(ve.ValidationErrors{err})
```
mtojek, line 179: nit: if `validateFn` returns `ValidationErrors` you wouldn't need the transformation here

code/go/internal/validator/semantic/version_integrity_test.go
```diff
@@ -0,0 +1,72 @@
...
+package semantic
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestChangelogLinks(t *testing.T) {
+
```
mtojek, line 14: nit: empty line

code/go/internal/validator/semantic/version_integrity_test.go
```diff
@@ -0,0 +1,72 @@
...
+	var tests = []struct {
+		name      string
+		links     []string
+		numErrors int
+	}{
```
mtojek, line 19: Hm... I think you can also prepare table tests with given URLs and expected outcomes (nil or error).

Then, I would add another single unit test to check if `ensureLinksAreValid` iterates over all links (one test with few broken URLs and few valid ones).

code/go/internal/validator/semantic/version_integrity.go
r00tu53r: Yes for that case the `errs` is nil. For other cases `errs` has errors (if found).

code/go/internal/validator/semantic/version_integrity_test.go
r00tu53r: The code has been refactored to accomodate other link types and error handling.

mtojek (commented):
nit-picks

code/go/internal/validator/semantic/version_integrity_test.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			errs := ensureLinksAreValid(test.links)
+			if test.errors == nil {
+				assert.Equal(t, test.errors, errs)
+			} else {
+				assert.Equal(t, len(errs), len(test.errors))
+			}
```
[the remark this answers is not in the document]
mtojek: Ah, ok :) Thanks for checking this.

mtojek (approved):
LGTM!

r00tu53r:
> LGTM!

@mtojek thanks for the review! I don't seem to have permissions to merge this PR.

mtojek:
Let's wait for review @jsoriano and then we'll merge it.

jsoriano:
Merged, thanks!

Issue #129, opened 2021-02-05
[Ingest Pipelines][discuss] Support custom pipelines 


simitt:
## Feature Request
Allow users to customize Ingest Pipelines per data stream. This might be done in multiple iterations, e.g. 
(1) allow to disable loading the default pipelines, and documenting how custom pipelines can be set up manually. Removing pipelines could break data streams. The package-spec could be extended to define which pipelines are required and which are optional. Optional pipelines could then be disabled by users. 
(2) support customizing default pipelines. Building on top of (1) the spec could be extended to allow adding custom pipelines. 

## Why is it important
APM Server today allows users to define pipelines via a [json file](https://github.com/elastic/apm-server/blob/master/ingest/pipeline/definition.json). It allows users to configure whether ingest pipelines should be [set up and overwritten](https://github.com/elastic/apm-server/blob/master/apm-server.yml#L54-L58). When migrating to managed APM Server users should at least be able to set up their pipelines manually and ensure these are not overwritten. 

## Current State
@ruflin started a [docs PR](https://github.com/elastic/beats/pull/23247) to document how users can set up their own pipelines.

axw:
One challenge around customising pipelines is that they are renamed by Fleet, based on the data stream and package version. When you upgrade the package, new pipelines will be installed and customisations to existing ones will be lost. Also, because pipelines may change between package versions, I'm not sure we can ever sensibly carry customisations across.

I'm inclined to find a solution that does not involve either manually modifying pipelines that are installed by Fleet, or changing which of those pipelines are run. Instead, can we limit the feature to enabling users to define a pipeline which runs either before or after the default pipeline (or both)? This could be implemented in multiple ways, e.g. post-default by using a final pipeline, or by having Fleet inject pipeline processor(s) into the package-defined pipeline, referencing the user-specified pipeline(s).

jalvz:
Users have already other ways to create ingest pipelines, right? (ingest API, Kibana) What if we don't give them another way to do it and resort existing mechanisms? Or in other words: how the suggested way is better/simpler?

axw:
The problem is not in modifying a pipeline, it's in carrying it across package upgrades. Users do not have an existing mechanism for that. They can create their own custom pipeline using the usual mechanisms.

Maybe we could just allow users to modify the pipeline installed by Fleet, and have Fleet perform a three-way merge with the new pipeline on upgrade. I'm not convinced that'll be simpler, but it's certainly an option.

ruflin:
[...]

PR #325, opened 2022-04-27
Add sample "input" packages

mtojek:
This PR is a follow-up on the [design document](https://docs.google.com/document/d/1vzBh2frnlxEBBcab8-ist8pu_PDZUjOHdsV-QuUn_WA/edit).

I will start with the sample package and once we agree on properties, I will write down the spec.


test/packages/custom_logs/manifest.yml
```diff
@@ -0,0 +1,46 @@
...
+description: >-
+  Collect custom logs with Elastic Agent.
+type: input
+version: 1.2.3
+release: ga
+license: basic
+categories:
+  - custom
+policy_templates:
+  - name: first_policy_template
```
mtojek, line 13: I didn't put the `elasticsearch` block here. Should I add few/all options from [here](https://github.com/elastic/package-spec/blob/ff81936f9495f64c26722314e5e86355f575f92f/versions/1/integration/data_stream/manifest.spec.yml#L179)?

test/packages/custom_logs/manifest.yml
```diff
@@ -0,0 +1,46 @@
...
+    description: Collect your custom log files.
+    template_path: input.yml.hbs
+    vars:
+      - name: paths
+        type: text
+        title: Paths
+        multi: true
+        required: true
+        show_user: true
+        default:
```
ruflin, line 25: I know this is a sample package but I wonder if an input package should even have defaults here?

test/packages/custom_logs/manifest.yml
```diff
@@ -0,0 +1,46 @@
...
+title: Custom Logs
+description: >-
+  Collect custom logs with Elastic Agent.
+type: input
+version: 1.2.3
+release: ga
+license: basic
+categories:
+  - custom
+policy_templates:
```
ruflin, line 12: Do policy templates make sense for input packages? It is really convenient that we can reuse the same logic as we already have so I'm not necessarily asking to change anything here but have a discussion. 

* Can an input package have more then one policy_templates?
* Does an input package need all features of policy_templates?

ruflin (commented):
++ on having 2-3 sample packages to derive the spec.

test/packages/custom_logs/manifest.yml
mtojek: I understand that "defaults" can limit the amount of work on the initial configuration. There might be cases where you don't know what to do and would like to kick off the ingesting pipeline with default settings.

test/packages/custom_logs/manifest.yml
mtojek: > Does an input package need all features of policy_templates?

It looks like we can improve on this. These are the properties available in integration's `policy_template`:
* name
* title
* categories
* description
* data_streams
* inputs
* multiple (?)
* icons
* screenshot

Clearly, we don't need all of them, so I assumed that this `policy_template` (a different one in terms of spec) will contain a subset (or a different set) of properties. If you consider a different name for `policy_template`, do you have any suggestions? maybe an `input_template`?

> Can an input package have more then one policy_templates?

Good question, but I believe that the answer might be the same as for `policy_templates`. In most cases, we will need one. What should we do in the future if we scope this for one only :)?

test/packages/custom_logs/fields/input.yml
```diff
@@ -0,0 +1,4 @@
+- name: input.name
```
jsoriano, line 1: I think this field was called `input.type` in Beats. Are we intentionally using `input.name` now?

test/packages/custom_logs/manifest.yml
jsoriano: Yeah, I wouldn't see a reason not to support the same options. Unless there is something we want to deprecate from the data stream manifests.

test/packages/custom_logs/manifest.yml
jsoriano: I would keep support for `default`, but I wouldn't put any default `paths` in the package for custom logs. I think that we can expect that a user knows at least what logs they want to collect :)

An example of a `default` in a log input package could be to have a variable to support [`ignore_older`](https://www.elastic.co/guide/en/beats/filebeat/8.1/filebeat-input-filestream.html#filebeat-input-filestream-ignore-older), that would default to 0.

jsoriano (commented):
Nice.

test/packages/custom_logs/fields/input.yml
mtojek: No no, it's just a sample, nothing mandatory. It's here to show that you can define input fields.

test/packages/custom_logs/manifest.yml
mtojek: So, we have:

`elasticsearch.index_template.settings`
`elasticsearch.index_template.mappings`
`elasticsearch.index_template.ingest_pipeline.name`
`elasticsearch.privileges.indices`

As we don't know in advance what would be the target index/ingest pipeline, we can't set these properties:
`elasticsearch.index_template.ingest_pipeline.name`
`elasticsearch.privileges.indices`

Am I right?

test/packages/custom_logs/manifest.yml
mtojek: This is not real `Custom logs` package. I rather wanted to present all options there, which seem to be valid in terms of spec.

test/packages/custom_logs/manifest.yml
ruflin: Does the spec allow to use the same name but for example if it is defined as input package, to not allow certain params?

test/packages/sql_input/manifest.yml
```diff
@@ -0,0 +1,67 @@
...
+        show_user: true
+        default: "SHOW GLOBAL STATUS LIKE 'Innodb_system%'"
+      - name: sql_response_format
+        type: text
+        title: Response format
+        description: "Supported response formats: variables, table"
+        multi: false
+        required: true
+        show_user: false
+        default: "variables"
```
mtojek, line 57: For such cases, it would be great to have a pattern or allowed values. cc @joshdover

test/packages/custom_logs/manifest.yml
jsoriano: > As we don't know in advance what would be the target index/ingest pipeline, we can't set these properties:
> `elasticsearch.index_template.ingest_pipeline.name`
> `elasticsearch.privileges.indices`

Umm, I don't see these are used in current integrations repo. What are use cases for these options in integration packages? Could they apply in input packages?

In any case, maybe you are right with not adding the block by now, so we start small and we see later possible use cases.

test/packages/custom_logs/manifest.yml
jsoriano: Well, you can have `paths` as example without defaults, and `ignore_older` as example with defaults :slightly_smiling_face:

mtojek: Fixed!

test/packages/custom_logs/manifest.yml
mtojek: > Umm, I don't see these are used in current integrations repo. What are use cases for these options in integration packages? Could they apply in input packages?

If you can find any package here, then most likely it's APM or Endpoint.

mtojek:
Do you think that we need more sample packages or is it enough to start drafting spec files?

test/packages/custom_logs/agent/stream/input.yml.hbs
```diff
@@ -0,0 +1,12 @@
+paths:
+{{#each paths}}
+  - {{this}}
+{{/each}}
+
+tags:
+{{#each tags}}
+  - {{this}}
+{{/each}}
+
```
P1llus, line 10: I think there is a few minimum additions to this, that needs to be in every input package, and many of them are in other integrations as well:

Pipelines:
```
{{#if pipeline}}
pipeline: {{pipeline}}
{{/if}}
```

Tags are slightly wrong, they should be like this:

```
{{#if tags}}
tags:
{{#each tags as |tag i|}}
  - {{tag}}
{{/each}}
{{/if}}
```

Since an input can ingest data both from a local host and external host, we need to add support for removing the host fields:

```
{{#contains "forwarded" tags}}
publisher_pipeline.disable_host: true
{{/contains}}
```

And custom processors:
```
{{#if processors}}
processors:
{{processors}}
{{/if}}
```

test/packages/custom_logs/manifest.yml
```diff
@@ -0,0 +1,46 @@
+format_version: 1.0.0
+name: custom_logs
+title: Custom Logs
+description: >-
+  Collect custom logs with Elastic Agent.
```
P1llus, line 5: We should add a requirement that category and description has to include the word "Custom", would be nice for filtering as well I believe?

test/packages/custom_logs/agent/stream/input.yml.hbs
jsoriano: Custom processing has been intentionally left out of the MVP, it would also require custom mappings.

Also, depending on how this is implemented, this wouldn't require anything on the spec, the pipelines and the mappings could be configured in Fleet and installed by it.

mtojek: I added a few of your ideas to the sample package.

test/packages/custom_logs/manifest.yml
mtojek: I see the benefits of keeping the `custom` category and appropriate description, but not sure if this should be a hard requirement. I'd like to see comments from other folks.

jsoriano: For the category I don't have a strong opinion, but if required, maybe it'd be better to use an `input` category. For the description I would prefer to don't have any restriction. A common use case is to centralize logs without any parsing, and no "customization" applies there.

Looking to the historical background, Filebeat inputs are not described with the "custom" word. An input like this one could be described as in Filebeat:
```suggestion
  Read lines from active log files with Elastic Agent.
```

ruflin:
Should we add one input package that has an open endpoint like http as an additional example? https://github.com/elastic/package-storage/tree/production/packages/http_endpoint/1.0.1 @andrewkroh Would be good to get your eyes also on this.

mtojek:
> Should we add one input package that has an open endpoint like http as an additional example? https://github.com/elastic/package-storage/tree/production/packages/http_endpoint/1.0.1 

I think the more useful would be `httpjson` and I'm working on that one.

ruflin:
httpjson is useful too. The reason I mentioned the http_endpoint is that it is different from the others as it is a push instead of pull endpoint.

test/packages/custom_logs/agent/stream/input.yml.hbs
```diff
@@ -0,0 +1,18 @@
...
+{{#if tags}}
+tags:
+{{#each tags as |tag i|}}
+  - {{tag}}
+{{/each}}
+{{/if}}
+
+{{#if pipeline}}
+pipeline: {{pipeline}}
```
jsoriano, line 14: I would remove this from here. Ok that this is only an example in the config, but it may look like if we support custom processing using pipelines, but is something that we are not doing in the MVP.

test/packages/sql_input/manifest.yml
jsoriano: I played with patterns (as regexp) here https://github.com/elastic/package-spec/pull/245. I would prefer a list of allowed values :)

jsoriano (approved):

test/packages/custom_logs/agent/stream/input.yml.hbs
jsoriano: We have settings in input-like packages that happen to support custom processing, but this is not completely integrated with the solution, and is not enforced to be coherent with other inputs, or to require field mappings for possible new fields that are generated by this processing.
If we follow this way, every input package needs to implement its own way to support custom processing and custom fields. Now that we start with a clean state for this new package type, we would like to provide a more integrated solutions for this.

Custom pipelines support is still an [open discussion](https://github.com/elastic/package-spec/issues/129) for packages in general.

We consider that inputs without any custom processing already provide value in use cases of centralized log collection, and we are currently planning to go in this direction for the initial MVP. Next steps after that would be towards supporting custom processing.

It is though a good question how we are going to migrate users from the current input-like integration packages to the new input packages. Depending on the answer we may need to maintain these settings for backwards compatibility, maybe deprecating them, and eventually removing them when we have a complete solution for this.

PR #327, opened 2022-04-29
Add flags to allow for routing to dynamic datasets and namespaces

### LINKED ISSUE #315: [Change Proposal] Ability to add privileges for all data streams of a specific type
Currently, the data stream permissions are specific to the dataset that an integration defines.
For example, when adding a custom log integration, you'll have to specify the `data_stream.dataset`, for example `foo`, and an API key will be generated with the permissions to send data to `logs-foo-default`.

This is limiting for integrations that define a data stream that routes events to other data streams. See:
- https://github.com/elastic/integrations/pull/2972

Example: a log line like this gets sent to the `logs-ecs_router-default` data stream:
```
{
  "message": "{\"@timestamp\":\"2022-04-01T12:09:12.375Z\", \"log.level\": \"INFO\", \"message\":\"With event.dataset\", \"data_stream.dataset\": \"foo\"}"
}
```

The default ingest pipeline for `logs-ecs_router-default` parses the JSON within the message and uses the `data_stream.dataset` from the log message to route the message to the `logs-foo-default` data stream (by overriding the `_index` field).

The issue is that this will lead to a [security exception](https://github.com/elastic/integrations/issues/1332#issuecomment-1087558333) as the API key used to ingest the data only has permissions to ingest data to `logs-ecs_router-default`.

This is also an issue for the [Azure springcloudlogs](https://github.com/elastic/integrations/tree/main/packages/azure/data_stream/springcloudlogs) integration: All logs are always sent to the `springcloudlogs` data stream, even if the logs are from different application and, thus, should ideally be routed to their own data streams. Other examples are CloudWatch, k8s logs, PCF logs, and httpjson.

This relates to the discussions about input-only packages but is an independent and decoupled task.

What I'm proposing is to add a flag to the package spec that behaves similar to `dataset_is_prefix`
https://github.com/elastic/package-spec/blob/d0173302e6b2435556abc060d67f1e10d232f2bd/versions/1/data_stream/manifest.spec.yml#L129-L132

But instead of just [adding `.*` to the index permissions](https://github.com/elastic/kibana/blob/e5ba90c5e7beff01c545158cb676dc16c8c6db54/x-pack/plugins/fleet/server/services/package_policies_to_agent_permissions.ts#L148-L150), the flag will allow access to all data streams of a given type, such as `logs-*-*`.

@ruflin @mtojek @joshdover

felixbarny:
## What does this PR do?

Adds `data_stream.elasticsearch.dynamic_dataset`/`data_stream.elasticsearch.dynamic_namespace`.

When set to true, instead of granting permissions to the exact data stream the integration is sending data to (for example `logs-router-default`), it grants permissions for all datasets and namespaces, respectively, for the `data_stream.type` (for example (`logs-*-*`).

## Why is it important?

This is required for event routing. Instead of creating an API key that has permissions for an exact data stream, such as `logs-generic-default`, in event routing situations, the target data stream is determined within an ingest node pipeline.

See also 
* https://github.com/elastic/integrations/pull/2972
* https://github.com/elastic/elasticsearch/issues/63798

## Checklist

## Related issues
- Closes #315


versions/1/integration/data_stream/manifest.spec.yml
```diff
@@ -212,5 +212,10 @@ spec:
               type: array
               items:
                 type: string
+            allow_routing:
```
ruflin, line 215: If I remember correctly, so far this structure was built after the structure that Elasticsearch has. The advantage with this is that Fleet can just pass it on to fleet-server and it will just work. If this is the case, we should not allow_routing in here. Instead I think there is a `names` flag from Elasticsearch which could be used here? https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html#roles-indices-priv

Assuming `names` works, it would make the implementation a bit different in Fleet and the spec. The spec would have to validate that it does not go beyond the permissions the fleet-server has. On the spec side, it should be validated that it is inside the data stream naming scheme.

ruflin:
I made a different proposal directly in the spec to use existing Elasticsearch primitives. Instead of having something `routing` specific it would allow any package to specify additional privileges. But Fleet should at some point inform users about these additional privileges.

versions/1/integration/data_stream/manifest.spec.yml
felixbarny: I think I prefer the semantics that this PR proposes over letting an integration directly configure `names`. If we retain the semantics, I agree that it should move out of the `elasticsearch.privileges`.

The reason I prefer `allow_routing` is that it only allows elevated privileges on data streams that conform to the `type` that the integration has specified. This ensures the integration only has the permissions it needs. A logging integration will never need privileges to `metrics-*-*`, for example.  

Therefore, I propose to just add the `allow_routing` flag to the top level:

```yaml
type: logs
allow_routing: true
```

This would result in privileges for `logs-*-*`.

ruflin: > data streams that conform to the type that the integration has specified

This might not always be the case. For example currently firehose only supports logs but soon is adding metrics. I'm not sure if it comes out the same stream or a separate setup is needed (still to be figured out). If it comes out of the same stream, it will need routing to logs and metrics.

The reason I like `names` is because we don't have to introduce a new config option just yet and wait on a few "routing integrations" to happen and then potentially introduce a specific config option for these that is more restrictive and descriptive. I'm worried we take one example and abstract based on it.

ruflin:
@joshdover @jsoriano @aspacca Would be good to get some additional opinions on the discussion in https://github.com/elastic/package-spec/pull/327#discussion_r862843182 Most important, lets get this moving forward.

versions/1/integration/data_stream/manifest.spec.yml
felixbarny: >  If it comes out of the same stream, it will need routing to logs and metrics.

Currently, the [`data_stream_router`](https://github.com/elastic/elasticsearch/pull/76511) can't change the `type` of the data stream to avoid that an event is accidentally or maliciously routed from `logs-router-default` to a non-logs data stream.

If that's a requirement for firehose or others, we'll also need to change that.

ruflin: Lets consider it a feature that it is not possible at the moment and add it only when needed.

test/packages/good/data_stream/foo/manifest.yml
```diff
@@ -1,5 +1,6 @@
 title: Nginx access logs
 type: logs
+allow_routing: true
```
mtojek, line 3: If the field `allow_routing` isn't supposed to be mandatory, I guess that you can leave only for the `good` package and drop it from other places. You don't need to verify the same feature multiple times.

versions/1/integration/data_stream/manifest.spec.yml
```diff
@@ -143,6 +143,11 @@ spec:
     hidden:
       description: Specifies if a data stream is hidden
       type: boolean
+    allow_routing:
+      description: When set to true, the integration is granted data stream privileges for all datasets and namespaces of its type
```
mtojek, line 147: nit: s/the integration/the package/

versions/1/integration/data_stream/manifest.spec.yml
```diff
@@ -143,6 +143,11 @@ spec:
...
+      type: boolean
+      items:
+        type: string
```
mtojek, line 150: Something is wrong here. The field is boolean, but you defined string items for it as it's an array.

mtojek (changes requested):

versions/1/integration/data_stream/manifest.spec.yml
```diff
@@ -143,6 +143,9 @@ spec:
     hidden:
       description: Specifies if a data stream is hidden
       type: boolean
+    allow_routing:
+      description: When set to true, the package is granted data stream privileges for all datasets and namespaces of its type
```
joshdover, line 147:
```suggestion
      description: When set to true, agents running this integration are granted data stream privileges for all datasets and namespaces of its type
```

joshdover (dismissed):
One suggestion on the description but overall LGTM.

@mtojek, question: do we have a way to specify new fields in the package-spec that aren’t considered “stable”? It’d be nice to be able to add fields to the spec to allow experimentation with new features before we make a hard commitment to it. We may want to explore other ways of configuring a feature in the future.

That said, I’m ok with this one going forward anyways, since it supporting it from the Fleet side should be quite trivial, but just something I think we should be thinking about. Being able to mark a package-spec field as beta or experimental could be valuable.

ruflin:
@joshdover It would be great if the PR for this feature on the Fleet side could be opened in parallel, so it can be merged at the same time. Otherwise I'm worried we have the case that it is in the for but not in Fleet yet.

mtojek:
> @mtojek, question: do we have a way to specify new fields in the package-spec that aren’t considered “stable”? It’d be nice to be able to add fields to the spec to allow experimentation with new features before we make a hard commitment to it. We may want to explore other ways of configuring a feature in the future.

I'm afraid that we don't have it at the moment as it seems to be hard to ensure compatibility with stacks. Consider the case, when the field is beta, but its type changes once it's GA. It's hard to predict the consequences of it.

mtojek (approved):
LGTM from the Ecosystem team, let's wait for @joshdover's official approval.

joshdover:
> @joshdover It would be great if the PR for this feature on the Fleet side could be opened in parallel, so it can be merged at the same time. Otherwise I'm worried we have the case that it is in the for but not in Fleet yet.

I started implementing this in https://github.com/elastic/kibana/pull/131609 but found that it’s going to be dependent on https://github.com/elastic/kibana/issues/115032 which I am assigned to for 8.3 and may be able to finish up this week. The issue is that we’re not reading the data stream information directly from the packages today, we’re using the registry API’s summary during package installation. We want to stop duplicating all of these fields that are only used during install and pull this info directly from the package instead.

I’ll follow up finishing this up after the blocking task is done. The amount of work is very small.

joshdover:
> Are the dependent tasks still expected to land in 8.3?

They got delayed, but I have requested they are included for 8.4.

test/packages/good/data_stream/foo/manifest.yml
jsoriano, line 3: As this is related to elasticsearch configuration, I think it could be at the root level of `elasticsearch`, even if it doesn't belong to the `elasticsearch.privileges` object https://github.com/elastic/package-spec/pull/327#discussion_r862843182.

We did something similar for the time series mode in https://github.com/elastic/package-spec/pull/357/files.

jsoriano: @felixbarny is this change still required? Wdyt about placing this parameter under `elasticsearch` object?

felixbarny: Yes, this is still required. Are all the blockers resolved now? I don’t care so much about where the flag is. Happy to change it to be under elasticsearch if you think that’s a better place.

jsoriano: This blocker was solved https://github.com/elastic/kibana/issues/115032, not sure if there were any others.

Yes, please update the branch to put the new setting under `elasticsearch`. We removed the `versions` directory, so you will also need to move the changes to the new structure.

felixbarny: I've moved it and renamed and split the property so that it's both more self-explanatory and flexible.

jsoriano (dismissed):
LGTM, do you know if there is an open issue in kibana to support these settings?

jsoriano:
> LGTM, do you know if there is an open issue in kibana to support these settings?

I guess it is this one https://github.com/elastic/kibana/issues/134971

jsoriano (approved):

joshdover:
We've had a use case come up that this is close to solving but doesn't quite hit the mark.

Users may want to add custom routing to ingest pipelines created for input-type packages to route to more specific pipelines. To do this, it'd be required that the API key granted to the Agent running the integration has adequate privileges. Ideally, this wouldn't be turned on by default, and instead would be an option in the integration policy, rather than always on for all instances of an input package.

@jsoriano what do you think about extending this package-spec option to be `allow_routing: false | true | 'optional'`? This would allow to expose this option in the integration policy editor in the UI.

jsoriano:
> Users may want to add custom routing to ingest pipelines created for input-type packages to route to more specific pipelines.

> This would allow to expose this option in the integration policy editor in the UI.

Is this something that we would want for some packages but not for others? If we add the flag to the package spec, we are making package developers to take the decision on this, but it looks like something that the user should decide.

If a user adds custom routing using the integration policy editor, Fleet could (should?) ensure that the Agent receives an API key with adequate privileges. This doesn't look like a responsibility of the package developer.

PR #328, opened 2022-05-04
Write input package spec

mtojek:
Follow-up on https://github.com/elastic/package-spec/pull/325

This PR defines the spec for the input package. I spotted a few problems with sample packages and addressed them here too:

1. Forbid already deprecated the "release" tag for input packages.
2. Replace "agent/stream" with "agent/input". Apparently, that directory was also defined for integration.
3. Fix changelog URL link.

Notes:

In the "input" spec I refer to the "integration" spec files to prevent redundancy of definitions. I decided not to extract common parts as there are only two specifications.


ruflin:
@joshdover Lets make sure we have the Fleet PR open to support this in parallel to make sure it will work as expected.

versions/1/integration/spec.yml
```diff
@@ -1,9 +1,8 @@
 ##
-## Entrypoint of package specification.
+## Entrypoint of "integration" specification.
```
ruflin, line 2:
```suggestion
## Entrypoint of "integration packages" specification.
```

versions/1/integration/manifest.spec.yml
```diff
@@ -1,5 +1,5 @@
 ##
-## Describes the specification for the package's main manifest.yml file
+## Describes the specification for the integration's main manifest.yml file
```
ruflin, line 2: See below, I think we should talk about integration packages.

versions/1/integration/spec.yml
mtojek: Fixed

versions/1/integration/manifest.spec.yml
mtojek: Fixed

mtojek:
> @joshdover Lets make sure we have the Fleet PR open to support this in parallel to make sure it will work as expected.

Don't hurry up with it, there is plenty of work to be done on the elastic-package and Package Registry side (covered in the [design doc](https://docs.google.com/document/d/1vzBh2frnlxEBBcab8-ist8pu_PDZUjOHdsV-QuUn_WA/edit#heading=h.grlm10yq9sb8)).

versions/1/input/changelog.spec.yml
```diff
@@ -0,0 +1,44 @@
+##
+## Describes the specification for the package's CHANGELOG file
+##
+spec:
```
jsoriano, line 4: Is it not possible to `$ref` the whole spec from the integration one? Or are there differences between them?

versions/1/input/manifest.spec.yml
```diff
@@ -0,0 +1,131 @@
...
+          description: Kibana conditions
+          type: object
+          additionalProperties: false
+          properties:
+            version:
+              type: string
+              description: Kibana versions compatible with this package.
+              examples:
+                - ">=7.9.0"
+    policy_templates:
```
jsoriano, line 63: Could we `$ref` the whole policy template definition? Or this one has differences with the `integration` one?

versions/1/input/spec.yml
```diff
@@ -0,0 +1,42 @@
...
+## Entrypoint of "input packages" specification.
+## The specification is considered "beta" at the moment, so it may change until we release it as GA.
+##
+## Describes the folders and files that make up a package.
+##
+spec:
+  additionalContents: true
+  totalContentsLimit: 65535
+  totalSizeLimit: 250MB
+  sizeLimit: 150MB
```
jsoriano, line 11: Maybe we can reduce these limits. I cannot think on a reason to have big files in inputs :thinking:
We could increase this later if we find a use case.
```suggestion
  totalSizeLimit: 50MB
  sizeLimit: 30MB
```

jsoriano (commented):
> In the "input" spec I refer to the "integration" spec files to prevent redundancy of definitions. I decided not to extract common parts as there are only two specifications.

I would still consider adding a folder with shared/common definitions, even if we only have two specifications by now. But I would do it in a follow up so it is clearer what is being added here.

versions/1/input/spec.yml
mtojek: Fixed.

versions/1/input/changelog.spec.yml
mtojek: It worked! I referred to the changelog for integration.

mtojek (commented):
> I would still consider adding a folder with shared/common definitions, even if we only have two specifications by now. But I would do it in a follow-up so it is clearer what is being added here.

I admit that I aligned the implementation with a rule of three, so when there is a new type I can extract a common part. If you find it clearer to do it now, I can focus on it in a follow-up.

jsoriano (dismissed):

mtojek:
@joshdover Friendly reminder about this PR :)

versions/1/integration/data_stream/spec.yml
```diff
@@ -25,12 +25,7 @@ spec:
       name: agent
```
joshdover, line 25: What's the difference between the agent directory in a `data_stream` vs. one in the overall package? Can we document this in the spec?

joshdover (commented):
I've taken a thorough look based on the spec changes, but I'll admit it's hard to know if this supports everything we need in the UI...without knowing what we want the UI to look like. @akshay-saraswat I think we need to produce some mocks on this before we can merge this. 

I also agree with @ruflin that we should have a Kibana PR up in parallel before merging to validate the spec changes before committing to them. That will require scheduling capacity on the Fleet UI team for 8.4. I spoke to @jen-huang about this yesterday.

---

Would it make sense for input packages to define a list of allowed data types for the data stream? For instance, httpjson could be used for metrics, logs, or traces, but not synthetics. Do we want this to be user configurable? It wasn't clear in the definition doc.

versions/1/input/manifest.spec.yml
```diff
@@ -0,0 +1,131 @@
...
+              - Apache logs and metrics
+          description:
+            description: Longer description of policy template.
+            type: string
+            examples:
+              - Collect logs and metrics from Apache instances
+          icons:
+            $ref: "../integration/manifest.spec.yml#/definitions/icons"
+          screenshots:
+            $ref: "../integration/manifest.spec.yml#/definitions/screenshots"
```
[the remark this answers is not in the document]
mtojek: I'd like rather leave an option to place any picture here, even for instruction purposes. If you consider this to be useless, I can drop the field.

versions/1/integration/data_stream/spec.yml
mtojek: Agent in spec is used by few integrations, which don't need a data stream, but they are still integrations:

https://github.com/elastic/integrations/blob/main/packages/fleet_server/agent/input/agent.yml.hbs
https://github.com/elastic/package-storage/blob/production/packages/apm/8.2.0/agent/input/template.yml.hbs

I was confused too, but didn't want to merge "agent" definitions into single one. Let me know what do you think.

mtojek:
> I've taken a thorough look based on the spec changes, but I'll admit it's hard to know if this supports everything we need in the UI...without knowing what we want the UI to look like. @akshay-saraswat I think we need to produce some mocks on this before we can merge this.

I don't have preference about this. Maybe it's because of the fact, that there will be a lot work around it to make it live :) Maybe we should release the input type as a beta to unblock other works?

joshdover:
> Maybe we should release the input type as a beta to unblock other works?

That works for me. Can we enforce somehow that no input packages get published as GA before we've moved the spec out of beta?

mtojek:
I think that we should be able to implement that feature.

joshdover (approved):
LGTM for beta

jsoriano (approved):
:+1:

PR #329, opened 2022-05-09
Prepare 1.8.0

jsoriano:
## What does this PR do?

Release 1.8.0.

## Why is it important?

It includes additional validations and support for osquery saved queries.

Additional validations will break integrations with duplicated fields definitions.


mtojek (approved):
I think that we make "those" PRs self-approved. It's just a version bump.

PR #335, opened 2022-05-17
Add custom agent deployer spec

marc-gr:
## What does this PR do?

Adds the required sped for the custom agent deployer

## Why is it important?

to be able to properly validate the test structure

## Checklist

## Related issues

-

- Relates https://github.com/elastic/elastic-package/pull/786


test/packages/deploy_custom_agent_invalid_property/changelog.yml
```diff
@@ -0,0 +1,5 @@
+- version: 0.1.2
+  changes:
+    - description: initial release
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/131
```
mtojek, line 5: nit: I used to link with the real PR number (so this one)

mtojek (commented):
Could you please also add a positive sample to the `good` package? For example a new data stream with a test?

mtojek (approved):

PR #336, opened 2022-05-17
Bump up version to v1.9.0

mtojek:
This PR bumps up the package-spec to 1.9.0

It also fixes the redundant changelog.


mtojek:
self-approved

PR #338, opened 2022-05-19
Re-enable validation of duplicated fields

### LINKED ISSUE #331: Enable check to avoid duplicated fields
This is a follow up of https://github.com/elastic/package-spec/pull/309.

We have found some issues we have to investigate. It seems that elastic-package is importing more fields than expected when using `external: ecs`, what complicates checking for fields duplication.
Also, fields validator in `elastic-package` uses only the fields defined in the source files, so it is sometimes not possible to pass both the validation of defined fields, and the validation to avoid duplicated fields. Original idea of the fields importer was to import only fields defined explicitly and not subfields of groups, so it is ok if fields validator only uses the source files.

So current plan is going to be:
- [x] Disable check for fields duplication and release package-spec 1.8.1 without this. https://github.com/elastic/package-spec/pull/332
- [x] Revert https://github.com/elastic/elastic-package/pull/814 as it shouldn't be needed. https://github.com/elastic/elastic-package/pull/817
- [x] Make `elastic-package` to import only fields explicitly configured, and not subfields. https://github.com/elastic/elastic-package/pull/818
- [ ] Remove duplicated fields from integration packages using a similar approach to the one to remove invalid ECS values (https://github.com/elastic/integrations/pull/3392). **Update:** Waiting for https://github.com/elastic/package-spec/issues/345, to do it package per package.
- [ ] Once everything is clear, enable the check again (https://github.com/elastic/package-spec/pull/338).

jsoriano:
## What does this PR do?

Re-enables check to avoid duplicated fields definitions.

## Why is it important?

To avoid ambiguous definition of fields.

## Checklist

## Related issues

- Fixes #331.


jsoriano:
Done as part of #384

PR #339, opened 2022-05-20
Add secret option to input vars

joshdover:
## What does this PR do?

This adds a new `secret` option to input vars to support securely stored values. These values can only be referenced via the upcoming `secret` provider in Elastic Agent. See https://github.com/elastic/ingest-dev/issues/999 for more context.

## Why is it important?

We need to securely store, transmit, and redact (eg. in a diagnostic) sensitive fields in agent policies.

## Checklist


mtojek (commented):
Please remember about updating the changelog file to include this PR :)

ph (dismissed):
This look to me, using a question as a description is strange at first but this is the norm when looking at the other boolean flags defined in the package spec.

I assume this is only additiona logic and we can communicate that change with the integrations teams and they can start using it.

@joshdover maybe create an issue to audit for common secrets fields:

- password
- key
- token

joshdover:
> There might be some UI definition that needs to happen, when you are editing an integration that has already a private key defined, you want to see that we have something set for the field but not the actual value.

I will put together a design doc for this end-to-end secret storage including UX around this.

jlind23:
@pierrehilbert @jen-huang we should let @joshdover work on this design doc first but then lets make sure that Control Plane and Fleet UI are aligned on this feature.

test/packages/input_groups/manifest.yml
```diff
@@ -30,6 +30,7 @@ vars:
   - name: secret_access_key
     type: text
     title: Secret Access Key
+    secret: true
```
mrodm, line 33: Could you add an example of field in the test package `good_v2` and `good_input`?

mrodm (commented):
Missign a new changelog entry for this feature.

One doubt about this feature, what is the behaviour in Kibana for those versions that do not support this secret ? Should we just allow this `secret` field from 2.9.0 or from certain kibana version ? 
cc @jsoriano

hop-dev:
@mrodm regarding backwards compatibility, the field will be ignored by older kibana versions and the package will not be affected, so I don't think a version constraint is necessary. 

Added examples + changelog in [a888ac4](https://github.com/elastic/package-spec/pull/339/commits/a888ac4acc7bfdcc96b77fe3f8e91c0be70001e9)

mrodm:
> @mrodm regarding backwards compatibility, the field will be ignored by older kibana versions and the package will not be affected, so I don't think a version constraint is necessary.

My concern is about that users/developers setting a field as `secret` but Kibana does not manage it as a secret and it could be shown afterwards. Should we disallow that? 
I would like to have other opinions about this, please @jsoriano if you can chime in here :pray:

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -83,6 +83,12 @@ spec:
             default: false
             examples:
               - true
+          secret:
```
mrodm, line 86: Not sure about removing this field in previous versions using a JSON Patch.
If we do so, probably it should be for 2.9.0 (and it would be needed to update changelog too) ? Would it be better to keep for all versions for security reasons ?  @jsoriano  WDYT ?

```yaml
versions:
  - before: 2.9.0 # other version 2.0.0 ? 
    patch:
      - op: remove
        path: /definitions/vars/items/properties/secret
```

spec/changelog.yml
```diff
@@ -4,6 +4,9 @@
 ##
 - version: 2.8.0-next
   changes:
+  - description: Add ability to specify secret vars
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/339
```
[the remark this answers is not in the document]
hop-dev: ah 🤦 Thanks!

hop-dev:
@juliaElastic Mario and I have had a chat about this, I think we will need to restrict it so that a package developer can only set a var as secret if they have a kibana version condition of 8.9 or greater.

we have function similar to this e.g for checking runtime fields https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/validate_minimum_kibana_version.go#L68

However we will need one to iterate over all of the vars in a package, including package level, input level and stream level vars and see if any is set to secret. I don't think I am going to have time to do this before I go I want to focus on getting the other PRs merged. @mrodm might be able to help with this next sprint if Jill feels she has enough on her plate?

juliaElastic:
Thanks for the update. Do we consider the condition as a requirement for this pr or should we raise a separate issue? 
If Mario can help out, that would be great.  cc @jlind23

jlind23:
@juliaElastic Thanks for the ping, as Mark is out @mrodm is the only one that can help us here.

jsoriano:
> > @mrodm regarding backwards compatibility, the field will be ignored by older kibana versions and the package will not be affected, so I don't think a version constraint is necessary.
> 
> My concern is about that users/developers setting a field as `secret` but Kibana does not manage it as a secret and it could be shown afterwards. Should we disallow that?

> Mario and I have had a chat about this, I think we will need to restrict it so that a package developer can only set a var as secret if they have a kibana version condition of 8.9 or greater.

In general I share the concern with Mario, it may be misleading for package developers to have an option silently ignored.

_But_ thinking about this specific case, it may be better to go on without the version check. If we force package developers to update the kibana version to use `secret`, they may need to maintain another package version if they want to keep supporting older versions of kibana, what can discourage them about using this secure setting. And at the end the result is going to be similar.

spec/integration/data_stream/manifest.spec.yml
jsoriano: +1 to add this to encourage updating the spec (and to update the version in the changelog).

mrodm:
> In general I share the concern with Mario, it may be misleading for package developers to have an option silently ignored.
> 
> _But_ thinking about this specific case, it may be better to go on without the version check. If we force package developers to update the kibana version to use `secret`, they may need to maintain another package version if they want to keep supporting older versions of kibana, what can discourage them about using this secure setting. And at the end the result is going to be similar.

Thanks @jsoriano for your comments!
Let's do that then, let's add this new option/field without adding a new validation for kibana versions. But let's force to update to 2.9.0 version of the package-spec as mentioned here https://github.com/elastic/package-spec/pull/339#discussion_r1192169924 @hop-dev @juliaElastic

jlind23:
@mrodm @jsoriano could you please re-review it?

jillguyonnet:
Hi @mrodm - I've pushed the changes and fixed the conflicts - can you have another look please?

mrodm (approved):

jillguyonnet:
@jsoriano Are you OK with merging this?

jsoriano (approved):
LGTM! :rocket:

PR #341, opened 2022-05-23
Define beta features in spec

mtojek:
This PR defines `release: beta` flag for folder specs. Requested by @joshdover in [comment](https://github.com/elastic/package-spec/pull/328#issuecomment-1133008672).

Related: https://github.com/elastic/package-spec/pull/328

It might be hard to test it without providing a fake spec, so I'd rather test it against https://github.com/elastic/package-spec/pull/328.

### DIFF
--- a/code/go/internal/validator/common_spec.go
+++ b/code/go/internal/validator/common_spec.go
@@ -19,6 +19,11 @@ type commonSpec struct {
 	DevelopmentFolder  bool             `yaml:"developmentFolder"`
 
 	Limits commonSpecLimits `yaml:",inline"`
+
+	// Release type of the spec: beta, ga.
+	// Packages using beta features won't be able to go GA.
+	// Default release: ga
+	Release string `yaml:"release"`
 }
 
 type commonSpecLimits struct {
--- a/code/go/internal/validator/folder_spec.go
+++ b/code/go/internal/validator/folder_spec.go
@@ -87,7 +87,9 @@ func (s *folderSpec) validate(pkg *Package, path string) ve.ValidationErrors {
 	}
 
 	// Don't enable beta features for packages marked as GA.
-	if s.Release == "beta" && pkg.Version.Prerelease() == "" {
+	if s.Release != "" && s.Release != "beta" {
+		errs = append(errs, errors.Errorf("unsupport release level, supported values: beta, ga"))
+	} else if s.Release == "beta" && pkg.Version.Prerelease() == "" {
 		errs = append(errs, errors.Errorf("spec for [%s] defines beta features which can't be enabled for packages with a stable semantic version", pkg.Path(path)))
 	} else if s.Release == "beta" && pkg.Version.Prerelease() != "" {
 		log.Printf("Warning: package with non-stable semantic version and active beta features (enabled in [%s]) can't be released as stable version.q", pkg.Path(path))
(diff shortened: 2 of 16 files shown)


code/go/internal/validator/folder_spec.go
```diff
@@ -71,44 +70,49 @@ func (s *folderSpec) load(fs fs.FS, specPath string) error {
...
 	// this limit in all cases to avoid having to read too many files.
 	if contentsLimit := s.Limits.TotalContentsLimit; contentsLimit > 0 && len(files) > contentsLimit {
-		errs = append(errs, errors.Errorf("folder [%s] exceeds the limit of %d files", fsys.Path(path), contentsLimit))
+		errs = append(errs, errors.Errorf("folder [%s] exceeds the limit of %d files", pkg.Path(path), contentsLimit))
 		return errs
 	}
 
+	// Don't enable beta features for packages marked as GA.
+	if s.Release == "beta" && pkg.Version.Prerelease() == "" {
+		errs = append(errs, errors.Errorf("spec for [%s] defines beta features which can't be enabled for packages marked as GA", pkg.Path(path)))
```
joshdover, line 90: nit:

```suggestion
		errs = append(errs, errors.Errorf("spec for [%s] defines beta features which can't be enabled for packages with a stable semantic version", pkg.Path(path)))
```

joshdover (dismissed):
LGTM, thanks for adding this. It would be nice if we could emit warnings when beta features are used in non-GA packages too so that developers aren't surprised right when they're hoping to make a package GA. Let's consider that nice-to-have though unless it's easy to implement now.

joshdover (dismissed):

code/go/internal/validator/folder_spec.go
```diff
@@ -71,44 +71,51 @@ func (s *folderSpec) load(fs fs.FS, specPath string) error {
...
-		errs = append(errs, errors.Errorf("folder [%s] exceeds the limit of %d files", fsys.Path(path), contentsLimit))
+		errs = append(errs, errors.Errorf("folder [%s] exceeds the limit of %d files", pkg.Path(path), contentsLimit))
 		return errs
 	}
 
+	// Don't enable beta features for packages marked as GA.
+	if s.Release == "beta" && pkg.Version.Prerelease() == "" {
+		errs = append(errs, errors.Errorf("spec for [%s] defines beta features which can't be enabled for packages with a stable semantic version", pkg.Path(path)))
+	} else if s.Release == "beta" && pkg.Version.Prerelease() != "" {
+		log.Printf("Warning: package with non-stable semantic version and active beta features (enabled in [%s]) can't be released as stable version.q", pkg.Path(path))
```
jsoriano, line 93: We need some way to return different "levels" of errors. A developer may want to have strict validation that fails on warnings too.

code/go/internal/validator/common_spec.go
```diff
@@ -19,6 +19,11 @@ type commonSpec struct {
 	DevelopmentFolder  bool             `yaml:"developmentFolder"`
 
 	Limits commonSpecLimits `yaml:",inline"`
+
+	// Release type of the spec: beta, ga.
+	// Packages using beta features won't be able to go GA.
+	// Default release: ga
+	Release string `yaml:"release"`
```
jsoriano, line 26: Should we check somewhere that we only use `beta` or `ga` here? Or even only `beta`?

jsoriano (approved):

code/go/internal/validator/folder_spec.go
jsoriano: Issue created: https://github.com/elastic/package-spec/issues/342

code/go/internal/validator/common_spec.go
mtojek: I added a condition to check for "beta" and "ga" and report an error whenever there is a different level used.

jsoriano (dismissed):

jsoriano (approved):

Issue #298, opened 2022-03-17
Add License and Subscription in integration details section


akshay-saraswat:
In integrations UI, if you click on a specific agent-based integration tile to navigate to the overview page, you will see a "license" row in the details section (highlighted by the red rectangle in the screenshot below). We have been using the term "License" incorrectly in this row. It's causing a lot of confusion because the value here is Basic which is actually a subscription tier, not a license.

### Acceptance Criteria:
- Replace "License" with "Subscription" here.
- Create a new row for a license with a new set of enum values to list ELv1, ELv2, or any other special licenses created for vector tiles or packages like ProblemChild.
- Provide a link to license text in the license row.

![image](https://user-images.githubusercontent.com/14043218/158758987-62cd1f5d-466d-48fc-822d-f0043bd03298.png)

jsoriano:
This will require changes in the spec and in fleet, pinging @joshdover for the changes in fleet.

Regarding the package spec, I would propose something like this:
* `license` field is deprecated to avoid confusion.
* Packages should include a `LICENSE` file in their root directory, with the content of their license.
* A new `condition` is supported, `stack.subscription`, that can have the different subscriptions as values.

In fleet:
* The "License" label is replaced by a "Subscription" label, the value would be:
  * The value in the `stack.subscription` condition if it is present.
  * The value in the `license` field otherwise (for compatibility with current packages).
* Not sure at what points we should show the `LICENSE`. If we use it only on installation time, and/or after installing the package, the `LICENSE` could be obtained from the package itself. If we also want to show it in the integrations UI, we may need to serve it from the registry.

Alternative to the `LICENSE` file in the root directory of the package:
* We limit the licenses that can be used for packages distributed through the registry.
* Add a `source` object to the manifest. This object contains information about the source of the package. `source.license` would be one of the allowed licenses. (This would allow to have other fields as `source.url` with the url to the source code of the package).
* `elastic-package` includes a `LICENSE` file in the zip files, with the text of the license indicated in `source.license`.
* Fleet can show the value of `source.license` and can have copies of the `LICENSE` texts in case they need to be shown in the integrations UI for packages not installed.

joshdover:
> Alternative to the `LICENSE` file in the root directory of the package:

[...]

PR #355, opened 2022-06-15
Add field for source license and condition for subscription

jsoriano:
## What does this PR do?

* Add a new `source.license` field, with the actual license of the package.
* Add a new `elastic.subscription` field, that allows to optionally declare a required subscription to install a package.
* Deprecate the current `license` field.

## Why is it important?

Current "license" field is referring to the subscription level, and there is no way to indicate the actual license of the package.

See https://github.com/elastic/package-spec/issues/298 for more information.

## Checklist

## Related issues

- Part of https://github.com/elastic/package-spec/issues/298.

### DIFF
--- a/test/packages/good/manifest.yml
+++ b/test/packages/good/manifest.yml
@@ -4,8 +4,11 @@ title: Good package
 description: This package is good.
 version: 1.0.0
 type: integration
+source:
+  license: "Apache-2.0"
 conditions:
   kibana.version: '^7.9.0'
+  elastic.subscription: 'basic'
 policy_templates:
   - name: apache
     title: Apache logs and metrics
--- a/test/packages/sql_input/manifest.yml
+++ b/test/packages/sql_input/manifest.yml
@@ -5,7 +5,10 @@ description: >-
   Execute custom queries against an SQL database and store the results in Elasticsearch.
 type: input
 version: 0.2.0
-license: basic
+source:
+  license: "Apache-2.0"
+conditions:
+  elastic.subscription: basic
 categories:
   - custom
   - datastore
--- a/versions/1/input/manifest.spec.yml
+++ b/versions/1/input/manifest.spec.yml
@@ -28,8 +28,11 @@ spec:
     version:
       description: The version of the package.
       $ref: "../integration/manifest.spec.yml#/definitions/version"
+    source:
+      $ref: "../integration/manifest.spec.yml#/definitions/source"
     license:
-      description: The license under which the package is being released.
+      description: The license under which the package is being released (deprecated, use subscription instead).
+      deprecated: true
       type: string
       enum:
       - basic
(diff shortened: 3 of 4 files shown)


jlind23 (approved):

test/packages/good/manifest.yml
```diff
@@ -4,8 +4,11 @@ title: Good package
 description: This package is good.
 version: 1.0.0
 type: integration
+source:
+  license: "Apache-2.0"
 conditions:
   kibana.version: '^7.9.0'
+  elastic.subscription: 'basic'
```
mtojek, line 11: Hm... it's a tough one. It is a condition or requirement to have a `basic` subscription. On the other hand, `conditions` will become a bag with various properties, which isn't great too. Maybe it can be moved to the `licensing` structure I proposed above?

mtojek (commented):
A few comments to consider to keep license/subscription options together.

test/packages/good/manifest.yml
```diff
@@ -4,8 +4,11 @@ title: Good package
 description: This package is good.
 version: 1.0.0
 type: integration
+source:
```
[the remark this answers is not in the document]
jsoriano: Not needed, but also we cannot use `license` as is the current problematic name.

We could have other metadata about the source, for example the url.

test/packages/good/manifest.yml
```diff
@@ -4,8 +4,11 @@ title: Good package
...
+  license: "Apache-2.0"
```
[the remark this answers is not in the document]
jsoriano: Umm, I consider this property implements this comment. This is a value, from an enum of allowed values. How does it violate the comment?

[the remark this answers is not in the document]
jsoriano: How would the `licensing` structure be? Something like this?
```
licensing:
  license: 'Apache-2.0'
  subscription: 'basic'
```

test/packages/good/manifest.yml
jsoriano: We need different fields for license and subscription.

The subscription is a requirement, this is why I thought that putting it as condition could make sense. I am ok with moving it to its own field though.

test/packages/good/manifest.yml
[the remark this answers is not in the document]
mtojek: In case we have multiple licenses per package we could use that structure:

```
licensing:
   licenses: [
      name: Apache-2.0,
      path: LICENSE.txt
   ]
   subscription: basic
```

Sometimes there is also `NOTICE.txt` file, which seems to be related to package license.

test/packages/good/manifest.yml
mtojek: Both options sound good, conditions may also contain regular expressions, for example, `basic || special`. Maybe we can ask other folks about their preferences. @joshdover

jsoriano: I was also thinking on the "at least" semantics.

Regarding `basic || special`, do we have any feature now that is included on different subscriptions? This would be for example to have something in `gold` and `enterprise` that is not included in `platinum`. This sounds confusing.

jsoriano: > Shouldn't we go ahead and add the other enum values: `basic`, `gold`, `platinum`, `enterprise`?

:+1: added.

jen-huang (approved):
Model LGTM from Fleet side

PR #359, opened 2022-06-22
[Cloud Posture] Migrating rule template schema - introducing metadata field

kfirpeled:
## What does this PR do?

In this PR we are updating the example of csp-rule-template

## Why is it important?

During the development of 8.4.0 we've refactored the schema of the rule templates and it is important to keep the examples in the spec updated as well

## Checklist

## Related issues

-


eyalkraft (approved):
LGTM

mtojek (dismissed):

jsoriano:
Is there anything pending here apart of the merge conflict?

jsoriano:
I have updated the branch with main, but after taking another look, I see the change is only in test files. We probably don't need a changelog entry for this.

jsoriano (approved):

PR #365, opened 2022-06-29
update csp rule template benchmark field

CohenIdo:
Update `benchmark` field in  `csp-rule-template` asset

### DIFF
--- a/test/packages/good/kibana/csp_rule_template/good-csp-rule-template-abc-1.json
+++ b/test/packages/good/kibana/csp_rule_template/good-csp-rule-template-abc-1.json
@@ -14,7 +14,7 @@
       "Kubernetes",
       "Containers"
     ],
-    "benchmark": { "name": "CIS Kubernetes", "version": "1.4.1" },
+    "benchmark": [{ "name": "CIS Kubernetes V1.23", "id": "cis_k8s", "version": "v1.0.0" }],
     "severity": "low"
   },
   "id": "good-csp-rule-template-abc-1",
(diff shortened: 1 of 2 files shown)


mtojek (approved):
It looks like a breaking change in format, but as long as we don't have advanced validation on the `csp_rule_template` format, it shouldn't be breaking any existing package.

versions/1/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/364
+  - description: Refactored benchmark field in kibana/csp-rule-template asset 
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/365
```
jsoriano, line 12: This PR is only changing a sample test package, there is no user-facing change. Do we need a changelog entry for that?
```suggestion
```

test/packages/good/kibana/csp_rule_template/good-csp-rule-template-abc-1.json
```diff
@@ -14,7 +14,7 @@
       "Kubernetes",
       "Containers"
     ],
-    "benchmark": { "name": "CIS Kubernetes", "version": "1.4.1" },
+    "benchmark": [{ "name": "CIS Kubernetes V1.23", "id": "cis_k8s", "version": "v1.0.0" }],
```
jsoriano, line 17: What is the context of this change? Is this template invalid? Should we add some validation to the spec?

kfirpeled (changes requested):
halting the change, we've decided not to go with this approach. 
The original idea was to share metadata of rules between benchmarks but as we recently learned they don't match. Besides that, these rule templates are going to be imported and generated directly from CIS. 
Therefore, it is better not to share metadata between rules.

PR #378, opened 2022-07-25
Add support for icons in dark mode

### LINKED ISSUE #11: Add support for icons in dark mode
Some icons like AWS, Kafka, IBM cloud, Github, etc. will not render well if a dark theme is used. In EUI, there is a mechanism for dealing with this, but since we're technically not using EUI icons (rather, we're defining our own custom ones which happen to be identical), we don't get this magic out of the box.

Since the majority of the icons we're using are already defined in EUI, it might be better if we just reference those icons by name instead, and only define custom icons when we need an icon that is not specified in EUI.

So in our manifest:

```
icons:
- src: /img/logo_postgres.svg
  title: logo postgres
  size: 32x32
  type: image/svg+xml
```

could become something much simpler like

```
icons:
- name: logoPostgres
```

I'm not sure why we need to define a title, size, and type. I think one caveat of referencing by name is that it places a dependency on EUI - meaning any consuming app that wants to see a package icon from the registry will need to have EUI setup.

image

image

mrodm:
## What does this PR do?

This PR adds the required support in package-spec to include icons to be used when dark mode is enabled in Kibana:
- Add a new field under "icons" named `dark_mode` whose type is boolean:
    - If this field is true, then it would indicate that this icon should be shown in dark mode.
    - By default, the value of this field is `false` to keep backwards compatibility.

Currently, I've checked that Kibana shows just the first icon defined in the manifest, when there are two or more icons in the list.

## Why is it important?

If dark mode is enabled in Kibana settings, there are some integrations whose icons cannot be seen correctly (more info and examples in [this issue](https://github.com/elastic/package-spec/issues/11)).

## Checklist

## Related issues

- Closes #11


versions/1/changelog.yml
```diff
@@ -4,6 +4,9 @@
 ##
 - version: 1.14.1-next
   changes:
+  - description: Add support for icons in dark mode
```
mtojek, line 7: Your PR should go to the bottom of the `changes` section.

versions/1/integration/manifest.spec.yml
```diff
@@ -97,6 +97,10 @@ spec:
             type: string
             examples:
               - "image/svg+xml"
+          dark_mode:
```
mtojek, line 100: You may want to add it also to the input type package.

mtojek (commented):
LGTM, but would like to @jen-huang or @kpollich to review it :)

versions/1/integration/manifest.spec.yml
mrodm: It should already be done automatically. Input packages are using directly the same definition as integration for icons, screenshots, vars, etc. From [input spec](https://github.com/elastic/package-spec/blob/main/versions/1/input/manifest.spec.yml#L107): 
```yaml
    icons:
      $ref: "../integration/manifest.spec.yml#/definitions/icons"
    screenshots:
      $ref: "../integration/manifest.spec.yml#/definitions/screenshots"
```

versions/1/changelog.yml
mrodm: 👍  Re-ordered

versions/1/integration/manifest.spec.yml
mtojek: Awesome 👍

kpollich (approved):
Changes LGTM. I created https://github.com/elastic/kibana/issues/137075 to track support for this in Kibana.

mtojek (approved):
Ship it!

PR #380, opened 2022-07-25
Introduce format version v2

### LINKED ISSUE #345: Introduce new package spec version
Introduce package spec format version example `format_version: '1.0'`

Per discussed during one of our team weekly, we will upgrade the package spec version as It is almost impossible to add control over package spec.

Most of the time it breaks a lot of packages and will imply bulk changes a the same time.

Changes packages spec will help us be more strict on validation for new incoming packages, leave older packages as is, add new types and fields without impacting packages without format_version.

Related:
https://github.com/elastic/package-spec/issues/313
https://github.com/elastic/package-spec/pull/338
https://github.com/elastic/package-spec/pull/281

jsoriano:
## What does this PR do?

Initialize a new format version v2 as a copy of current v1 code. Semantic rules are instantiated from a new set of builders, one per major.

If/when this PR is merged, changes should be done in v2, and copied to v1 when non breaking.

## Why is it important?

Allow to introduce breaking changes in a more controlled way.

## Checklist

## Related issues

- Closes #345.


test/packages/good/manifest.yml
```diff
@@ -1,4 +1,4 @@
-format_version: 1.0.4
+format_version: 2.0.0
```
jsoriano, line 1: @elastic/ecosystem what should we do with test packages? I see some options:
1. Leave them as they are, update or add packages only to test new features in new versions.
1. Upgrade all of them to v2, copy `good` to `goodv1` and leave this one as reference for v1.
2. Duplicate all packages, leaving v1 packages to test backwards compatibility.

versions/2/changelog.yml
```diff
@@ -0,0 +1,323 @@
+##
+## This file documents changes in the package specification. It is NOT a package specification file.
+## Newer entries go at the bottom.
+##
+- version: 2.0.0-next
+  changes:
+  - description: Add definition for the license file in a package
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/367
```
jsoriano, line 9: We should probably release all pending changes before merging this.

versions/2/changelog.yml
```diff
@@ -0,0 +1,323 @@
+##
+## This file documents changes in the package specification. It is NOT a package specification file.
+## Newer entries go at the bottom.
+##
+- version: 2.0.0-next
```
jsoriano, line 5: Current approach keeps both majors in the same branch. This implies that releasing a minor for one major would require the release of new versions for all majors with changes. Releasing a new version will require to close all changelogs with changes, and push tags for all new versions mentioned in changelogs.

This makes me wonder a couple of things:
* AFAIK we are not using minors for anything, should we do something about this? Or should we reduce versioning to only majors?
  If we implement minors we could have more control on what is released for each major, but following current approach we would need to duplicate the spec and rules for each minor.
* Should we have one branch per major? This would complicate other things, we would need an entry point for `NewSpec`, that instantiates one spec per major from a different import path.

jsoriano: With reducing versioning I mean to use `format_version: 2` in manifests. We would keep the versioning here in changelogs.

mtojek (commented):
>Initialize a new format version v2 as a copy of current v1 code. Semantic rules are instantiated from a new set of builders, one per major.

I reviewed a similar project, [compose-go](https://github.com/compose-spec/compose-go), to see how they process incoming changes. It looks like the owners manage only one spec file instead of many spec-per-version files. 

I'm wondering if we can try a similar approach, and mark properties/features in spec as `deprecated in X version`, `removed in Y` version.

WDYT?

test/packages/good/manifest.yml
mrodm: I would go also for the option 1, to avoid having more folders/files copied.

I was wondering about the test packages whose goal is to test errors. Should they be updated/copied to test those errors in both formats v1 and v2 ?

jsoriano:
> I reviewed a similar project, [compose-go](https://github.com/compose-spec/compose-go), to see how they process incoming changes. It looks like the owners manage only one spec file instead of many spec-per-version files.
> 

Interestingly they are ignoring [`version`](https://github.com/compose-spec/compose-go/blob/7be7b1184ac5a94c8e26c08029924f77fe4f8aa4/schema/compose-spec.json#L11), does it mean that they follow a strong backwards compatibility path?

Did you find anything like support for multiple versions there?

> I'm wondering if we can try a similar approach, and mark properties/features in spec as `deprecated in X version`, `removed in Y` version.

We would also need something line "added in X version", and not sure how would this play with `additionalContents` depending on the version.

versions/2/changelog.yml
jsoriano: We can completely freeze version 1 when 2 is released, but I would expect a period where version 2 is in development, there are several potentially breaking changes that will need some time to be implemented. At least during this time we are going to need some kind of synchronization between both versions.

mtojek:
I guess that we can review their history :)

https://github.com/compose-spec/compose-spec/issues/12
https://github.com/compose-spec/compose-spec/issues/123

jsoriano:
> I guess that we can review their history :)
> 
> [compose-spec/compose-spec#12](https://github.com/compose-spec/compose-spec/issues/12) [compose-spec/compose-spec#123](https://github.com/compose-spec/compose-spec/issues/123)

After some analysis it seems that compose v2 doesn't use versioning, so they can keep a single spec file.

On https://github.com/elastic/package-spec/pull/384 I follow a mixed approach, also with a single spec, but with rules that can be applied depending on the `format_version`.

jsoriano:
Closing in favor of https://github.com/elastic/package-spec/pull/384.

Issue #322, opened 2022-04-20
Expose a package "loader" functionality


jsoriano:
Package spec currently exposes validation functionality through the `validator` package. Create another package, `loader`, with package loading functionality, to read the contents of packages in a best-effort basis, without requiring strict validation.

Ideally this loader could be mostly auto-generated from the package spec itself. It could access content in a lazy way, so parsing is only done for what is needed for each application. It should be tolerant to failures, so content can be accessed even if the package is partially invalid or incomplete. This loader would abstract handling of different format versions or package types.

Partial package loading functionality is required in several places, each of them is currently maintaining their specific implementation:
* Indexers read packages. Package registry reads packages to index them. Package storage v2 indexer will also need to read packages to generate indexes.
* Kibana reads packages. Exposing information from packages was requested as [feedback for the wasm poc](https://github.com/elastic/package-spec/pull/285#issuecomment-1075363107).
* `elastic-package` reads packages manifests and fields files.
* Documentation exporters read the manifest and documentation files.

mtojek:
Maybe we can go further and generate the model from spec? It will prevent having at least 3 partially implemented models in our codebase.

jsoriano:
Yes, something like this is what I mean with auto-generating from the spec.

Issue #287, opened 2022-03-02
Consider alternatives to jsonschema for validation


jsoriano:
We are using jsonschema for validation, one of the targets was to try to define a spec that can be easily validated in other languages that have support for jsonschema, but we have needed several customizations that required more Go code, making difficult to use the spec in other languages.

Some options we have to solve the issues found can be the following ones, we probably need a mix of them to address all issues:
* Use some other framework for validation of directory structure.
* Use a more extensible `jsonschema` library for validation of json/yaml files.
* Fully replicate the same logic in other languages, and implement a solution to certify the different implementations.
* Expose spec validation functionality through a CLI tool.
* Expose spec validation functionality through some portable target such as WebAssembly ([POC](https://github.com/elastic/package-spec/pull/285)).

Limitations found:
* jsonschema is good for the structure of json/yaml files, but we also try to use something similar for the structure of directories. This validation of directories requires additional custom logic.
* We need some custom validation of fields in yaml files that go beyond the basic json types, and this is tricky to implement with current library (for example we check that relative paths exist and comply with size limits using a custom formatter, but this is not about format, and formatters cannot be parameterized).
* Current library doesn't have good support for virtual filesystems, what produces [issues on Windows](https://github.com/xeipuuv/gojsonschema/issues?q=is%3Aissue+is%3Aopen+windows), https://github.com/elastic/package-spec/pull/358.
* "Semantic" validation requires additional completely custom code.
* We are using jsonschema to validate yaml files, this works fine by now, but there may be tricky cases, and this may limit portability of the spec.
* Custom code is not portable to other languages.

Other things to take into account:
* When reusing the same code in different places, we may need to be less strict on some checks. For example elastic-package can be strict on `additionalContents`, but if Kibana is strict on that, new packages may not work in old versions of Kibana.

jsoriano:
https://cuelang.org/ could be an option if we consider to refactor the spec, thanks @florianl for the suggestion!

jsoriano:
I am going to close this by now. We have invested more in jsonschema since I opened this issue:
* Refactors to more clearly separate the code between yaml/jsonschema, and the spec for directories make it clearer that a single solution wouldn't be enough.
* We have forked the gojsonschema library, we can make changes there as we did to better support `fs.FS`.
* Tried different approaches, like using CUE, but it also had its problems.
* Added support for "patches" to support multiple versions of the spec.

PR #388, opened 2022-08-11
Refactor validator to separate spec loading code

jsoriano:
## What does this PR do?

Refactor validation to split responsibilities between spec loading and package validation.

A new `specschema` internal package is created to contain the functionality to load directories spec. All code related to "yamlschema" has been moved to `yamlschema`.

It is possible now to load a spec without using it for validation, this can allow us to use the spec for other uses as package loading (#322).

## Why is it important?

Current code for validation also contains the code for schema loading. This complicates the test of different approaches for schema definition and loading. It also makes it difficult to separate issues in the schema with issues in the packages.

## Checklist

- [x] ~~I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/main/versions/1/changelog.yml).~~ It doesn't change the spec.

## Related issues

- Relates #287.
- Relates #380.


code/go/internal/jsonschema/folder_spec.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	propagateContentLimits(&s.commonSpec)
+
+	return nil
+}
+
+func (s *folderItemSpec) loadContents(fs fs.FS, specPath string) error {
+	var content *folderItemSpec
+	contents := append([]*folderItemSpec{}, s.Contents...)
```
mrodm, line 60: Where is this Contents field from ? I don't see it in folderItemSpec struct.

code/go/internal/jsonschema/common_spec.go
```diff
@@ -2,7 +2,7 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package validator
+package jsonschema
```
mrodm, line 5: What about naming the package something more generic as "schema" or "specschema" ? I see that it is used gojsonschema library, but the spec files for instance are defined using YAML.

code/go/internal/validator/folder_spec.go
```diff
@@ -7,71 +7,43 @@ package validator
...
-
-	s.fs = fs
-	s.specPath = specPath
-	return nil
+func (v *validator) Validate() ve.ValidationErrors {
+	return v.validateSpec(v.spec, v.pkg, v.folderPath)
 }
 
-func (s *folderSpec) validate(pkg *Package, folderPath string) ve.ValidationErrors {
+func (v *validator) validateSpec(spec spectypes.ItemSpec, pkg *Package, folderPath string) ve.ValidationErrors {
```
mrodm, line 46: A `validator` struct has already spec, pkg and folderPath fields, cannot they be accesed through `v` instead of parameters ?
For instance [here](https://github.com/elastic/package-spec/pull/388/files?diff=split&w=0#diff-dbe281159d5f34e721f41774cf23c7914ebde34855a11a4126aa29f1f5217599R42-R43) are used as parameters the same fields of `v`:
```
func (v *validator) Validate() ve.ValidationErrors {
	return v.validateSpec(v.spec, v.pkg, v.folderPath)
```

code/go/internal/validator/folder_spec.go
```diff
@@ -119,127 +91,94 @@ func (s *folderSpec) validate(pkg *Package, folderPath string) ve.ValidationErro
...
 		}
 	}
 	return errs
 }
 
-func (s *folderSpec) findItemSpec(packageName string, folderItemName string) (*folderItemSpec, error) {
-	for _, itemSpec := range s.Contents {
-		if itemSpec.Name != "" && itemSpec.Name == folderItemName {
-			return &itemSpec, nil
+func findItemSpec(s spectypes.ItemSpec, packageName string, folderItemName string) (spectypes.ItemSpec, error) {
```
mrodm, line 169: Should this function be defined as part of the `validator struct` ? As it is `validateSpec()`, something like:
```
func (v *validator) findItemSpec(s spectypes.ItemSpec, packageName string, folderItemName string) (spectypes.ItemSpec, error) {
```
If so, probably those parameters could be accessed from the validator `v` directly.

code/go/internal/validator/folder_spec.go
jsoriano: Good point, this comes from a previous version where I was reusing the validator for nested directories and calling it recursively with different parameters. I will refactor this for clarity.

code/go/internal/validator/folder_spec.go
jsoriano: Makes sense, yes :+1:

code/go/internal/jsonschema/common_spec.go
jsoriano: Yeah, I called it this way because it uses the jsonschema, but it is right that our way to use it is quite custom, I will rename it.

jsoriano: Ok, I have also split everything related to "yamlschema" to the already existing `yamlschema` package.

Everything related to the load of specs from directories, that is custom of this project, has been moved to `specschema`. Schema loading for files can be customized now, so we could replace "yamlschema" with other thing while keeping the logic for directory specs.

code/go/internal/validator/folder_spec.go
jsoriano: Done.

code/go/internal/validator/folder_spec.go
jsoriano: Done.

jsoriano:
Thanks @mrodm for the suggestions! I think I have addressed them, please take another look.

code/go/internal/specschema/folder_item_spec.go
```diff
@@ -0,0 +1,211 @@
...
+	source := reflect.ValueOf(&o).Elem()
+	for i := 0; i < target.NumField(); i++ {
+		field := target.Field(i)
+		if field.IsZero() {
+			field.Set(source.Field(i))
+		}
+	}
+}
+
+var formatCheckersMutex sync.Mutex
```
mrodm, line 204: A leftover ? I don't see any usage inside specshema package.

mrodm (approved):
Great refactor! :100:

code/go/internal/specschema/folder_item_spec.go
jsoriano: Good catch, removed.

mrodm (approved):
LGTM

PR #389, opened 2022-08-12
Add warning/error for dashboards including visualizations by reference

### LINKED ISSUE #316: Warning when dashboard with visualisations by reference instead of value exist
In Kibana, dashboards can be built with visualisations inside stored by value or by reference. Historically only by reference was supported so most of our dashboards still use by reference. The goal is to move to by value to have dashboards and visualisations directly bundled together as a unit which speeds up the installation of it in Fleet and also makes sure the visualisations do not all show up in the search, only the dashboard.

To help with this effort and notify engineers about usage of by reference, it would be nice if during building a package, package-spec or the build process would show a warning if a visualisation in a dashboard is added by reference instead of by value. Changes should not be blocked on this but engineers notified about it.

This is related to https://github.com/elastic/kibana/issues/129303 which should help engineers with an easy migration to by value.

mrodm:
## What does this PR do?

Add a new validator to check about dashboards using visualizations by reference instead of value. Example output from tests:

```
=== RUN   TestValidateFile/visualizations_by_reference
2022/08/17 13:01:10 Warning: references found in dashboard kibana/dashboard/visualizations_by_reference-82273ffe-6acc-4f2f-bbee-c1004abba63d.json: visualizations_by_reference-5e1a01ff-6f9a-41c1-b7ad-326472db42b6 (visualization), visualizations_by_reference-8287a5d5-1576-4f3a-83c4-444e9058439b (lens)
```

These references are obtained from the dashboard json file:

```json
{
  ...
  "id": "dashboard-id",
  "references": [
    {
      "id": "visualizations_by_reference-5e1a01ff-6f9a-41c1-b7ad-326472db42b6",
      "name": "panel_0",
      "type": "visualization"
    },
    {
      "id": "visualizations_by_reference-8287a5d5-1576-4f3a-83c4-444e9058439b",
      "name": "panel_1",
      "type": "visualization"
    }
  ],
  "type": "dashboard"
}
```

## Why is it important?

This new check wants to highlight (using warning messages) those dashboards using still visualizations by reference.

As mentioned in #316:
> The goal is to move to by value to have dashboards and visualizations directly bundled together as a unit which speeds up the installation of it in Fleet and also makes sure the visualizations do not all show up in the search (visualization library), only the dashboard.

## Checklist

## Related issues

- Closes #316


code/go/internal/validator/semantic/validate_dashboards_used_by_value.go
```diff
@@ -0,0 +1,89 @@
...
+	objectFiles, err := pkgpath.Files(fsys, filePaths)
+	if err != nil {
+		errs = append(errs, errors.Wrap(err, "error finding Kibana Dashboard files"))
+		return errs
+	}
+
+	for _, objectFile := range objectFiles {
+		filePath := objectFile.Path()
+
+		objectReferences, err := objectFile.Values(`$.references[?(@.type=="visualization")]`)
```
mrodm, line 44: is this the right way to check for visualizations included by reference in a dashboard?

code/go/pkg/validator/validator_test.go
```diff
@@ -90,6 +90,7 @@ func TestValidateFile(t *testing.T) {
 				"field services.docker-custom-agent: Must not validate the schema (not)",
 			},
 		},
+		"visualizations_by_reference": {},
```
mrodm, line 93: Adding this new element here, the package will be tested, but no errors can be checked. Should it be moved to another test ?

code/go/internal/validator/semantic/validate_dashboards_used_by_value.go
mrodm: I would want to confirm that this reference key in the dashboard json files is the one defining that a visualization is set as reference.

I've just tested that adding some visualizations from the library in the EPR integration, and it resulted in these changes:

```diff
--- packages/elastic_package_registry/kibana/dashboard/elastic_package_registry-313c2700-099b-11ed-91b6-3b1f9c2b2771.json
+++ packages/elastic_package_registry/kibana/dashboard/elastic_package_registry-313c2700-099b-11ed-91b6-3b1f9c2b2771.json
@@ -2215,6 +2215,38 @@
                 "title": "EPR Indexer - Get duration seconds per indexer - Percentile 95",
                 "type": "lens",
                 "version": "8.0.0"
+            },
+            {
+                "embeddableConfig": {
+                    "enhancements": {}
+                },
+                "gridData": {
+                    "h": 15,
+                    "i": "6beb6552-317f-442e-a6d5-397dc849aa98",
+                    "w": 24,
+                    "x": 0,
+                    "y": 82
+                },
+                "panelIndex": "6beb6552-317f-442e-a6d5-397dc849aa98",
+                "panelRefName": "panel_6beb6552-317f-442e-a6d5-397dc849aa98",
+                "type": "lens",
+                "version": "8.0.0"
+            },
+            {
+                "embeddableConfig": {
+                    "enhancements": {}
+                },
+                "gridData": {
+                    "h": 15,
+                    "i": "a3ae1ef0-3e98-4d2c-9b19-8b2ddc54ef4d",
+                    "w": 24,
+                    "x": 24,
+                    "y": 82
+                },
+                "panelIndex": "a3ae1ef0-3e98-4d2c-9b19-8b2ddc54ef4d",
+                "panelRefName": "panel_a3ae1ef0-3e98-4d2c-9b19-8b2ddc54ef4d",
+                "type": "visualization",
+                "version": "8.0.0"
             }
         ],
         "refreshInterval": {
@@ -2417,6 +2449,16 @@
             "id": "metrics-*",
             "name": "d237dd46-1074-47ee-8186-3957b956acd9:indexpattern-datasource-layer-cfdc23c6-0606-45ea-b1d2-782ba8103f82",
             "type": "index-pattern"
+        },
+        {
+            "id": "elastic_package_registry-b82baae0-1e06-11ed-92f0-1397dcee01a8",
+            "name": "6beb6552-317f-442e-a6d5-397dc849aa98:panel_6beb6552-317f-442e-a6d5-397dc849aa98",
+            "type": "lens"
+        },
+        {
+            "id": "elastic_package_registry-fd226440-1e06-11ed-92f0-1397dcee01a8",
+            "name": "a3ae1ef0-3e98-4d2c-9b19-8b2ddc54ef4d:panel_a3ae1ef0-3e98-4d2c-9b19-8b2ddc54ef4d",
+            "type": "visualization"
         }
     ],
     "type": "dashboard"
```

From this test, I've just realised that it should be checked whether or not there are references from both Lens and Visualization. Should it be checked for other types too ?

code/go/pkg/validator/validator_test.go
jsoriano: Maybe we can have some undocumented environment variable like `PACKAGE_SPEC_WARNINGS_AS_ERRORS` that produce errors on warnings. We could enable it in tests.

code/go/internal/validator/semantic/validate_dashboards_used_by_value.go
```diff
@@ -0,0 +1,89 @@
...
+	var refs []reference
+	jsonbody, err := json.Marshal(val)
+	if err != nil {
+		log.Printf("error encoding reference list: %s", err)
+		return refs, nil
+	}
+
+	err = json.Unmarshal(jsonbody, &refs)
+	if err != nil {
+		log.Printf("error unmarshaling references: %s", err)
```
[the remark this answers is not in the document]
mrodm: These conversion errors are now returned and the callers can handle them properly.

code/go/internal/validator/semantic/validate_dashboards_used_by_value.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, reference := range references {
+		log.Printf("Warning: visualization by reference found: %s", reference.ID)
+	}
+	return true, nil
+
+}
+
+func toReferenceSlice(val interface{}) ([]reference, error) {
+	var refs []reference
+	jsonbody, err := json.Marshal(val)
```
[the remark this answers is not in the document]
mrodm: I've changed the code to use casting/expected types. For the moment, as it would be just one usage, I've tried to not use that `mapstructure` library.

code/go/internal/validator/semantic/validate_dashboards_used_by_value.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, objectFile := range objectFiles {
+		filePath := objectFile.Path()
+
+		objectReferences, err := objectFile.Values(`$.references[?(@.type=="visualization")]`)
+		if err != nil {
+			errs = append(errs, errors.Wrapf(err, "unable to get Kibana Dashboard references in file [%s]", fsys.Path(filePath)))
+			continue
+		}
+
+		anyReference(objectReferences, fsys.Path(filePath))
```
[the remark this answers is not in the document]
mrodm: Updated this function to return references and errors (if any). With this change, `ValidateVisualizationsUsedByValue` function is the one in charge of showing the warnings.

code/go/internal/validator/semantic/validate_visualizations_used_by_value.go
```diff
@@ -0,0 +1,106 @@
...
+		errs = append(errs, errors.Wrap(err, "error finding Kibana Dashboard files"))
+		return errs
+	}
+
+	for _, objectFile := range objectFiles {
+		filePath := objectFile.Path()
+
+		objectReferences, err := objectFile.Values(`$.references`)
+		if err != nil {
+			// no references key in dashboard json
```
mrodm, line 45: There are test packages that do not have "references" key. To avoid errors in case of this key does not exist, this error is ignored.

code/go/internal/validator/semantic/validate_visualizations_used_by_value.go
```diff
@@ -0,0 +1,106 @@
...
+		references, err := anyReference(objectReferences)
+		if err != nil {
+			errs = append(errs, errors.Wrapf(err, "error getting references in file: %s", fsys.Path(filePath)))
+		}
+		if len(references) > 0 {
+			s := fmt.Sprintf("%s (%s)", references[0].ID, references[0].Type)
+			for _, ref := range references[1:] {
+				s = fmt.Sprintf("%s, %s (%s)", s, ref.ID, ref.Type)
+			}
+			log.Printf("Warning: references found in dashboard %s: %s", filePath, s)
```
mrodm, line 58: Example of output here:

```
2022/08/17 13:01:10 Warning: references found in dashboard kibana/dashboard/visualizations_by_reference-82273ffe-6acc-4f2f-bbee-c1004abba63d.json: visualizations_by_reference-5e1a01ff-6f9a-41c1-b7ad-326472db42b6 (visualization), visualizations_by_reference-8287a5d5-1576-4f3a-83c4-444e9058439b (lens)
```

code/go/pkg/validator/validator_test.go
mrodm: Added that environment variable with some helpers to be able to do tests on warnings
https://github.com/mrodm/package-spec/commit/06f27a9a8f7bc212ad4017cbbae45b8d43f6a4e8

code/go/internal/validator/common/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package common
+
+import (
+	"os"
+	"strconv"
+)
+
+// EnvVarWarningsAsErrors is the environment variable name used to use warnings as errors
```
jsoriano, line 12: Please add a comment here mentioning that this mechanism will be removed if/when structured errors are supported.

code/go/internal/validator/spec_test.go
```diff
@@ -5,9 +5,10 @@
 package validator
 
 import (
-	"github.com/elastic/package-spec/code/go/internal/fspath"
 	"testing"
 
+	"github.com/elastic/package-spec/code/go/internal/fspath"
+
```
jsoriano, line 11: Nit. Move package spec imports to the bottom of the import block.

code/go/pkg/validator/validator_test.go
```diff
@@ -28,9 +29,6 @@ func TestValidateFile(t *testing.T) {
 		"deploy_terraform":    {},
 		"time_series":         {},
 		"missing_data_stream": {},
-		"custom_logs":         {},
-		"httpjson_input":      {},
-		"sql_input":           {},
```
jsoriano, line 33: Could we keep these packages here if we are not enabling errors as warnings here?

code/go/pkg/validator/validator_test.go
```diff
@@ -302,6 +300,58 @@ func TestValidateDuplicatedFields(t *testing.T) {
...
+					}
+					return
+				}
+				require.Equal(t, errs.Error(), expectedWarnContains[0])
+			}
+		})
+	}
+	if err := common.DisableWarningsAsErrors(); err != nil {
+		require.NoError(t, err)
+	}
```
jsoriano, line 352: Nit. Run it in a `defer` after enabling this?

code/go/pkg/validator/validator_test.go
mrodm: If I keep those packages here, they are going to be checked twice. 
In the new test suite where warnings are considered as errors too, if there is any error that is raised (and not expected), then the test should fail too as they do no expect any error (`{}`).

For instance, I added a new field `foo` in the manifest in one of them:
```
=== RUN   TestValidateWarnings/custom_logs
    validator_test.go:334: 
        	Error Trace:	validator_test.go:334
        	Error:      	"found 2 validation errors:
        	            	   1. Warning: package with non-stable semantic version and active beta features (enabled in [../../../../test/packages/custom_logs]) can't be released as stable version.
        	            	   2. file "../../../../test/packages/custom_logs/manifest.yml" is invalid: field (root): Additional property foo is not allowed
        	            	" should have 1 item(s), but has 2
        	Test:       	TestValidateWarnings/custom_logs
```
I could keep them in this test suite, in case the new test suite gets removed after structured errors are implemented. WDYT ? @jsoriano

code/go/pkg/validator/validator_test.go
mrodm: Changed using defer

code/go/pkg/validator/validator_test.go
jsoriano: Ok, you are right, it is probably fine as it is. If we add structured errors we probably need to revisit these tests in any case.

jsoriano (dismissed):

code/go/internal/validator/spec_test.go
mrodm: I thought goimports would reorder it and I didn't check that. Updated.
Thanks!

jsoriano (approved):

PR #384, opened 2022-07-28
Introduce initial version 2 and stronger versioning support for the spec

jsoriano:
## What does this PR do?

This PR introduces some changes in how versioning is done in the spec:
* Versioned spec directories are removed.
* `format_version` is strongly checked now, it must exist in the spec `changelog.yml`, prereleases of the version are allowed.
* Multiple prereleases may exist in the changelog file, one for each major. This allows to keep development of multiple majors in a single branch. Some alternatives for this:
  * Keep a separated `changelog.next.yml` file for features only enabled for next major.
  * Have a branch per major, and use backports. 
* Semantic rules can be selected by version now. Validation of unique fields definition (https://github.com/elastic/package-spec/issues/331) has been re-enabled to illustrate that.
* YAML Schema files can now include a `versions` list, with patches to apply to reconstruct older versions of the schema. `license` field is removed from manifests to illustrate that.

## Why is it important?

To allow introducing breaking changes (or loosening validation) in a more controlled way.

## Checklist

- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.
- [x] I have added an entry in [`versions/N/changelog.yml`](https://github.com/elastic/package-spec/blob/main/versions/1/changelog.yml).

## Related issues

- Alternative to #380.
- Closes #331.
- Closes #345.

PR #391, opened 2022-08-15
Use CUE to define schema for files

### LINKED ISSUE #345: Introduce new package spec version

jsoriano:
## What does this PR do?

Uses CUE to define the schema for files instead of JSON schema (only integration manifest and changelog by now).

It includes https://github.com/elastic/package-spec/pull/384 for semantic rules versioning.

## Thoughts on using CUE for schema validation

TLDR; I am not convinced that it provides general advantages.

Pros:
* Expressive and compact syntax.
* Conditional code.
* Nice features like checking backwards compatibility.

Cons:
* No first-level support for descriptions, examples and so on, we would need to work on something else for documentation.
* Only approach for reusability of definitions is the use of packages, but packages loading don't support `fs.FS` (https://github.com/cue-lang/cue/issues/607). Hacks needed.
* Syntax behaves a bit unexpectedly sometimes. Examples:
  * Data injection with tags is quite tricky, and doesn't always work.
  * Additional contents are only checked for named definitions, they are not checked when validating against the spec in the root of a file. This forces us to work with definitions, not such a big issue.
* Internals are hidden in `internal` libraries. No custom Go logic or types can be added, we'd need to fork or continue with semantic validation. Same limitations in that sense than jsonschema at the end.
* Validation of relative paths and data stream names would need to be done with semantic validation (current approach of using custom "formats" in JSON schema is tricky in any case).
* Errors are sometimes quite cryptic, at least in the context of schema validation. Transformations needed.
* Not 1.0 yet, Go APIs may change.

## Why is it important?

To support multiple versions of the spec with the same files.

## Checklist

- [x] Validate data stream names.
- [x] Validate relative paths.

## Related issues

- Alternative to #380.
- Includes #384.
- Related to https://github.com/elastic/package-spec/issues/287.
- Closes #345.


jsoriano:
Closing in favor of https://github.com/elastic/package-spec/pull/384.

PR #392, opened 2022-08-15
Keep track of gotestsum in go.mod

jsoriano:
## What does this PR do?

Add gotestsum to the tools package.

## Why is it important?

So it is maintained as other dependencies using go modules.


mrodm (approved):
:+1:

Issue #400, opened 2022-08-24
Generate consumable documentation from the spec


jsoriano:
Spec files currently contain information about how each file and field should be used, with descriptions, examples, accepted values and so on. But these are source files that may be renamed, moved or removed, breaking documentation builds that depend on them.

Generate some intermediary documentation files in a well-known location (like `docs/spec.asciidoc`) that other projects may refer to.

PR #401, opened 2022-08-24
Recover files used by docs builds

jsoriano:
Workaround to avoid breaking the docs builds till we find a better solution (like #400).

It recovers the files references by docs builds.


mtojek (approved):

karenzone:
Can I get an authorized user to merge, please?

bmorelli25:
It is now safe to revert this PR.

PR #402, opened 2022-08-25
adjust fields name pattern

tetianakravchenko:
Signed-off-by: Tetiana Kravchenko <tetiana.kravchenko@elastic.co>

## What does this PR do?

Add missing character `/` to fields name patter of the fields spec.

we already have some fields with `/` , like k8s `labels.*`, example - https://github.com/elastic/integrations/blob/main/packages/kubernetes/data_stream/state_container/sample_event.json#L36-L45, likely `annotations.*` fields as well. I think it was just never added to `fields.yml` with `/` , as those fields are defined with wildcard:
```
- name: labels.*
  type: object
  object_type: keyword
```
## Checklist

## Related issues

-

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -17,7 +17,7 @@
     link: https://github.com/elastic/package-spec/pull/398
   - description: Adjust field name pattern to support '/'.
     type: enhancement
-    link: https://github.com/elastic/package-spec/issues/331
+    link: https://github.com/elastic/package-spec/issues/402
 - version: 1.16.0
   changes:
   - description: New version validation. Internal changes to start preparing major version 2.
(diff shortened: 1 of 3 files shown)


jsoriano (approved):
LGTM, thanks @tetianakravchenko!

@mtojek do you know of any reason to don't support `/`?

mtojek:
I have no idea, but @andrewkroh provided a semi-explanation - consistency with Beats fields. It looks like we didn't need the slash so far.

jsoriano:
Ok, let's merge this. If we already have labels and annotations with `/` I guess that is not problematic to support these fields.

tetianakravchenko:
any estimations when version with this change will be released and available to use for integration package?

PR #405, opened 2022-09-06
add pipeline benchmarks spec

marc-gr:
## What does this PR do?

Adds the spec for the pipeline benchmarks

## Why is it important?

To be able to define custom pipeline benchmarks configuration

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-package/pull/906


spec/changelog.yml
```diff
@@ -10,6 +10,9 @@
   - description: Remove deprecated `license` field.
     type: enhancement
     link: https://github.com/elastic/package-spec/issues/331
+  - description: Add pipeline benchmarks support
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/
```
mtojek, line 15: nit: PR number is missing

spec/integration/data_stream/_dev/benchmark/pipeline/config.spec.yml
```diff
@@ -0,0 +1,7 @@
+spec:
+  type: object
+  additionalProperties: false
+  properties:
+    num_docs:
```
mtojek, line 5: Should this field be mandatory?

spec/integration/data_stream/_dev/benchmark/pipeline/event.spec.yml
```diff
@@ -0,0 +1,15 @@
+##
+## Describes the specification for a data stream's various field definition files
+##
+spec:
+  # Everything under here follows JSON schema (https://json-schema.org/), written as YAML for readability
+  type: object
+  additionalProperties: false
+  properties:
+    events:
+      description: Given events
```
mtojek, line 10: nit: maybe elaborate about the comment a bit :)

spec/changelog.yml
```diff
@@ -10,6 +10,9 @@
...
 - version: 1.16.1-next
```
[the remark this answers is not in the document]
jsoriano: I think we can release this in 1.x as doesn't seem to be breaking, please move the changelog to the 1.16.1-next section.

spec/integration/data_stream/_dev/benchmark/pipeline/config.spec.yml
marc-gr: It has a default value if not present, so should not break things to have the file empty, apart from being useless. I can make it mandatory if that makes more sense to you. Or maybe we can set a spec constraint of "should have at least one", this way if/when more options are added would be enough to make sure it is not an empty file.

marc-gr: I added the `minProperties` and all tests failed with a message such as 

```
                Error Trace:    validator_test.go:302
                Error:          []string{"could not read root folder spec file: could not load spec for \"integration/data_stream/spec.yml\": could not load spec for \"integration/data_stream/_dev/spec.yml\": could not load spec for \"integration/data_stream/_dev/benchmark/spec.yml\": could not load spec for \"integration/data_stream/_dev/benchmark/pipeline/spec.yml\": could not load schema for \"integration/data_stream/_dev/benchmark/pipeline\": failed to load schema for \"integration/data_stream/_dev/benchmark/pipeline/config.spec.yml\": minProperties must be of an integer"} does not contain "field \"event.dataset\" is defined multiple times for data stream \"wrong\", found in: ../../../../test/packages/bad_duplicated_fields/data_stream/wrong/fields/base-fields.yml, ../../../../test/packages/bad_duplicated_fields/data_stream/wrong/fields/ecs.yml"
```

even the ones without that file, I guess is an issue with the validation or schema generation, I will make it mandatory for now.

jsoriano (approved):

Issue #408, opened 2022-09-07
[Change Proposal] Properly support or remove the array field type


jsoriano:
The package-spec allows the use of the [`array` field type](https://github.com/elastic/package-spec/blob/65942c65ffda8e207c3ab23cb47ccc02b0a43e36/spec/integration/data_stream/fields/fields.spec.yml#L50), but this filed type doesn't exist. [Elasticsearch doesn't have any type for arrays](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/array.html). Instead, any field can contain single or multiple values.

This type seems to come from Beats, where it seems to be some kind of convention when multiple values are expected, but it is also not heavily used there.

On integrations, where it is used, it doesn't generate any mapping, what is probably unexpected.

This type should be properly supported or removed.

Option 1: Removing it
----

We could just remove it on package-spec 2.0.0 (#399), and force package developers to make a decision on these fields when upgrading to this format.
In most of the cases it should be just replaced by the actual type expected there.

Option 2: Add support for it
---

Fleet should need to implement the generation of mappings for these fields. Following Beats implementation I think that we could translate fields with `type: array` to `type: keyword`, and fields with `type: array` and `object_type: something` to `type: something`.

In elastic-package we could add some validation equivalent to the one for array normalization (https://github.com/elastic/elastic-package/pull/765).

jsoriano:
In Elasticsearch adding fields dynamically refers to adding fields that don't have a mapping before, then Elasticsearch tries to guess the type. In the case of arrays, it uses the first value.
With the fields definitions we are not adding fields dynamically, but providing the mappings beforehand, we have to decide on a type there.

I agree though that doing a silent automatic translation from `array` to `keyword` may look unexpected, we could go on the direction of explicitly using both `type: array` and `object_type` for the type of the elements, or with the option of removing support for arrays.

Removing support is probably better, in coherence with Elasticsearch data types.

jsoriano:
I have finally opted by the option of removing support for `array`. I don't see any case that cannot be properly covered with existing field types, and `array` leads to confusion as what to expect from this type.

PR #412, opened 2022-09-12
Remove "array" as field data type

jsoriano:
`array` doesn't correspond to any type available in Elasticsearch. Instead, Elasticsearch supports setting single or multiple values in the same fields. All values must be of the same type.

Current implementations in Beats and Fleet [assume that `object_type` is set on these cases](https://github.com/elastic/kibana/blob/2068b2eeadc3a0b4de0f5a940e446262069d1665/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L286), and the type defined there is the one finally used in the index template. But there is nothing actually checking this, so we have packages using `type: array` without an `object_type`, what doesn't produce a valid mapping.

On the cases where `object_type` could be used, `type` alone can be used to produce the same result, so `type: array` with an `object_type` is redundant and leads to confusing results.

Stop allowing to define fields as arrays, as this doesn't produce what the package developer is probably expecting.

This change will be applied for packages starting on format version 2.0.0.

Closes https://github.com/elastic/package-spec/issues/408.


jsoriano:
Troubleshooting for package developers migrating added added in https://github.com/elastic/elastic-package/pull/974.

mtojek (approved):

Issue #416, opened 2022-09-13
`object_type_mapping_type` is not defined nor documented


endorama:
This is a followup on #415. With `object_type` is usual to see another field, `object_type_mapping_type`.

This field is for [sure actively used in Kibana](https://github.com/elastic/kibana/blob/fd19ec33f7e1ea56117364bf0090a0784079d538/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L174-L205) but is not documented in package-spec. 
There is an [old PR related](https://github.com/elastic/package-spec/pull/314) to missing fields that names it, but the field is yet to be documented. 
It's purpose and its relation with `object_type` is quite unclear.

Any clarification on this would be greatly appreciated!

Issue #415, opened 2022-09-13
`object_type` documentation and values


endorama:
While working on https://github.com/elastic/integrations/pull/4004 a question emerged around `object_type` valid values.

This field is documented in a confusing way in package spec:
https://github.com/elastic/package-spec/blob/c8a0fe0fa2629fcba1dccb753abc67b7c10d03ef/spec/integration/data_stream/fields/fields.spec.yml#L235-L238

After https://github.com/elastic/package-spec/pull/412, that removed `type: array`, I think the description is misleading and leaves a lot of questions unanswered around when and how to use `object_type`.

Any clarification on this would be greatly appreciated!

PR #417, opened 2022-09-14
Document and validate object_type

jsoriano:
Document and validate the accepted values for `object_type` and `object_type_mapping_type`.

Adding the list of accepted values could be a breaking change, but this is not heavily used, and when used, it is only with a reduced subset of the values accepted in this PR. I could apply this only for 2.0.0 if this is a concern.

Implemented in Kibana in https://github.com/elastic/kibana/pull/137772.

Reference to the use in the `array` type removed because of https://github.com/elastic/package-spec/pull/412.

Fixes https://github.com/elastic/package-spec/issues/415
Fixes https://github.com/elastic/package-spec/issues/416

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -18,6 +18,9 @@
   - description: Add map as another reference type to check to show warnings
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/414
+  - description: Document and validate `object_type` and `object_type_mapping_type`.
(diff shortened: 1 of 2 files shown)


mrodm (approved):
LGTM

As you mentioned, as the current packages are already using a subset of the values defined here, I think we can release it before 2.0.0

Issue #318, opened 2022-04-12
 Warning when assets are built with a stack SNAPSHOT version


ruflin:
We recently had some issues with some dashboards and visualisations based off a snapshot version of the stack. Snapshot builds can contain bugs which can lead to invalid dashboards / visualisations exported.

The general recommendation is to build dashboards and visualisations off stable versions of the stack. But this is not always possible as a feature is needed that is only released in the next minor but I expect this to be the exception.

The saved objects normally contain some information related to from which version of the stack these were exported from (https://github.com/elastic/package-storage/blob/production/packages/apache/1.3.5/kibana/dashboard/apache-Logs-Apache-Dashboard.json#L15). If this info is incorrect, we should work with the Kibana team to have this more standardised (@flash1293 ). If this version contains a `-SNAPSHOT` string, a warning should be raised.

Dashboards and visualisations for a package should always be built with the minimal version of the stack that it is compatible with. So if a package is compatible with 8.0, this is the version that dashboards should be built with. This is an additional point that can be validated that assets are not newer then the minimal supported stack version.

jsoriano:
Something we could do is to disallow having resources built with `-SNAPSHOT` in stable versions of packages. Though this may complicate development of packages aligned with stack releases.

ruflin:
An idea brought up today by @joshdover related to a different issue was around using label. This could work here as well. The logic I could see happening:

* Prerelease package versions are allowed to contain -SNAPSHOT
* Stable versions are not allowed to contain -SNAPSHOT
* Stable versions can contain -SNAPSHOT if `allow_snapshot` label is added

This will allow development of prerelease packages based on prelease release. If it should be released with the iteration, the dashboards should be exported again after FF. The same is true for bundled packages. One question around this: What happens if you run export dashboard on a RC, will it contain -SNAPSHOT?

mrodm:
I have some doubts about which field should it be considered to get the Elastic stack version that was used to build the assets from. Maybe @flash1293 , could you help me with this ? or redirect me to some doc about it ?

- What is the difference between `coreMigrationVersion` and `migrationVersion`?
    - I've checked that those fields could have different versions, Should we focus on `migrationVersion` field to get the version?
    - Example:
        - [Different versions](https://github.com/elastic/integrations/blob/main/packages/zscaler_zpa/kibana/visualization/zscaler_zpa-552331e0-4c4f-11ec-9023-a76a2cb41dcd.json#L76-L80)
[...]

PR #418, opened 2022-09-14
Add validation rule to show warnings in case assets are built from SNAPSHOT versions of Elastic stack

mrodm:
## What does this PR do?

Add a validation rule whose aim is to show warnings in case the Kibana assets in a package are built from a Elastic stack SNAPSHOT version (e.g. 8.5.0-SNAPSHOT) instead of a stable version.

It also adds a new optional field into the manifest of the package, to indicate whether or not it is allowed to export assets using Elastic stack SNAPSHOT versions.

```diff
 policy_templates:
       - type: prometheus/metrics
         title: Example metrics
         description: Collect metrics via Prometheus exporters"
+allow_snapshots: true
 owner:
   github: elastic/ecosystem

```

Example of the warning message:
```
=== RUN   TestValidateFile/bad_assets_elastic_snapshot_versions
2022/09/20 11:52:23 Warning: snapshot version found in dashboard kibana/dashboard/bad_assets_elastic_snapshot_versions-overview.json: 8.4.1-SNAPSHOT
2022/09/20 11:52:23 Warning: snapshot version found in visualization kibana/visualization/bad_assets_elastic_snapshot_versions-visualization1.json: 8.4.1-SNAPSHOT
```

## Why is it important?

These warnings will enforce to create/export Kibana assets using stable Elastic stack versions. Allowing just to export those assets from a Elastic SNAPSHOT version in some special scenarios (e.g. technical previews or prerelease versions of the package).

## Checklist

    - Updated good with an asset using `migrationVersion`
    - Added test packages to show 3 scenarios
        - Package with warnings (containing assets built from SNAPSHOT versions)
        - Package with a prerelease version that contains assets built from SNAPSHOT versions
        - Package with a `allow_snapshot` set

## Related issues

- Relates #318


test/packages/assets_elastic_snapshot_versions_prerelease/manifest.yml
```diff
@@ -0,0 +1,34 @@
+format_version: 1.0.0
+name: assets_elastic_snapshot_versions_prerelease
+title: "Asse1s Elastic snapshot versions in Prerelease package"
+version: 0.0.1
```
mrodm, line 4: Package whose version is technical preview (0.0.1). This package is allowed to contain assets built from Elastic stack SNAPSHOT versions

test/packages/assets_elastic_versions_allowed_snapshot/manifest.yml
```diff
@@ -0,0 +1,35 @@
...
+    type: image/svg+xml
+policy_templates:
+  - name: sample
+    title: Sample logs
+    description: Collect sample logs
+    inputs:
+      - type: logfile
+        title: Collect sample logs from instances
+        description: Collecting sample logs
+allow_snapshot: true
```
mrodm, line 33: Package that has set `allow_snapshot` as true. This package is allowed to contain assets built from Elastic stack SNAPSHOT versions

spec/integration/manifest.spec.yml
```diff
@@ -309,6 +309,10 @@ spec:
       $ref: "./data_stream/manifest.spec.yml#/definitions/vars"
     owner:
       $ref: "#/definitions/owner"
+    allow_snapshot:
+      type: boolean
+      description: Can assets be exported from a SNAPSHOT version?
+      default: false
```
mrodm, line 315: In https://github.com/elastic/package-spec/issues/318#issuecomment-1096811751, it is mentioned that if it exists a label, package could contain assets built from Elastic stack SNAPSHOT versions.

My first approach to achieve this is adding a new field `allow_snapshot` into the spec of the root package. This field must be added explicitly in case of stable package versions if they need to use assets built from Elastic stack SNAPSHOT versions. WDYT ? Any other suggestion for this ? @elastic/ecosystem

mtojek: I'm not convinced about the approach as the manifest describes the package content, but `allow_snapshot` is related to the procedure of building/exporting assets.

> These warnings will enforce to create/export Kibana assets using stable Elastic stack versions

Shouldn't we warn/block the `export` command if this isn't a stable stack? We could detect the Kibana/ES version instead. I just feel that we don't really don't need a new manifest field :)

mrodm: Agreed, it does not feel the place to add a field for that purpose. This first approach was to try to develop everything under the package-spec umbrella.

I'll check the export command you suggested for options.
Maybe it can be done something like:
- in package-spec, the validation rule shows warning messages if assets are built from Elastic Stack
    - Exceptions: technical preview (0.x.x) and prerelease packages that could have those kind of assets.
- in elastic-package:
    - export dashboard fails or shows a warning if it is from an SNAPSHOT version
    - add a parameter to export command `-allow-snapshot`, so developers need to add it explicitly if they need those dashboards:
      ```
      elastic-package export dashboards -allow-snapshot
      ``` 
        - warning messages defined by package-spec would be shown.

WDYT about this other approach?

mtojek: I mean, from the package-spec perspective it shouldn't be bad to use snapshots, so we shouldn't warn about it? I would see this feature rather scoped within the `export` command with a switch (as you suggested), but wouldn't leak the logic to spec.

Feel free to discuss the idea also with Jaime as I might be biased here (used to work with spec since it was born.)

jsoriano: Yeah, I agree that the versions used in the dashboards are not so related to a package specification, it is not that these versions are _wrong_.

I like the option of making `elastic-package export dashboards` to fail unless `-allow-snapshot` is used. This is a way to make the developer decide if they know what they are doing. We can also start with this option, and add more warnings in the future if this is not enough.

Another option could be to add the warning in `elastic-package`, as `elastic-package` can be more opinionated towards best-practices, while the spec is more focused on producing valid packages.

But summarizing, I think we can start by adding the `-allow-snapshot` flag to `elastic-package export dashboards`, and make it fail, with a clear error message, if snapshots are found when this flag is not used.

jsoriano: Btw, food for thought, the main concern with the use of dashboards generated by snapshot versions is that the developer may be using features that are not released at the end, or are in a different way. If a dashboard breaks in a stable version because of this, it could be detected by tests on dashboards, that would be nice to have at some point.

mrodm: Ok! It's true that dashboards are not wrong even if they are being exported from a SNAPSHOT version in the scope of the spec. As you both suggested, this PR can be closed in favor of https://github.com/elastic/elastic-package/pull/986 where I'm adding that logic in the export command.

Thanks @mtojek and @jsoriano for your comments here!!

mrodm: > If a dashboard breaks in a stable version because of this, it could be detected by tests on dashboards, that would be nice to have at some point.

If CI could detect those scenarios, that would be even better

Issue #340, opened 2022-05-20
[Change Proposal] Add support for synthetic source


jsoriano:
Elasticsearch is going to introduce a new feature called [synthetic source](https://github.com/elastic/elasticsearch/pull/85649). It allows to ingest documents without keeping the source, and rebuild it if needed from the indexed data. Not storing the source produces significative disk space savings.

We may decide in the future to enable synthetic source everywhere, but in the meantime we should have a way to enable it per data stream. 

Slightly Blocked by: https://github.com/elastic/kibana/issues/132818

ruflin:
++ having this available in the package spec. Ideally, it would also be possible to turn it on / off in the Fleet UI somewhere.

joshdover:
This is a 3rd use case for expanding our proposal in https://github.com/elastic/kibana/issues/132818 to allow enabling this on an integration or data stream level for testing before we enable this in packages by default.

andresrc:
My preferred approach as of today would be:

## Initial Phase - enabling integrations

- We would have a boolean configuration option at the [data stream level](https://github.com/elastic/package-spec/tree/main/versions/1/integration/data_stream) in order to enable synthetic source.
  - The default value would initially be `false` so that we keep the current behaviour by default.
  - If set to `true` Fleet would add the appropriate info to the mapping (and roll over if needed).
  - We would need to know if previous versions of Fleet would just ignore the option, to understand if we need to bump `kibana.version` when adding this option or not.
- Integrations would be able to gradually opt-in just by releasing a new version of the packages

## Future phase

- We might have options in Fleet to dynamically change this setting. Granularity would need to be defined.
- The default might change at some point.

jsoriano:
@andresrc the current idea we had as of https://github.com/elastic/kibana/issues/132818 is the following:
* Nothing is added to the spec by now.
* The advanced opt-in flag is added in Fleet, so users (and developers) can optionally enable synthetic source on any data stream.
* Depending on feedback/results:
  * If no problems or limitations are found, we remove the opt-in flag, and we enable synthetic source everywhere.
  * If it depends on the use case, we add the flag to the spec, with a default to be determined.

This is functionally similar to what you are suggesting, but we wouldn't need to add a new kind of configuration, and this feature would be available in all data streams, without needing changes on integrations.

andresrc:
Thanks @jsoriano , that approach looks better, yes. 

[...]

PR #419, opened 2022-09-15
Add setting to configure the source mode of a data stream

jsoriano:
## What does this PR do?

Add setting to configure the source mode of a data stream.

## Why is it important?

Depending on the use case, deciding how to store the source may bring important disk space savings or perfomance improvements.

This allows to enable [`_synthetic` source](https://www.elastic.co/guide/en/elasticsearch/reference/8.4/mapping-source-field.html#synthetic-source) for a data stream.

We may want to decide on a general default for the source mode depending on the results of the technical review implemented in https://github.com/elastic/kibana/issues/140095, but we will still need a toggle in the package for exceptions (https://github.com/elastic/package-spec/issues/340#issuecomment-1228368172).

## Checklist

## Related issues

- Related to https://github.com/elastic/kibana/issues/132818.
- Related to https://github.com/elastic/kibana/issues/140095.
- Closes https://github.com/elastic/package-spec/issues/340.


jsoriano:
@kpollich depending on the outcome of this PR we may need a follow up on https://github.com/elastic/kibana/issues/140095.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -222,5 +222,19 @@ spec:
               type: array
               items:
                 type: string
+        source_mode:
+          description: |-
+            Source mode to use. This configures how the document source (`_source`) is stored
+            for this data stream.
+            If configured as `default`, this mode is not configured and it uses Elasticsearch defaults.
```
ruflin, line 229: It would be nice to add a link to https://www.elastic.co/guide/en/elasticsearch/reference/8.4/mapping-source-field.html#synthetic-source here so anyone trying to figure out what this is can directly read up on it.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -222,5 +222,19 @@ spec:
...
+            If configured as `synthetic`, it doesn't store the source, but tries to rebuild it
+            from the indexed fields when queried.
+            If no configured, users may override the setting from Fleet UI.
```
ruflin, line 232: If it is configured, users can't override it?

ruflin:
Overall LGTM, left 2 minor questions.

spec/integration/data_stream/manifest.spec.yml
jsoriano: I guess that if a package developer explicitly uses `source_mode: default` is because they know that the data stream doesn't work with synthetic mode, so it shouldn't be possible to override it.

For synthetic mode it may make sense to override it, as a escape hatch in case there is some unexpected problem with the data or the integration.

I can change to this if you agree.
```suggestion
            If no configured or set to `synthetic`, users may override the setting from Fleet UI.
```

spec/integration/data_stream/manifest.spec.yml
jsoriano: Added :+1:

spec/integration/data_stream/manifest.spec.yml
ruflin: Agree on the change. If we later find out that even when specifically set to default there is a reason to change, it would be an extension and not a breaking change.

mrodm (approved):
LGTM

kpollich (dismissed):
LGTM 🚀 

I created https://github.com/elastic/kibana/issues/141211 to capture Fleet support for this.

jsoriano:
> I created [elastic/kibana#141211](https://github.com/elastic/kibana/issues/141211) to capture Fleet support for this.

Thanks!

Issue #399, opened 2022-08-24
Checklist for version 2.0.0


jsoriano:
List of pending changes that we want to have in 2.0.0 before releasing it:
- [x] https://github.com/elastic/elastic-package/issues/615
  - [x] https://github.com/elastic/elastic-package/pull/765
- [x] https://github.com/elastic/elastic-package/issues/837
  - [x] https://github.com/elastic/elastic-package/pull/961
- [x] https://github.com/elastic/package-spec/issues/217
  - [x] https://github.com/elastic/elastic-package/pull/968
- [x] https://github.com/elastic/package-spec/issues/408
  - [x] https://github.com/elastic/package-spec/pull/412
- [x] https://github.com/elastic/package-spec/issues/92
- [x] https://github.com/elastic/package-spec/issues/282
  - [x] https://github.com/elastic/package-spec/pull/420

Postponed
- [ ] https://github.com/elastic/elastic-package/issues/147
       It should be built upon https://github.com/elastic/package-spec/issues/199, and it won't be so breaking in that case.
- [ ] https://github.com/elastic/package-spec/issues/57
       There are also checks in Kibana, if some package is broken now it cannot be used. We can add validations here in a next version.

Related:
- [x] https://github.com/elastic/elastic-package/issues/963
  - [x] https://github.com/elastic/elastic-package/pull/964 
  - [x] https://github.com/elastic/elastic-package/pull/974

jsoriano:
Checklist completed and version 2 announced, closing this.

PR #420, opened 2022-09-15
Stricter validation of field definitions

jsoriano:
## What does this PR do?

Stricter validation of field definitions.

## Why is it important?

To reduce the risk of typos in field definitions that can lead to unexpected mappings.

## Checklist

- [x] Test these changes with current integrations.

## Related issues

- Follow up of https://github.com/elastic/package-spec/pull/314.
- Part of https://github.com/elastic/package-spec/issues/399.
- Migration docs added in https://github.com/elastic/elastic-package/pull/978.
- Closes https://github.com/elastic/package-spec/issues/282.


jsoriano:
Tested with integrations and some adjustments done. I think that the number of changes required on existing integrations now are reasonable.

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -6,7 +6,17 @@ spec:
   type: array
   items:
     type: object
-    additionalProperties: true
+    additionalProperties: false
+    patternProperties:
+      # Soft validation on some properties that can be present but not validated or used yet.
```
jsoriano, line 11: The main reason I am leaving it here is to reduce the number of changes that package developers should make to migrate to v2, and to avoid conflicts with fields copied from ECS. But these fields are not really used.

mrodm: Ok, if it is not required too many changed done by developer to migrate, it will be easier that packages get updated to 2.0.0 :+1: 
Those could be removed in a following major if it is required.

mrodm (approved):
LGTM!

PR #423, opened 2022-09-20
Fix go import paths for v2

jsoriano:
## What does this PR do?

Updates go code to use `github.com/elastic/package-spec/v2` as import path, and releases `v2.0.0-rc2` including the changes that were planned for 1.18.1.

## Why is it important?

Package Spec is imported in several go projects, and go has strict rules for versioning and import paths.


mrodm (approved):
LGTM

PR #428, opened 2022-09-28
Run sematic validation also when schema validation fails

jsoriano:
## What does this PR do?

It always run sematic validation, even when schema validation fails.

## Why is it important?

It gives earlier feedback to package developers, as they can see semantic validation failures also before solving other schema issues.

## Checklist


efd6 (approved):
Thank you

mrodm (approved):

Issue #437, opened 2022-10-13
[Change Proposal] Support package specific index mappings for file upload feature


juliaElastic:
There are requirements for `endpoint` and `elastic-agent` packages to support file uploads. These files would live in Elasticsearch, in an index/data stream specific to the package (separate index/ds for metadata and blob).

The indices should be created at package installation time, including an ILM policy that ensures that uploaded files are deleted periodically (e.g. after 30 days).
The files would be read by Kibana Files Service, called by Fleet and Security plugins.

Based on the discussion [here](https://github.com/elastic/security-team/issues/4661#issuecomment-1276321835) it is not yet decided whether an index or data stream would be used.
It seems that package spec doesn't support indices yet, however endpoint package already uses indices "out-of-spec".

Related issues:

- https://github.com/elastic/security-team/issues/4661
- https://github.com/elastic/ingest-dev/issues/1224

cc @jsoriano @joshdover @pzl @kevinlog

jsoriano:
> Proposed index pattern name: `fleet-files-*`. E.g. for agent: `fleet-files-elastic-agent` (meta) and `fleet-files-elastic-agent-data` (binary)

So packages would only need to specify the schema for the documents in the meta index, right?
I guess we can support this with something like a new `file_upload` directory that can contain field definitions.

> Based on the discussion [here](https://github.com/elastic/security-team/issues/4661#issuecomment-1276321835) it is not yet decided whether an index or data stream would be used.
> It seems that package spec doesn't support indices yet, however endpoint package already uses indices "out-of-spec".

I guess this is something to decide on install time. The package provides the mapping, Fleet decides if it installs it as a data stream or an index.

If an index is used, I guess that this is because files or their metadata can be updated or removed. In that case, do we also want ILM?

kevinlog:
@juliaElastic thanks for creating this issue.

@jsoriano 

> So packages would only need to specify the schema for the documents in the meta index, right?
I guess we can support this with something like a new `file_upload` directory that can contain field definitions.

Interesting, so we could define a directory structure like this in package spec, keeping with the `fleet-files-elastic-agent` example:

`package/fleet/file_upload/elastic_agent/fields/fields.yml`

This contains the mappings for the index or data stream.

[...]

PR #439, opened 2022-10-18
[Security Solution][Fleet] File upload test pacakge

kevinlog:
## What does this PR do?

This PR adds a test package to facilitate discussion on proposed package spec changes referenced in this ticket: https://github.com/elastic/package-spec/issues/437

Our proposal is to add a new `file_upload` folder which will contain mappings and ilm policies for an upcoming feature where we will store the file chunks and file metadata of requested files from remote Agents.

## Why is it important?

This is important to support features in both Fleet and Security.

In Fleet, we want to request Agent diagnostics files from remote hosts for use in debugging.  They will be uploaded and stored in the indices specified in this test package and eventual package-spec change

In Security, we want to support a feature where users can request a file from a remote host for further inspection in the event of an attack.

## Checklist

## Related issues

Related to: https://github.com/elastic/package-spec/issues/437

-


test/packages/file_upload_example/file_upload/files/elasticsearch/ilm/files.json
```diff
@@ -0,0 +1,15 @@
...
+                "actions": {
+                    "rollover": {
+                        "max_size": "10gb",
+                        "max_age": "1d"
+                    }
+                }
+            }
+        }
+    }
+}
```
jsoriano, line 15: Do we expect ILM policies for this feature to be different between packages?

test/packages/file_upload_example/file_upload/files/fields/fields.yml
```diff
@@ -0,0 +1,12 @@
...
+  type: binary
+  description: file chunk
+- name: bid
+  type: keyword
+  description: bid
+  index: false
+- name: last
+  type: boolean
+  description: is the last file chunk
+  index: false
```
jsoriano, line 11: How will this mapping be different in other packages? Would you have an example?

test/packages/file_upload_example/file_upload/files_data/fields/fields.yml
```diff
@@ -0,0 +1,180 @@
...
+    - name: id
+      type: keyword
+      description: event id
+- name: host
+  title: Host
+  type: group
+  fields:
+    - name: hostname
+      type: keyword
+      description: hostname
```
jsoriano, line 180: These mappings look to me like quite generic files metadata. How will these mappings be different in other packages? Do you have an example?

test/packages/file_upload_example/file_upload/files_data/fields/fields.yml
```diff
@@ -0,0 +1,180 @@
...
+  fields:
+    - name: Status
+      type: keyword
+      description: status
+    - name: ChunkSize
+      type: integer
+      description: chunk size
+    - name: Compression
+      type: keyword
+      description: compression type
```
jsoriano, line 22: Nit. Different casing used here an in other parts of the mapping.
```suggestion
    - name: status
      type: keyword
      description: status
    - name: chunk_size
      type: integer
      description: chunk size
    - name: compression
      type: keyword
      description: compression type
```

jsoriano (commented):
Thanks for uploading an example! I think it will be helpful to continue with the discussion.

I still wonder why we need so much information in packages for this feature. It looks to me that the indexes and ILM policies added could be basically the same for any package.
If that is the case, then I don't think that packages should contain these configurations.

I think we need to simplify this, as it is now it is going to be quite dependent of each package, and Fleet is going to need to support a too big variation of mappings and policies.

We could have a generic mapping and ILM policies installed by Fleet when this feature is used, then packages wouldn't need to have anything, and Fleet would only need to support this single configuration for any package that supports this feature.

If this feature should be available only in some packages, then we could have a boolean parameter in the manifest to enable or disable it.

test/packages/file_upload_example/file_upload/files/elasticsearch/ilm/files.json
kevinlog: I could only see ILM policies being different if Fleet and Security want to delete docs at different rates.

@pzl @kpollich @joshdover @juliaElastic - do you see any reason for us to have different ILM policies?  On the Endpoint side, we were considering fairly quick cleanup, perhaps removing files that are more than 1-3 days old.  Is that OK with Fleet?

test/packages/file_upload_example/file_upload/files/fields/fields.yml
kevinlog: I don't think this would be different between packages.  Each of these docs represents a file chunk which wouldn't have variation.

@pzl let me know if that sounds correct.

cc @kpollich @juliaElastic @joshdover

pzl: this is the minimum (and mandatory) set of chunk fields. Endpoint likely doesn't need any more. I don't imagine fleet would either. But I haven't thought of any creative use cases outside of that

test/packages/file_upload_example/file_upload/files_data/fields/fields.yml
pzl: to follow ECS patterns: lowercase fields represent ECS standard.

ECS is explicitly lowercase only and encourages initial capitals to indicate custom extensions (i.e. fields that are not ECS)

kevinlog: @pzl I was basing the above on your working set of mappings/indices.  Let me know if there's any issue with the naming above.

kevinlog:
@jsoriano thank you for taking a look at this, your comments are very helpful!

> We could have a generic mapping and ILM policies installed by Fleet when this feature is used, then packages wouldn't need to have anything, and Fleet would only need to support this single configuration for any package that supports this feature.

> If this feature should be available only in some packages, then we could have a boolean parameter in the manifest to enable or disable it.

If I'm understanding correctly, then we could greatly simplify the package-spec change to either enable or disable the `file_upload` feature which would simply inform Fleet they they should install the relevant indices, mappings, ilm policies for the given package.  Further, this would work best if Fleet and Endpoint used the same mappings and ilm policies.

This seems reasonable to me, but would need comment from the other stakeholders.  @pzl @joshdover @kpollich @juliaElastic - I think this implies that we would store the generic mappings and ilm policy information somewhere in Fleet/Kibana as opposed to looking to packages to provide that info.

jsoriano:
> I think this implies that we would store the generic mappings and ilm policy information somewhere in Fleet/Kibana as opposed to looking to packages to provide that info.

Exactly, Fleet would have these configurations, and it would manage their installation on packages using this feature.

test/packages/file_upload_example/file_upload/files_data/fields/fields.yml
pzl: There is a mandatory minimum set of fields for the functionality to work.

I believe the set is currently (but not finalized):

- `file.Size`
- `file.Status`
- `file.ChunkSize`
- `file.Name`


Any integration (endpoint, agent, or future users) should be free to map any additional properties that might aid in:

- displaying the content (e.g. mime type, file path)
- filtering and discovery (e.g. event.id, owner, permissions)
- investigation and information (e.g. inode, owner, create time, access times)

Importantly, an integration is responsible for its own discovery and filtering of files as desired. This index mapping directly feeds whatever is available for search and filtering. What Endpoint may want to index for searching could be different from Agent. And I can see Endpoint's mapping expanding as time goes on, as security information about a quarantined file. It is unlikely these expanded uses would be _as_ useful to Agent who are uploading diagnostic logs for the time being. It is even less likely those additional fields need to be explicitly indexed rather than just _present in source doc_, if sent at all.


Other integrations, if this feature does end up useful in the future to others, will have their own needs to filtering and file discovery and this mapping would control that. I would posit that it needs to be flexible (with the few feature-required minimums)

pzl:
I left a comment above that somewhat addresses the fleet-installs-the-mappings. Or at least explains that there can and may be differences in the future. There actually probably _are_ differences *now*, agent is unlikely to use the same mappings for Endpoint out of the gate here. The maximalist mappings given are for Endpoint's usecase. Agent probably wont be sending `file.inode` and other deep-level filesystem info.

For the mappings to differ, they probably shouldn't live in fleet.

Additionally, as Endpoint changes and adds new fields, and needs the mapping to change, the integration/package mapping will map best to the Endpoint version sending new fields.

If these were controlled by fleet then the calculus between endpoint version and what mapping to use becomes much worse

juliaElastic:
> This seems reasonable to me, but would need comment from the other stakeholders. @pzl @joshdover @kpollich @juliaElastic - I think this implies that we would store the generic mappings and ilm policy information somewhere in Fleet/Kibana as opposed to looking to packages to provide that info.

I think this makes sense if we don't anticipate a different mapping or ILM policy needed for endpoint and agent.
So that means we would store the mappings in kibana as code and simply install it at package installation time?

test/packages/file_upload_example/file_upload/files/elasticsearch/ilm/files.json
kpollich: I think it'd be fine to globally apply the same ILM policy for all file upload indices - Fleet doesn't have any unique constraints here that I'm aware of.

kpollich:
> We could have a generic mapping and ILM policies installed by Fleet when this feature is used, then packages wouldn't need to have anything, and Fleet would only need to support this single configuration for any package that supports this feature.
> 
> If this feature should be available only in some packages, then we could have a boolean parameter in the manifest to enable or disable it.

I agree that this would simplify things greatly, but @pzl's comments above about Endpoint's more specific use cases for file metadata related to their own uploading use cases (quarantined files, etc) versus agent's use case (more generic "upload a somewhat arbitrary .zip archive") raises concerns about this approach. It wouldn't make sense to "hardcode" mappings, ILM policies, etc for the file upload indices in Fleet when Endpoint will want to update mappings between package releases. 

We could define a minimal set of mappings in Fleet, and then allow Endpoint to extend these base mappings via its package manifest, but I worry about multiple sources of truth here and resolving collision between the mappings. It'd probably be better to have the mappings, etc live entirely in one location rather than having competing sources of truth here.

jsoriano:
> We could define a minimal set of mappings in Fleet, and then allow Endpoint to extend these base mappings via its package manifest, but I worry about multiple sources of truth here and resolving collision between the mappings. It'd probably be better to have the mappings, etc live entirely in one location rather than having competing sources of truth here.

I think this could be a good option, this way each package would only need what is different for them.

Another option is to provide conventions based on dynamic mappings, or the `flattened` type, so each package can store any metadata they want using specific field name patterns. Then each package wouldn't need to declare anything.

Take into account that the difference in complexity between including the mappings on each package vs. having predefined configurations in Fleet is quite big. Also the complexity needed to leverage this feature in other packages.

kpollich:
If we can include dynamic mappings in the hardcoded file metadata mappings maintained by Fleet, would that be suitable for Endpoint's needs? Or are more specific mappings necessary to drive the UI around these uploaded files @pzl?

pzl:
I guess enabling dynamic mapping could work here.

I'm not sure what the mechanism will be for fleet to know to install the needed indices for a given package/integration. A setting somewhere in the package manifest for `file_upload: true` and that makes fleet create `.fleet-{integrationName}-files` and `.fleet-{integrationName}-file-data`?

kpollich:
> I'm not sure what the mechanism will be for fleet to know to install the needed indices for a given package/integration. A setting somewhere in the package manifest for file_upload: true and that makes fleet create .fleet-{integrationName}-files and .fleet-{integrationName}-file-data?

This sounds acceptable to me, and pushing a single boolean flag through the package spec change process will be exponentially faster than aligning on a new directory structure, etc. I'm +1 on this approach:

1. Packages can define `enable_file_uploads` as a boolean value in their top-level `manifest.yml` file
2. Fleet will detect this field during package installation
3. If `enable_file_uploads` is `true`, Fleet will `PUT` two indices during package installation: `.fleet-{integrationName}-files` and `.fleet-{integrationName}-file-data` with minimal mappings + `dynamic: true` for mappings
4. Fleet will also define an ILM policy for these indices with sane defaults for expiring uploads

There are probably a few more implementation details on Fleet's end (what to do when the value of `enable_file_uploads` changes between releases, for example), but we can capture those in an implementation issue on the Fleet side.

pzl:
in agreement with above.

@jsoriano how does this approach strike you?

jsoriano:
This approach looks good to me, thanks!

> 3\. If `enable_file_uploads` is `true`, Fleet will `PUT` two indices during package installation: `.fleet-{integrationName}-files` and `.fleet-{integrationName}-file-data` with minimal mappings + `dynamic: true` for mappings

Question: Are these indexes always needed? or only if/when the user uses some specific feature from Fleet? If they are only needed when a user uses some feature maybe these indexes could be installed at this point.

kpollich:
> Question: Are these indexes always needed? or only if/when the user uses some specific feature from Fleet? If they are only needed when a user uses some feature maybe these indexes could be installed at this point.

We could potentially defer installation lazily, but there aren't really major gains to approaching it this way. It'll be more maintainable to add logic to Fleet's existing installation process (we have consistent patterns for adding new steps here) as opposed to deferring to the first time a user takes some action (e.g. upload agent diagnostics or upload quarantined file via Endpoint).

joshdover:
Couldn't this be solved by installing an index template rather than
creating a concrete index? I believe this is needed for ILM to work
anyways. In that case, the index won't be created until a document is
ingested on a matching path via auto_create.

kpollich:
@joshdover You are correct - we should be creating an index template here. 

Could we simply create catch-all index templates for all file uploads and file upload metadata during Fleet installation and avoid a package spec change altogether?

kpollich:
> One question I have is would there be any downside to installing the index templates without the installation of a relevant package, but I guess if you're using Fleet, you'd always have the fleet agent package installed, so we'd always install these templates anyway.

Yep this is accurate. At the very least we want to _always_ support agent diagnostics, so to me it makes to install these templates as part of Fleet's default setup operation. 

I'll work on updating https://github.com/elastic/kibana/issues/143459 based on this approach. Is it best to use the `fields` defined in this PR to determine the base set of mappings to define on these templates? I'd like to prescribe those mappings as part of that implementation issue above.

kevinlog:
@kpollich 

> Is it best to use the `fields` defined in this PR to determine the base set of mappings to define on these templates?

The `fields` in `file_upload/files` representing the chunks I believe are the base fields and could be used directly.

The `fields` in `file_upload/files_data` representing the `metadata` has several fields that will be used by Endpoint only to start, I believe.  Based on @pzl [comment here](https://github.com/elastic/package-spec/pull/439#discussion_r999447994) there is the minimal set.  Perhaps we use that and the rely on `dynamic: true` for the rest?  Dan P. could comment.

kevinlog:
Leaving a separate comment here - can I get 👍 from all stakeholders to close this PR and [related package-spec discuss topic](https://github.com/elastic/package-spec/issues/437) with the conclusion that we will NOT make any package-spec or individual package changes and rely on Fleet installation code to add all requires ES components (index templates and ILM policies).

@joshdover @pzl @kpollich @juliaElastic @jsoriano

kpollich:
Hi folks, I've updated https://github.com/elastic/kibana/issues/143459 based on our conversations here. @pzl could you review the mappings/ILM policies I've included in the description of that issue to make sure they're consistent with what's specced out in this PR?

PR #452, opened 2022-11-23
Remove deprecated release tag in 2.3.0

jsoriano:
## What does this PR do?

Remove deprecated release tag in 2.3.0, and fix issue merging patches.

## Why is it important?

It causes confusion.

Adding this change has also exposed a bug when merging patches for multiple versions, this has been fixed too.

## Checklist

## Related issues

- Follow up of https://github.com/elastic/package-spec/issues/225.


code/go/internal/yamlschema/schema_spec.go
```diff
@@ -59,28 +59,18 @@ func (i *itemSchemaSpec) resolve(target semver.Version) (map[string]interface{},
...
-		if err != nil {
-			return nil, fmt.Errorf("failed to marshal patch: %w", err)
-		}
-
-		if len(patch) == 0 {
-			patch = patchData
-			continue
-		}
-
-		patch, err = jsonpatch.MergeMergePatches(patch, patchData)
```
jsoriano, line 80: This doesn't do what I thought. The library supports JSON patches, and "merge patches", this method is used to merge merge patches, but here we are working with JSON patches. JSON patches can be merged just by concatenating their actions.

PR #456, opened 2022-12-21
depracting csp_template fields

ofiriro3:
## What does this PR do?

We are deprecating two fields from our CSP template object.

## Why is it important?
During the development of 8.7.0 we've refactored the schema of the rule templates and it is important to keep the examples in the spec updated as well

## Checklist

## Related issues
- https://github.com/elastic/kibana/issues/143524


kfirpeled (approved):
LGTM

jsoriano (approved):

PR #461, opened 2023-01-13
Prepare version 2.3.0

jsoriano:


test/packages/good_v2/_dev/build/build.yml
```diff
@@ -1,4 +1,4 @@
 dependencies:
   ecs:
-    reference: git@v1.9.2
```
jsoriano, line 3: This tag doesn't exist.

mrodm: This package was copied from the test package `good`.
Looking into the test packages, there are some using the same reference:
```
 $ git grep 1.9.2 |grep build.yml
test/packages/good/_dev/build/build.yml:    reference: git@v1.9.2
test/packages/good_input/_dev/build/build.yml:    reference: git@v1.9.2
test/packages/good_v2/_dev/build/build.yml:    reference: git@v1.9.2
test/packages/missing_required_fields/_dev/build/build.yml:    reference: git@v1.9.2
test/packages/sql_input/_dev/build/build.yml:    reference: git@v1.9.2
```

Looking at the tags available in ecs, there is a v1.9.0 and v1.10.0. Should it be changed to one of those ?

mrodm (approved):

Issue #459, opened 2023-01-10
[Change Proposal] Allow security_rule objects to have rule IDs different from the object IDs


xcrzx:
Security Solution migrates the detection rules package from storing a single saved object per rule to multiple saved objects. See https://github.com/elastic/kibana/issues/137420 for more context regarding the change.

The package will contain rule saved objects with rule id and version in the name (`security_rule/[ruleId]_[ruleVersion].json`) with the following content:

```json
{
  "id": "[ruleId]_[ruleVersion]",
  "type": "security-rule",
  "attributes": {
    "rule_id": "[ruleId]",
    "version": "[ruleVersion]",
    // Other rule attributes
  }
}
```

So the saved object ID and the rule ID do not match anymore, making this validation check always return an error: https://github.com/elastic/package-spec/blob/fafbc4b947ed199a50430c4a6f204442bc207470/code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go#L50-L53

I would like to know why that validation exists in the first place and if we could remove or update it.

spong:
> I would like to know why that validation exists in the first place and if we could remove or update it.

So when this check was first added, @rw-access (who did the original implementation on the security side) [commented](https://github.com/elastic/package-spec/pull/175/files#r632049111):


> There's no reason these have to match.
>
> The `rule_id` is only relevant to the Security solution.
> The top level `id` is the saved object ID. It's possible that in the future we want better named SO ids. But we can loosen up this spec up then if that happens.

That said, as we were discussing earlier, there is some inherent use to this check as it ensures that all security-rule assets included as part of an integration are guaranteed to have a unique `rule_id`, a requirement of the Security Solution app. If rule_id must equal id, and the SO filename must be the `id`, and since all SO's are stored in the same directory, you can't have duplicates, and so can't have duplicate `rule_id`'s. Turns out this ended up being a roundabout and accidental safety check it seems.

I'd say Ross's comment stands then, as this check doesn't need to be happening this way, and can be removed. Security Solution needs to handle the case correctly where there are multiple security-rule assets with the same `rule_id`, but it would still be nice if there was some guard in place at the package layer that prevented developers from shipping packages that would be known to have invalid rules that would result in a more complicated UX (making the user pick which rule to install or something).

[...]

PR #463, opened 2023-01-16
Allow security_rule objects to have rule IDs different from the object IDs

xcrzx:
## What does this PR do?

Change validation check that enforces saved object ID to match `rule_id` for security solution prebuilt detection rules. Allow saved object IDs to have arbitrary suffixes.

## Why is it important?

Security Solution migrates the detection rules package from storing a single saved object per rule to multiple saved objects. See https://github.com/elastic/kibana/issues/137420 for more context regarding the change.

The package will contain rule saved objects with rule id and version in the name (`security_rule/[ruleId]_[ruleVersion].json`) with the following content:

```json
{
  "id": "[ruleId]_[ruleVersion]",
  "type": "security-rule",
  "attributes": {
    "rule_id": "[ruleId]",
    "version": "[ruleVersion]",
    // Other rule attributes
  }
}
```

So the saved object ID and the rule ID do not match anymore, making this validation check always return an error.

## Checklist

## Related issues

- Closes https://github.com/elastic/package-spec/issues/459


test/packages/good/kibana/security_rule/000047bb-b27a-47ec-8b62-ef1a5d2c9e19_5.json
```diff
@@ -0,0 +1,36 @@
...
+      "Okta",
+      "Continuous Monitoring",
+      "SecOps",
+      "Identity and Access"
+    ],
+    "timestamp_override": "event.ingested",
+    "type": "query",
+    "version": 5
+  },
+  "id": "000047bb-b27a-47ec-8b62-ef1a5d2c9e19_5",
```
banderror, line 34: "id": "000047bb-b27a-47ec-8b62-ef1a5d2c9e19"?

code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go
```diff
@@ -41,16 +41,11 @@ func ValidateKibanaObjectIDs(fsys fspath.FS) ve.ValidationErrors {
...
+			_, err := objectFile.Values("$.attributes.rule_id")
 			if err != nil {
 				errs = append(errs, errors.Wrapf(err, "unable to get rule ID in file [%s]", fsys.Path(filePath)))
 				continue
 			}
-
-			if ruleID != objectID {
-				errs = append(errs, errors.New("rule ID is different from the object ID"))
-				continue
-			}
```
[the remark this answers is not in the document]
xcrzx: > Could that (at least partially) address the possibility of someone copy-pasting a rule asset, updating the object ID but forgetting to update the `rule_id`?

Good idea 👍 Changed the validation for checking if the object ID starts with the rule ID for a given rule.

> Another question: is there an explicit validation somewhere that verifies the uniqueness of saved object IDs? Not the file names, but the combinations of SO id and type from the file contents?

I don't know, haven't seen checks like that in this repo. It looks like only filename uniqueness enforces saved object id uniqueness.

banderror (dismissed):
LGTM!

code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go
```diff
@@ -47,8 +47,9 @@ func ValidateKibanaObjectIDs(fsys fspath.FS) ve.ValidationErrors {
 				continue
 			}
 
-			if ruleID != objectID {
-				errs = append(errs, errors.New("rule ID is different from the object ID"))
+			if !strings.HasPrefix(objectID.(string), ruleID.(string)) {
```
mrodm, line 50: These type assertions could result in a panic if `objectId` or `ruleID` are not string.
Could you add those checks before this if statement ? Something like:
```suggestion
            stringObjectID, ok := objectID.(string)
            if !ok {
                errs = append(errs, errors.Wrapf(err, ...)
                continue
            }
            stringRuleID, ok := ruleID.(string)
            if !ok {
                errs = append(errs, errors.Wrapf(err, ...)
                continue
            }
			if !strings.HasPrefix(stringObjectID, stringRuleID) {
```

mrodm (commented):
LGTM, just a note about avoiding possible panics in some type assertions.

code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go
xcrzx: @mrodm Thanks for the suggestion 👍 Added additional checks

mrodm (approved):
Thanks!

LGTM! :+1:

code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go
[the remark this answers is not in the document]
spong: > >Another question: is there an explicit validation somewhere that verifies the uniqueness of saved object IDs? Not the file names, but the combinations of SO id and type from the file contents?
>
> I don't know, haven't seen checks like that in this repo. It looks like only filename uniqueness enforces saved object id uniqueness.

As far as I understand it's indeed the filename uniqueness that's enforcing the SO id uniqueness as well. Current behavior right now is last-one-in-wins if there are duplicate SO ID's ([comment](https://github.com/elastic/kibana/issues/128202#issuecomment-1084600441)).

xcrzx:
@mrodm It seems like this PR has all the required approves. How could I merge it? The merge button is inactive, saying, `You're not authorized to merge this pull request.`

mrodm:
> @mrodm It seems like this PR has all the required approves. How could I merge it? The merge button is inactive, saying, `You're not authorized to merge this pull request.`

@xcrzx As all the required approves are there, I'll merge it.

Issue #464, opened 2023-01-18
[Change Proposal] Support index template -> data_stream -> hidden boolean


kpollich:
The package spec doesn't expose a way to set the `data_stream.hidden` property on generated index templates, resulting in some hacky workarounds for some use cases. We should expose this setting directly in the `data_stream/**/manifest.yml` file, e.g. 

```yml
# some_package/data_stream/foo/manifest.yml
title: Foo
dataset: foo.bar
elasticsearch:
  index_template:
     data_stream:
       hidden: true
```

mrodm:
There is already a `hidden` field to define a data_stream hidden in the spec (at top level):
https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml#L170

Could it be used that field instead of creating a new one ? 

Here it is an example of how it should be used this field:
https://github.com/elastic/package-spec/blob/main/test/packages/good_v2/data_stream/hidden_data_stream/manifest.yml

Adding a new field for the same goal (hide a datastream) could cause some confusions on when to use which one.

felixbarny:
I agree that having both a top-level `hidden` field and a `elasticsearch.index_template.data_stream.hidden` field isn't ideal.

The one we're trying to add here is a direct 1:1 match to the `data_stream.hidden` field of the index template (see https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-put-template.html#put-index-template-api-request-body).

IMHO, the top-level `hidden` field is a misnomer and should be called something like `system` or `hidden_system` or something like that.

I suggest the following course of action:
1. Add `elasticsearch.index_template.data_stream.hidden` now
2. Deprecate the top-level `hidden` field
3. In case some integrations really need that functionality, re-name the field. The old `hidden` field name may still be kept for backwards compatibility.
4. Remove the top-level `hidden` field in the next major version.

kpollich:
I'm +1 to @felixbarny's approach above. Deprecating a package spec field has no impact on Fleet's functionality, and gives us a clear path towards addressing https://github.com/elastic/package-spec/issues/266.

mrodm:
@felixbarny's approach looks good :+1:
I guess the top level `hidden` should be deprecated before releasing a new package-spec version cc @elastic/ecosystem 

The only consideration is that if the field is renamed at some point, fleet should have to add support for that too.

PR #465, opened 2023-01-19
Add index_template.data_stream.hidden flag

felixbarny:
## What does this PR do?

Adds `index_template.data_stream.hidden` flag to data stream manifest

## Why is it important?

We need this urgently in APM as we want to release an important performance improvement in Kibana that makes use of continuously rolled up metrics for 1m, 10m, and 60m granularities. We need to hide the data streams with the non-default 10m and 60m buckets so that the OOTB ML jobs and existing custom dashboards don't break.

Sorry for having identified this blocker so late but we'd really hope to be able to get this in 8.7. This feels like a rather trivial and non-controversial change. Hopefully this turns out to be true.

## Checklist

## Related issues

Closes https://github.com/elastic/package-spec/issues/464

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -18,6 +18,9 @@
   - description: Add flag in _dev/build/build.yml to import ECS mappings.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/455
+  - description: Add index_pattern.data_stream.hidden flag to data stream manifest
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/465
 - version: 2.2.0
   changes:
   - description: Support definition of system tests for input packages
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -7,6 +7,9 @@
   - description: Allow security_rule objects to have rule IDs different from the object IDs.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/463
+  - description: Add index_pattern.data_stream.hidden flag to data stream manifest
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/465
 - version: 2.3.0
   changes:
   - description: Remove the release tag, semantic versioning should be used instead.
@@ -18,9 +21,6 @@
   - description: Add flag in _dev/build/build.yml to import ECS mappings.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/455
-  - description: Add index_pattern.data_stream.hidden flag to data stream manifest
-    type: enhancement
-    link: https://github.com/elastic/package-spec/pull/465
 - version: 2.2.0
   changes:
   - description: Support definition of system tests for input packages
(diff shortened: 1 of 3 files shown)


simitt (approved):
Changes LGTM

spec/changelog.yml
```diff
@@ -18,6 +18,9 @@
   - description: Add flag in _dev/build/build.yml to import ECS mappings.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/455
+  - description: Add index_pattern.data_stream.hidden flag to data stream manifest
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/465
```
jsoriano, line 23: Please move this entry to the 2.4.0-next section above.

felixbarny: done

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -126,6 +126,14 @@ spec:
...
           required:
           - name
+        data_stream:
+          description: Data stream section of index template
+          type: object
+          additionalProperties: false
+          properties:
+            hidden:
+              description: Makes the data stream hidden
+              type: boolean
```
mrodm, line 136: If it is defined here in this key, input packages use this key as a reference to introduce the `elasticsearch_index_template`:

https://github.com/elastic/package-spec/blob/main/spec/input/manifest.spec.yml#L105

But it should not be there for input packages.

felixbarny: Do you think it could be an issue if input packages can define `index_pattern.data_stream.hidden`? Is there a good reason to only allow it for integrations but not for input packages?

mrodm: I've just realized that it was going to be applied to both types of packages and I wanted to highlight it.
I don't have any reason to just allow this field in one of them 
@jsoriano do you see any issue on allowing this field in input packages?

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -126,6 +126,14 @@ spec:
               type: string
           required:
           - name
+        data_stream:
+          description: Data stream section of index template
+          type: object
+          additionalProperties: false
+          properties:
+            hidden:
+              description: Makes the data stream hidden
```
mrodm, line 135: Should it be remarked somehow the difference between the other `hidden` field? Or update the description of the other `hidden` field at the top level ?

spec/integration/data_stream/manifest.spec.yml
jsoriano: I don't see any usage of this in input packages, but I guess it won't cause problems neither, let's go on like this.

spec/integration/data_stream/manifest.spec.yml
jsoriano: +1 to update the description and possibly deprecate the top level `hidden` field.

PR #466, opened 2023-01-19
Update top hidden field description in datastream manifest

mrodm:
## What does this PR do?

Updates the description of the top level `hidden` field in the manifest datastream:
https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml#L178-L180

This field is also set as "deprecated". It should be used the [new `hidden` field](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml#L134) that can be defined under `elsaticsearch`: `elasticsearch.index_template.data_stream.hidden`

## Why is it important?

Updating the description of the top level allows to differentiate from the `hidden` defined under the `elasticsearch` key, to be sure which one should used in each case.

## Checklist

## Related issues

- Relates #464

### DIFF
--- a/spec/integration/data_stream/manifest.spec.yml
+++ b/spec/integration/data_stream/manifest.spec.yml
@@ -176,8 +176,9 @@ spec:
       examples:
       - metrics
     hidden:
-      description: Specifies if a data stream is hidden
+      description: Specifies if a data stream is hidden, resulting in dot prefixed system indices
       type: boolean
+      deprecated: true # https://github.com/elastic/package-spec/issues/464
     streams:
       description: Streams offered by data stream.
       type: array
--- a/spec/integration/data_stream/manifest.spec.yml
+++ b/spec/integration/data_stream/manifest.spec.yml
@@ -176,7 +176,9 @@ spec:
       examples:
       - metrics
     hidden:
-      description: Specifies if a data stream is hidden, resulting in dot prefixed system indices
+      description: >
+        Specifies if a data stream is hidden, resulting in dot prefixed system indices.
+        To set the data stream hidden without those dot prefixed indices, check `elasticsearch.index_template.data_stream.hidden` flag.
       type: boolean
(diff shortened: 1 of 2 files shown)


spec/integration/data_stream/manifest.spec.yml
```diff
@@ -176,8 +176,9 @@ spec:
       examples:
       - metrics
     hidden:
-      description: Specifies if a data stream is hidden
+      description: Specifies if a data stream is hidden, resulting in dot prefixed system indices
       type: boolean
+      deprecated: true # https://github.com/elastic/package-spec/issues/464
```
mrodm, line 181: Set the field as deprecated, but not removed nor renamed for now
WDYT ?

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -176,8 +176,9 @@ spec:
       examples:
       - metrics
     hidden:
-      description: Specifies if a data stream is hidden
+      description: Specifies if a data stream is hidden, resulting in dot prefixed system indices
```
mrodm, line 179: Any other suggestion for this description ?

For context, the description of the other `hidden` field under `elasticsearch` key is:
```yaml
  properties:                                  
    hidden:                                    
      description: Makes the data stream hidden
      type: boolean                            
```

felixbarny (dismissed):
LGTM

spec/integration/data_stream/manifest.spec.yml
jsoriano: Sounds good, maybe add a reference to the other setting for hidden data streams.

kpollich (approved):

spec/integration/data_stream/manifest.spec.yml
mrodm: Updated description of the top level hidden field

jsoriano (approved):

PR #469, opened 2023-01-27
Prepare version 2.5.0 with new subcategories

jsoriano:
Subcategories were added to the package registry in https://github.com/elastic/package-registry/pull/914, but we didn't add them to the spec, so they cannot be actually used in packages.

### DIFF
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -12,31 +12,71 @@ spec:
       items:
         type: string
         enum:
+          - analytics_engine
+          - application_observability
+          - app_search
+          - auditd
+          - authentication
           - aws
           - azure
+          - big_data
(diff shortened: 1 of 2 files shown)


spec/integration/manifest.spec.yml
```diff
@@ -12,31 +12,71 @@ spec:
...
           - custom
+          - custom_logs
+          - database_security
           - datastore
+          - dns_security
+          - edr_xdr
+          - elasticsearch_sdk
           - elastic_stack
+          - email_security
+          - enterprise_serch
```
mrodm, line 43: There is a typo here. 
I see that this is also in package-registry https://github.com/elastic/package-registry/blob/main/categories/categories.yml#L30

mrodm (approved):

Issue #480, opened 2023-02-22
Release input packages spec as GA


jsoriano:
Spec for input packages is currently in beta, what doesn't allow to release input packages as GA, as seen in https://github.com/elastic/integrations/pull/5347.

We have been already implementing some input packages, we can already have system tests and support is implemented in Kibana/Fleet. We should start considering to release the input package spec as GA. The spec should check that the kibana constraint is set for versions >= 8.8.0 https://github.com/elastic/package-spec/issues/480#issuecomment-1441537571.

Some code related to this:
* Input spec marked as beta in https://github.com/elastic/package-spec/blob/ff8286d0c40ad76bb082e9c8ea78f4551c2519c1/spec/input/spec.yml#L14
* Validation checking that packages using beta specs cannot be released as GA: https://github.com/elastic/package-spec/blob/ff8286d0c40ad76bb082e9c8ea78f4551c2519c1/code/go/internal/validator/folder_spec.go#L61-L63 
* Input test packages don't have GA versions:
  * `good_input`: https://github.com/elastic/package-spec/blob/ff8286d0c40ad76bb082e9c8ea78f4551c2519c1/test/packages/good_input/manifest.yml#L7
  * `sql_input`: https://github.com/elastic/package-spec/blob/ff8286d0c40ad76bb082e9c8ea78f4551c2519c1/test/packages/sql_input/manifest.yml#L7

hop-dev:
@jsoriano I think we should add a check to only allow input packages to be GA if they have a condition for kibana version to be greater than or equal to 8.8.0 which is when we plan to make them GA in kibana, do you think that is reasonable? 

CC @kpollich

jsoriano:
> @jsoriano I think we should add a check to only allow input packages to be GA if they have a condition for kibana version to be greater than or equal to 8.8.0 which is when we plan to make them GA in kibana, do you think that is reasonable?

Yes, I think this could be done.

jsoriano:
@hop-dev if you do this task, I think that releasing the spec as GA is quite straight forward, it would be to remove the [`release: beta` line in the input spec](https://github.com/elastic/package-spec/blob/ff8286d0c40ad76bb082e9c8ea78f4551c2519c1/spec/input/spec.yml#L14,), and to have some test case, modify the `good_input` package version to `1.0.0`.

[...]

PR #485, opened 2023-03-01
Release input package spec as GA

hop-dev:
## What does this PR do?

Removes `release: beta` from the input package spec.

Adds validation to check that a GA input package enforces kibana version must be ^8.8.0.

Changes the test input packages to test this new behaviour.

## Why is it important?

This allows input packages to have release versions >= 1.0.0. For example the [custom logs integration](https://github.com/elastic/integrations/pull/5347) 

## Checklist

## Related issues

Closes https://github.com/elastic/package-spec/issues/480


code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -0,0 +1,153 @@
...
+
+	sVal, ok := val.(string)
+	if !ok {
+		return "", errors.New("manifest kibana version is not a string")
+	}
+
+	return sVal, nil
+}
+
+func kibanaVersionConditionIsGreaterThanOrEqualTo8_8_0(kibanaVersionCondition string) bool {
```
hop-dev, line 126: @jsoriano I am not sure if this function is a bit too hacky. I am a bit confused about the spec for `conditions.kibana.version`. All of the integrations use `^` and `||` syntax. However in the spec we say that `>=` can be used https://github.com/elastic/package-spec/blob/57e322febbb4b22e72b60326d3e64955a2d9d7b5/spec/integration/manifest.spec.yml#L113.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, errors.New("package version is not a string")
+	}
+
+	sVersion, err := semver.NewVersion(sVal)
+	return sVersion, nil
+}
+
+func getKibanaVersionCondition(manifest pkgpath.File) (string, error) {
+
+	val, err := manifest.Values("$.conditions[\"kibana.version\"]")
```
hop-dev, line 113: do we need to handle:

```
conditions:
    kibana.version: 1234
```

and

```
conditions:
    kibana:
        version: 1234
```

here? It seems the lib treats them differently, all packages use `kibana.version`

jsoriano: This actually looks like an issue with this `Values` helper. The spec allows both `kibana.version`, or `version` as an attribute inside `kibana`.

It should probably use ucfg to parse these files instead of plain yaml. See https://github.com/elastic/go-ucfg#dot-notations

hop-dev: I'll have a go at this 👌

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
jsoriano: Yes, in principle any comparison is allowed here as supported by the underlying library https://github.com/Masterminds/semver/#checking-version-constraints

Your implementation makes sense to me. I guess it will only have problems if someone uses "less than" conditions, like `< 8.10.0`, but I don't think they make a lot of sense in this context.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, match := range matches {
+		matchVersion, err := semver.NewVersion(match)
+		if err != nil {
+			return false
+		}
+
+		if matchVersion.LessThan(semver8_8_0) {
+			return false
+		}
+	}
```
mrodm, line 150: Maybe this can be changed to use `semver.NewConstraint`? 

https://pkg.go.dev/github.com/Masterminds/semver/v3#hdr-Checking_Version_Constraints_and_Comparing_Versions

There is an example in elastic-package that just do those checks (there the kibana version is already in the struct):
- https://github.com/elastic/elastic-package/blob/main/internal/packages/conditions.go#L28

IIRC and following that example in elastic-package, the requirement kibana version would be "8.8.0" and the constraint the one defined in the manifest.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -0,0 +1,153 @@
...
+
+	packageType, err := getPackageType(*manifest)
+	if err != nil {
+		return ve.ValidationErrors{err}
+	}
+
+	packageVersion, err := getPackageVersion(*manifest)
+	if err != nil {
+		return ve.ValidationErrors{err}
+	}
```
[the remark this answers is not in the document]
hop-dev: Ah I missed this util, thanks!

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
hop-dev: Will do 👍

mrodm: Doing some quick tests about this, probably it could not be achieved using semver.NewConstraint. It would be like comparing two constraints, the one from the package (kibana.version) against `>= 8.8.0`, and I think that could not be done with that.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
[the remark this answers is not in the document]
hop-dev: Cyclic dependency prevents me from doing this unfortunately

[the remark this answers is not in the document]
mrodm: An option to break this cyclic dependency could be move all the `code/go/internal/validator/package*.go` files to its own package. For instance it could be created a new package like `code/go/internal/packages/` and update all the references.

spec/changelog.yml
```diff
@@ -2,11 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom of each in-development version.
 ##
-- version: 2.5.2-next
+- version: 2.6.0
```
mrodm, line 5: I'd prefer to keep 2.5.2-next version in changelog for now. There could be other branches merged before creating the next release version.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, errors.Wrap(err, "can't read manifest version")
+	}
+
+	sVal, ok := val.(string)
+	if !ok {
+		return nil, errors.New("package version is not a string")
+	}
+
+	sVersion, err := semver.NewVersion(sVal)
+	return sVersion, nil
```
mrodm, line 108: you can return directly the value from `semver.NewVersion` function

```suggestion
	return semver.NewVersion(sVal)
```

.gitignore
```diff
@@ -2,3 +2,4 @@
 temp/
 fuzz
 /build/
+.DS_Store
```
jsoriano, line 5: Tip: you can add files specific to your environment to `~/.gitignore_global`. These will apply to all your repos. :slightly_smiling_face:

hop-dev: Ah, TIL! Thanks I've removed it

hop-dev:
@mrodm I have resolved the cyclic dependency issue, my validator now uses `NewPackageFromFS`, just resolving the merge conflict now

mrodm (approved):
LGTM

Issue #453, opened 2022-11-24
[Change Proposal] Add support for "select/option" type for Integration parameters


miltonhultgren:
In the Elasticsearch Integration, when configuring metrics collection we have an option for the scope of collection. There are only two valid settings (node or cluster), as we point out in the field description.

<img width="945" alt="Screenshot 2022-11-24 at 12 42 12" src="https://user-images.githubusercontent.com/2564140/203776231-3edc0848-e1b1-462e-a7de-95f30ef6906d.png">

Ideally we would instead expose a dropdown list of the possible valid options so that the user cannot make a mistake by misspelling or writing something that isn't supported.

kpollich:
As part of this work we'll want to do two additional follow-ups

1. Communicate the availability of the new variable type to the `elastic-package` and `fleet-contributors` mailing groups
2. File a PR against https://github.com/elastic/integrations updating the Elasticsearch integration mentioned in the description to use the new `select` variable type for its `scope` variable

cc @jillguyonnet

jillguyonnet:
Here are proposed changes (see ❓ for open questions).

#### package-spec
Introduce a new `select` var type for integrations. In order to make use of a `select` field, the integration needs to provide the following data:
* A list of options, each with a value and a text (label).
* (Optional): a default value. ❓ An alternative could be to decide that whichever option is listed first is the default.

A suggestion on how to implement this is in https://github.com/elastic/package-spec/pull/486, where a new `options` property with `array` type is introduced. The items are objects with a `value` and a `text` field. ❓ Note that this could be simplified by assuming the value and text always match, in which case `options` could simply be an array of strings. The default value was left out for now.

❓ If we follow this approach, perhaps a good addition would be to add a linting rule that the `options` property is only allowed on `select` types.

#### kibana
With this new var type in place, adding support for the new select type in Kibana should be straightforward enough (https://github.com/elastic/kibana/pull/152550).

#### integrations
3. And the Elasticsearch integration could be updated accordingly (https://github.com/elastic/integrations/pull/5451). ❓ As part of this, the description of the setting should probably be amended slightly.

@elastic/ecosystem Can you please weigh in on this approach?

kpollich:
> (Optional): a default value. ❓ An alternative could be to decide that whichever option is listed first is the default.

I'm +1 on using the first option listed as the default and documenting this in the `options` field's description. Adding another spec option seems excessive here, and a documented note seems to accomplish what we need.

[...]

PR #486, opened 2023-03-02
Add select var type

jillguyonnet:
## What does this PR do?

This PR adds a new `select` var type for integrations.

As integrations need a way to provide a list of options for the select, it also adds an `options` property which is only relevant to `select` types.

The `options` property is conditionally added, such as it is required when `type` is `select` and required **not to be there** for other types. The conditional rule to enforce this was inspired from [this post](https://stackoverflow.com/questions/16583485/how-to-define-the-min-size-of-array-in-the-json-schema).

I also added a requirement for the `options` array to contain at least one element.

## Why is it important?

As detailed in https://github.com/elastic/package-spec/issues/453, the Elasticsearch integration has an setting with only two valid options. Currently, this option is set in the UI via a text field with a description, which is error prone. Adding a select (dropdown) var type will enhance this UX.

## Checklist

~~- [ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~~

## Related issues

- Relates https://github.com/elastic/package-spec/issues/453


spec/integration/data_stream/manifest.spec.yml
```diff
@@ -75,6 +76,27 @@ spec:
...
+                  - cluster
+                text:
+                  type: string
+                  examples:
+                  - node
+                  - cluster
+              required:
+              - value
+              - text
+            default: []
```
mrodm, line 99: As this options property should just be defined if the `select` type is chosen, it could be tried to use the if-then-else of json-schema to force that:
https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else

If that does not work, we would need to add a new validate rule for that. There are some similar validator rules in place now , for instance:
https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/validate_field_groups.go
If a new rule is added, it should be added here too: https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/spec.go#L86

mrodm (commented):
That new type `select` and the related `options` object look a good approach :+1: 

Please, update the test package good_v2 to include a new field using this new `select` type with some options.

And add also some new test "bad" package, for instance, with a package  where the options is defined but with another type. It should be added here a new test case https://github.com/elastic/package-spec/blob/60560379ad6ed5ffa6fc3beb13ec22ad89bdddc4/code/go/pkg/validator/validator_test.go#L21

code/go/pkg/validator/validator_test.go
```diff
@@ -119,6 +119,15 @@ func TestValidateFile(t *testing.T) {
...
 		},
+		"bad_select": {
+			"data_stream/foo_stream/manifest.yml",
+			[]string{
+				"field streams.0.vars.1: Must validate \"then\" as \"if\" was valid",
+				"field streams.0.vars.1: options is required",
+				"field streams.0.vars.1: Must validate \"then\" as \"if\" was valid",
+				"field streams.0.vars.2.options: Invalid type. Expected: array, given: null",
+			},
+		},
```
jillguyonnet, line 130: It is not ideal that each failed validation gives two error messages:
1. `field streams.0.vars.1: Must validate "then" as "if" was valid`
2. And the actual error message.

Also I wish the error message for when `options` is an empty array was better than `Expected: array, given: null`.

Any thoughts?

mrodm: > Also I wish the error message for when options is an empty array was better than Expected: array, given: null.
weird, maybe the gojsonschema lib transforms the empty array to null ? 


About the error messages (two messages per each error), as they are the ones reported by the jsonschema validation, I think it cannot be done nothing here.

test/packages/good_input/manifest.yml
```diff
@@ -58,6 +58,18 @@ policy_templates:
...
+        title: Response caching
+        description: "Optional response caching time"
+        multi: false
+        required: true
+        show_user: false
+        options:
+          - value: "0"
+            text: None
+          - value: "1"
+            text: "1 day"
```
jillguyonnet, line 72: Bit of a contrived example, a `select` would probably have been way more suitable for the `driver` or `sql_response_format` variables. Happy to change it if you want.

code/go/pkg/validator/validator_test.go
jillguyonnet: Agreed. The new error message (when `options` are passed when `type` is `text`) is even worse 🙂

test/packages/good_input/manifest.yml
mrodm: That example looks good. With that we have a test also to check this field type for input packages :+1:

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -102,6 +104,39 @@ spec:
           default:
             description: Default value(s) for variable
             $ref: "#/definitions/input_variable_value"
+        if:
```
jsoriano, line 107: Oh nice, maybe we can use if/then in more cases where we are validating with go now.

code/go/pkg/validator/validator_test.go
```diff
@@ -119,6 +119,17 @@ func TestValidateFile(t *testing.T) {
...
 		},
+		"bad_select": {
+			"data_stream/foo_stream/manifest.yml",
+			[]string{
+				"field streams.0.vars.1: Must validate \"then\" as \"if\" was valid",
+				"field streams.0.vars.1: options is required",
+				"field streams.0.vars.2: Must validate \"then\" as \"if\" was valid",
+				"field streams.0.vars.2.options: Invalid type. Expected: array, given: null",
+				"field streams.0.vars.3: Must validate \"else\" as \"if\" was not valid",
+				"field streams.0.vars.3: Must not validate the schema (not)",
```
jsoriano, line 130: Do these error end up in the user output? They look weird.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -102,6 +104,40 @@ spec:
...
+                required:
+                - value
+                - text
+              min_items: 1
+              default: []
+        else:
+          not:
+            required:
+              - options
         required:
```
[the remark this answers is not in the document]
jsoriano: As you prefer. Historically the main reasons we have had to force the use of a spec version are:
* Change is potentially breaking, so we don't want it to apply to packages using old versions of the spec (this doesn't seem to be the case here).
* Encouraging developers to use more recent versions of the spec.

In the future we may be more strict, and always introduce new features in new versions. If we want to follow this approach already here, this feature should be introduced in 2.6.0. For that we would need to use `before: 2.6.0`, and add a `2.6.0-next` block in the changelog.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -52,6 +52,7 @@ spec:
               - email
               - integer
               - password
+              - select
```
jsoriano, line 55: What do current versions of kibana do when finding this type?

jillguyonnet: They would render a text field (empty in this case), which is the fallback in case the type is unknown.

code/go/pkg/validator/validator_test.go
jillguyonnet: I agree they are not very helpful. I guess they will show up in the user output - is there a way to test this now? At the moment `elastic-package check` doesn't know about select and options.

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
jillguyonnet: Thanks @jsoriano , that sounds good to me. @mrodm WDYT?

code/go/pkg/validator/validator_test.go
mrodm: You can build `elastic-package` replacing the package-spec dependency.
For that you can run these commands:
```
go mod edit -replace github.com/elastic/elastic-package2=github.com/jillguyonnet/elastic-package@<your_latest_changeset_in_PR>

go mod tidy 

make build
```

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
mrodm: Agreed! :+1:

code/go/pkg/validator/validator_test.go
jillguyonnet: Thanks! Yep, confirmed, this is the output. As you said, I guess we can't do much about it.
```
Error: checking package failed: linting package failed: found 6 validation errors:
   1. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.1: Must validate "then" as "if" was valid
   2. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.1: options is required
   3. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.2: Must validate "then" as "if" was valid
   4. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.2.options: Invalid type. Expected: array, given: null
   5. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.3: Must validate "else" as "if" was not valid
   6. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.3: Must not validate the schema (not)
```

jsoriano: We could filter-out certain errors. Here I think it'd be safe to assume that we can remove all these errors starting with `Must validate`/`Must not validate`. This could be done before returning the errors in `ValidatePackage`.

This would leave these errors, that are still a bit redundant, but more meaningful for users:
```
Error: checking package failed: linting package failed: found 2 validation errors:
   1. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.1: options is required
   2. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.2.options: Invalid type. Expected: array, given: null
```

spec/integration/data_stream/manifest.spec.yml
jsoriano: Ok, I guess this is fine then, users can still configure the packages.

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
jillguyonnet: Mmh, there seems to be a catch 22: that makes the tests fail as version is not 2.6.0 yet :) Does this need to go into the next PR?

code/go/pkg/validator/validator_test.go
mrodm: For these two error messages, it should be shown something to the user 
```
   5. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.3: Must validate "else" as "if" was not valid
   6. file "/Users/jillguyonnet/Code/jillguyonnet/package-spec/test/packages/bad_select/data_stream/foo_stream/manifest.yml" is invalid: field streams.0.vars.3: Must not validate the schema (not)
```

Those errors are for the case that options is defined in a different type (e.g. string). However, those messages do not show helpful information about it :/

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
mrodm: What is the error ? 
Just in case, if this is added just from spec version >= 2.6.0, `format_spec` field in the manifests of the test packages should be updated to at least that version.

code/go/pkg/validator/validator_test.go
jillguyonnet: I think it would be OK to only have 1 error instead of 2, so 3 instead of 6 in the example above:
```
options is required
Invalid type. Expected: array, given: null
Must not validate the schema (not)
```
As you said, however, the third one is really unhelpful, it really should be something like `options is forbidden when type is not select` or something similar. I guess in addition to filtering errors we could transform the message... but I'm not sure it can be done in a way that will scale to future conditional cases.
The second error is also confusing (the issue is that the array is empty), but it's a bit less bad.

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
jillguyonnet: Thanks, it works now!

code/go/pkg/validator/validator_test.go
jillguyonnet: I've been trying to set up a custom error message in the JSON schema without luck so far. I can add filtering if you want on the following errors:
```
"Must validate \"then\" as \"if\" was valid"
"Must validate \"else\" as \"if\" was not valid"
```
but I think we should keep `Must not validate the schema (not)`, otherwise we'll silence the error when options should not be present, which is worse than a bad error message. :)
Let me know your thoughts.

jsoriano: Ok. Maybe we can replace `Must not validate the schema (not)` with `Must not be present`, around the same place where we do the filtering for the other errors.

jillguyonnet: Ok, I've pushed a tentative solution (Mario warned me I'd be missing map and filter on arrays, he was right 😆), let me know of any improvements.
Note that there already was an error message with `Must not validate the schema (not)` [elsewhere](https://github.com/elastic/package-spec/pull/486/files#diff-4dd1d701e9e3346e2b2ccdf83fbb4f2f94f3ae3373a8cfe7c962716fc3dc323bR101), I modified it too.

code/go/internal/validator/spec.go
```diff
@@ -64,7 +65,38 @@ func (s Spec) ValidatePackage(pkg Package) ve.ValidationErrors {
...
+		"Must validate \"then\" as \"if\" was valid",
+		"Must validate \"else\" as \"if\" was not valid",
+	}
+	for _, e := range errs {
+		for _, msg := range msgTransforms {
+			if strings.Contains(e.Error(), msg.original) {
+				processedErrs = append(processedErrs, errors.New(strings.Replace(e.Error(), msg.original, msg.new, 1)))
+			} else if !substringInSlice(e.Error(), redundant) {
+				processedErrs = append(processedErrs, e)
+			}
```
jsoriano, line 86: Nit. I tend to prefer to avoid elses and have each exception on its own if:
```suggestion
			if strings.Contains(e.Error(), msg.original) {
				processedErrs = append(processedErrs, errors.New(strings.Replace(e.Error(), msg.original, msg.new, 1)))
				continue
			}
			if substringInSlice(e.Error(), redundant) {
				continue
			}
			processedErrs = append(processedErrs, e)
```
Or maybe use a switch.
```suggestion
			switch {
			case strings.Contains(e.Error(), msg.original):
				processedErrs = append(processedErrs, errors.New(strings.Replace(e.Error(), msg.original, msg.new, 1)))
			case substringInSlice(e.Error(), redundant):
			default:
				processedErrs = append(processedErrs, e)
			}
```

code/go/pkg/validator/validator_test.go
jsoriano: > I'd be missing map and filter on arrays

Now that there are generics in Go there are some attempts :smile: https://pkg.go.dev/golang.org/x/exp/slices

jsoriano (approved):
:+1:

code/go/pkg/validator/validator_test.go
```diff
@@ -119,6 +119,14 @@ func TestValidateFile(t *testing.T) {
 				fmt.Sprintf("field ilm_policy: ILM policy \"logs-bad_custom_ilm_policy.test-notexists\" not found in package, expected definition in \"%sbad_custom_ilm_policy/data_stream/test/elasticsearch/ilm/notexists.json\"", osTestBasePath),
 			},
 		},
+		"bad_select": {
+			"data_stream/foo_stream/manifest.yml",
+			[]string{
+				"field streams.0.vars.1: options is required",
+				"field streams.0.vars.2.options: Invalid type. Expected: array, given: null",
+				"field streams.0.vars.3: Must not be present",
+			},
```
mrodm, line 128: Perfect, this is more helpful for the developers :+1:

mrodm (approved):
LGTM!

jillguyonnet:
Thank you for the review @jsoriano and @mrodm 👍 I've rebased the branch to update spec/changelog.yml.

mrodm (approved):

PR #487, opened 2023-03-06
Use ucfg for file parsing

hop-dev:
## Checklist

## Related issues

-

### DIFF
--- a/code/go/internal/pkgpath/files.go
+++ b/code/go/internal/pkgpath/files.go
@@ -73,7 +73,9 @@ func (f File) Values(path string) (interface{}, error) {
 			return nil, errors.Wrapf(err, "parsing yaml file failed (path: %s)", f.fsys.Path(fileName))
 		}
 
-		config.Unpack(&v)
+		if err = config.Unpack(&v); err != nil {
+			return nil, errors.Wrap(err, "can't unpack file (path: %s)", f.fsys.Path(fileName))
+		}
 	} else if fileExt == "json" {
 		config, err := json.NewConfig(contents)
 		if err != nil {
--- a/code/go/internal/pkgpath/files.go
+++ b/code/go/internal/pkgpath/files.go
@@ -85,7 +85,7 @@ func (f File) Values(path string) (interface{}, error) {
 	return jsonpath.Get(path, v)
 }
 
-// Same as values but returns an array of values
+// ValuesArray Same as values but returns an array of values
 func (f File) ValuesArray(path string) ([]interface{}, error) {
 	fileName := f.Name()
 	fileExt := strings.TrimLeft(filepath.Ext(fileName), ".")
--- a/code/go/internal/pkgpath/files.go
+++ b/code/go/internal/pkgpath/files.go
@@ -5,7 +5,6 @@
 package pkgpath
 
 import (
-	"encoding/json"
 	"fmt"
 	"io/fs"
 	"os"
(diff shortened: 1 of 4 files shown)


code/go/internal/pkgpath/files.go
```diff
@@ -66,20 +66,70 @@ func (f File) Values(path string) (interface{}, error) {
...
+	if fileExt != "json" && fileExt != "yaml" && fileExt != "yml" {
+		return nil, fmt.Errorf("cannot extract values from file type = %s", fileExt)
+	}
+
+	contents, err := fs.ReadFile(f.fsys, f.path)
+	if err != nil {
+		return nil, errors.Wrap(err, "reading file content failed")
+	}
+
+	var v []interface{}
```
hop-dev, line 102: only difference with Values is that we initialise v as []interface which then allows ucfg to assign the top level array of changelog correctly. If we just use interface{} then the changelog is returned as an empty struct.

code/go/internal/pkgpath/files.go
```diff
@@ -66,20 +66,70 @@ func (f File) Values(path string) (interface{}, error) {
...
+	v := make(map[string]interface{})
 	if fileExt == "yaml" || fileExt == "yml" {
-		if err := yaml.Unmarshal(contents, &v); err != nil {
-			return nil, errors.Wrapf(err, "unmarshalling YAML file failed (path: %s)", f.fsys.Path(fileName))
+		config, err := yaml.NewConfig(contents)
+		if err != nil {
+			return nil, errors.Wrapf(err, "parsing yaml file failed (path: %s)", f.fsys.Path(fileName))
 		}
+
+		config.Unpack(&v)
```
mrodm, line 76: missing check the error returned by the Unpack function

```suggestion
		if err = config.Unpack(&v); err != nil {
			return nil, errors.Wrap(err, "can't unpack file (path: %s)", f.fsys.Path(fileName))
		}
```

code/go/internal/pkgpath/files.go
```diff
@@ -66,20 +66,70 @@ func (f File) Values(path string) (interface{}, error) {
...
+		}
+
+		config.Unpack(&v)
+	} else if fileExt == "json" {
+		config, err := json.NewConfig(contents)
+		if err != nil {
+			return nil, errors.Wrapf(err, "parsing json file failed (path: %s)", f.fsys.Path(fileName))
+		}
+		config.Unpack(&v)
+	}
```
mrodm, line 116: nit. `config.Unpack(&v)` call can be moved out from the if-else. Same for `Values()` function

```suggestion
	if fileExt == "yaml" || fileExt == "yml" {
		config, err := yaml.NewConfig(contents)
		if err != nil {
			return nil, errors.Wrapf(err, "parsing yaml file failed (path: %s)", f.fsys.Path(fileName))
		}
	} else if fileExt == "json" {
		config, err := json.NewConfig(contents)
		if err != nil {
			return nil, errors.Wrapf(err, "parsing json file failed (path: %s)", f.fsys.Path(fileName))
		}
	}
	if err := config.Unpack(&v); err != nil {
		return nil, errors.Wrapf(err, "can't unpack file (path: %s)", f.fsys.Path(fileName))
	}
```

code/go/internal/pkgpath/files.go
```diff
@@ -66,20 +66,70 @@ func (f File) Values(path string) (interface{}, error) {
...
+			return nil, errors.Wrapf(err, "parsing json file failed (path: %s)", f.fsys.Path(fileName))
 		}
+		config.Unpack(&v)
 	}
 
 	return jsonpath.Get(path, v)
 }
 
+// ValuesArray Same as values but returns an array of values
+func (f File) ValuesArray(path string) ([]interface{}, error) {
```
mrodm, line 89: Would it be useful to check whether or not `fileName` is `changelog.yml` to just use one function? Depending on that maybe it could be unpack the config into one or another object type.

code/go/internal/pkgpath/files.go
hop-dev: Thanks I'll give it a go

PR #490, opened 2023-03-15
Remove `monitoring_infrastructure` category from package spec

jen-huang:
## What does this PR do?

Related to https://github.com/elastic/package-registry/pull/970. With the deprecation of parent `Infrastructure` category, remove `monitoring_infrastructure` category as no integrations are currently using it.

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -2,11 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom of each in-development version.
 ##
-- version: 2.5.2-next
+- version: 2.5.2
   changes:
-  - description: Prepare for next version
+  - description: Remove monitoring_infrastructure category
     type: enhancement
-    link: https://github.com/elastic/package-spec/pull/479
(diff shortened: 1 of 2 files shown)


spec/changelog.yml
```diff
@@ -2,11 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom of each in-development version.
 ##
-- version: 2.5.2-next
+- version: 2.5.2
```
jsoriano, line 5:
```suggestion
- version: 2.6.0-next
```

spec/changelog.yml
```diff
@@ -2,11 +2,11 @@
...
   changes:
-  - description: Prepare for next version
+  - description: Remove monitoring_infrastructure category
     type: enhancement
```
jsoriano, line 8:
```suggestion
    type: breaking-change
```

jsoriano (commented):
This should be released as breaking change, possibly in a new minor.

spec/changelog.yml
jen-huang: thanks! I wasn't sure of the release process

PR #491, opened 2023-03-20
add support for transform `settings.unattended`

joeypoon:
## What does this PR do?

This adds support for the transform `settings.unattended` option.

## Why is it important?

We [recently added unattended: true](https://github.com/elastic/endpoint-package/pull/353) into `endpoint-package`. We were previously managing the health of our transforms manually (kibana task for restarting / reinstalling our package transforms).

## Checklist

## Related issues

- Relates https://github.com/elastic/security-team/issues/6165

### DIFF
--- a/test/packages/good/elasticsearch/transform/metadata_current/transform.yml
+++ b/test/packages/good/elasticsearch/transform/metadata_current/transform.yml
@@ -18,3 +18,5 @@ sync:
   time:
     field: event.ingested
     delay: 1s
+settings:
+  unattended: true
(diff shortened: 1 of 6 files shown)


test/packages/good/elasticsearch/transform/metadata_current/transform.yml
```diff
@@ -18,3 +18,5 @@ sync:
   time:
     field: event.ingested
     delay: 1s
+settings:
+  unattended: true
```
mrodm, line 22: Add it also in the test package `good_v2`. That package is based on the package spec 2.x

mrodm (commented):
LGTM!
Just a comment to add also this new field in `good_v2`.

mrodm (approved):
:+1:

Issue #483, opened 2023-02-27
[Change Proposal] Add benchmark definitions to packages


marc-gr:
In order to keep adding benchmarking capabilities to `elastic-package`, a definition for benchmarking scenarios would be useful to define and run end-to-end benchmarks for packages.

An initial simple benchmark definition would consist of:

- The target package version
- The target data-stream
- The variables to set up the policy
- Some basic settings to control warm-up and ramp-up (similar to [the ones available in rally](https://esrally.readthedocs.io/en/stable/track.html#schedule))
- Settings to define corpora sources (ie. local or S3 to begin with)
- Settings to define how the corpora logs will be sent to the agents (initially we can leverage the [stream tool](https://github.com/elastic/stream) that is currently used for system tests)

A tentative spec definition could look like:

`spec.yml`
```yaml
spec:
  additionalContents: false
  developmentFolder: true
  contents:
    - description: Folder containing end-to-end benchmarks scenarios.
      type: folder
      name: e2e
      required: false
      additionalContents: true
      contents:
        - description: Folder containing docker compose files for benchmarks.
          type: folder
          name: deploy
          additionalContents: true
          required: false
          contents:
            - description: Docker Compose configuration file for deploying benchmarks additional services.
              type: file
              name: "docker-compose.yml"
              contentMediaType: "application/x-yaml"
              required: true
        - description: A benchmark scenario.
          type: file
          pattern: '^.+\.yml$'
          contentMediaType: "application/x-yaml; require-document-dashes=true"
          required: false
          $ref: './scenario.spec.yml'
```

`scenario.spec.yml`
```yml
spec:
  type: object
  additionalProperties: false
  properties:
    version:
      description: The package version to benchmark.
      required: true
    data_stream:
      description: The data-stream to benchmark.
      required: true
    description:
      description: A description of the benchmark scenario.
      required: true
    corpora:
      description: Corpora settings.
      required: true
      type: object
      additionalProperties: false
      properties:
        path:
          description: The local path to the corpora.
        s3_bucket:
          description: The S3 bucket containing the corpora.
    input_service:
      description: The service (found in ./deploy) that will send corpora logs to the elastic agents.
    warmup-time-period:
      description: A time period in seconds for warmup of the benchmark. All response data captured during warmup will not show up in the measurement results.
```

aspacca:
> ```yaml
>     corpora:
>       description: Corpora settings.
>       required: true
>       type: object
>       additionalProperties: false
>       properties:
[...]

PR #492, opened 2023-03-21
Initial e2e benchmarks spec

marc-gr:
## What does this PR do?

Initial spec for end to end integration benchmarks. Only adds the basics to have an initial implementation.

## Checklist

## Related issues

- Closes https://github.com/elastic/package-spec/issues/483


spec/integration/_dev/benchmarks/scenario.spec.yml
```diff
@@ -0,0 +1,24 @@
...
+  type: object
+  additionalProperties: false
+  required:
+    - version
+    - data_stream
+    - description
+    - corpora
+  properties:
+    version:
+      description: The package version to benchmark.
```
mrodm, line 11: For this field, it can be added also a pattern to match the value.
For instance in `version` field in the main manifest, there is an example of pattern (and examples) that could be used:

https://github.com/elastic/package-spec/blob/main/spec/integration/manifest.spec.yml#L212

Would it make sense to use `$ref` to point to the same definition as in that manifest.spec.yml ?

spec/integration/_dev/benchmarks/spec.yml
```diff
@@ -0,0 +1,33 @@
...
+              required: false
+              contents:
+                - description: Docker Compose configuration file for deploying benchmarks additional services.
+                  type: file
+                  name: "docker-compose.yml"
+                  contentMediaType: "application/x-yaml"
+                  required: true
+        - description: A benchmark scenario.
+          type: file
+          pattern: '^.+\.yml$'
```
mrodm, line 30: As in the test package, I see that the files is named `tcp-benchmark.yml`.
Would it be required some kind of prefixes or suffixes in these files ? Or could they have any name?

code/go/pkg/validator/validator_test.go
```diff
@@ -63,8 +63,9 @@ func TestValidateFile(t *testing.T) {
 				"field icons.0.src: relative path is invalid, target doesn't exist or it exceeds the file size limit",
 			},
 		},
-		"input_template": {},
-		"input_groups":   {},
+		"integration_benchmarks": {},
```
mrodm, line 66: :+1:

spec/integration/_dev/benchmarks/spec.yml
marc-gr: I was not thinking about limiting that to any particular naming conventions, do you think could be benefficial?

spec/integration/_dev/benchmarks/scenario.spec.yml
```diff
@@ -0,0 +1,30 @@
...
+      additionalProperties: false
+      required:
+        - input_service
+      properties:
+        input_service:
+          description: The service (found in ./deploy/docker) that will send corpora logs to the elastic agents.
+          type: string
+    warmup_time_period:
+      description: A time period in seconds for warmup of the benchmark. All response data captured during warmup will not show up in the measurement results.
+      type: integer
```
mrodm, line 30: This type does not matches the values given in the test packages (strings):
```yaml
warmup_time_period: 1m
```

If we set string as a type, probably it would be better some kind of regex to allow just duration format (1m, 1s, 1h... ?).

Or should it be updated the test package?

spec/integration/_dev/benchmarks/spec.yml
mrodm: I just wanted to check whether or not they should follow a pattern. If it is not needed, I think we could keep it as it is :+1:

spec/integration/_dev/benchmarks/scenario.spec.yml
```diff
@@ -0,0 +1,30 @@
...
+  type: object
+  additionalProperties: false
+  required:
+    - version
+    - data_stream
+    - description
+    - corpora
+  properties:
+    version:
+      $ref: "../integration/manifest.spec.yml#/definitions/version"
```
mrodm, line 11: This relative path for this reference should be:

```suggestion
      $ref: "../../manifest.spec.yml#/definitions/version"
```

spec/integration/_dev/benchmarks/scenario.spec.yml
```diff
@@ -0,0 +1,24 @@
...
+    description:
+      description: A description of the benchmark scenario.
+    corpora:
+      description: Corpora settings.
+      type: object
+      additionalProperties: false
+      properties:
+        input_service:
+          description: The service (found in ./deploy/docker) that will send corpora logs to the elastic agents.
+    warmup_time_period:
```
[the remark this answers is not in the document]
mrodm: So, would it be like adding as default value 0?

```yaml
warmup_time_period:
  type: integer
  default: 0
```
Depending on the type for this field (according to test packages) it could be:
```yaml
warmup_time_period:
  type: string
  default: 0s
```

It could also be added this behavior into the description of the field too.

mrodm (approved):
Thanks for addressing all the comments, LGTM!

mrodm (approved):

PR #503, opened 2023-04-19
Profiling: add data stream type

inge4pres:
- data-streams: add profiling type
- test case for profiling data streams

## What does this PR do?

We add a new data stream type `profiling` to allow creating dedicated data streams for Universal Profiling.

## Why is it important?

We are going to use it in Cloud for the Universal Profiling symbolization service.

## Checklist

- [X] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.
- [X] I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).


jsoriano (commented):
There was a discussion about using the new profiling vs. the existing traces, and there it was also commented that this kind of event-like data would also fit in the logs type.

The main point of the discussion was to avoid having too many types of data streams.

Has there been some decision around this? Is it confirmed that logs cannot be used?

inge4pres:
> Has there been some decision around this? Is it confirmed that logs cannot be used?

I think so.

There is really no point in naming logs the profiling data.

jsoriano:
Change looks good, will wait for @joshdover and @kpollich regarding the addition of the new type and its implications.

kpollich (dismissed):

kpollich:
> Change looks good, will wait for @joshdover and @kpollich regarding the addition of the new type and its implications.

All good on my end.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -208,6 +208,7 @@ spec:
       - logs
       - synthetics
       - traces
+      - profiling
```
jsoriano, line 211: Btw, I think we should ensure that packages using this type use the latest version of the spec, so we should include this only for packages with format spec >= 2.7.0, for this include something like this in the `versions` section below:
```
  - before: 2.6.0
    patch:
      - op: remove
        path: /properties/type/enum/4
```

inge4pres: added in 37f1f69

Issue #39, opened 2020-08-26
Add support for runtime fields in packages


ruflin:
Soon Elasticsearch is planned to support runtime fields: https://github.com/elastic/elasticsearch/issues/59332 I assume these can not be specified currently in our fields.yml. We should add support for these fields in fields.yml and on the Kibana side be able to create the correct Elasticsearch templates and Kibana index patterns.

ruflin:
@ycombinator for the spec
@skh @neptunian For awareness for the template / index pattern creation
@ph For awareness
@andresrc We should figure out if and where we want to use this in packages in the future.

ruflin:
Adding to this, Elasticsearch supports to set the `dynamic: runtime`: https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html A package dev should be able to select this option as the default for dynamic fields. APM asked for this in the context of the apm package defaults (@simitt ).

simitt:
@elastic/apm-server feel free to pick this up for `7.15`

ruflin:
We would really like to see this prioritised. The way I would propose this to be implemented is as following:

```
    - name: foo.bar
      type: long
      runtime: true
```

This is slightly different from the way it [looks in Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-mapping-fields.html) where a separate block for runtime fields has to be specified. Having it directly as a boolean in the fields definition allows a very simple way to enable / disable it for devs. Fleet still has to create the additional block.

amitkanfer:
@jlind23 putting this on your radar.

joshdover:
A couple things to consider here. Runtime fields is mostly optimized for extracting new fields out of other indexed fields that are available in the "doc values" and the runtime field's script has to be written differently depend on whether or not the field that we're extracting from is being indexed. So there's really 2 use cases here:

- Runtime fields that have supply a custom script to extract data from another _indexed_ field
- Runtime fields that simply populate a value from `_source` that is not indexed
	- This is the use case that is covered by @ruflin's comment in https://github.com/elastic/package-spec/issues/39#issuecomment-1363711624

When Fleet creates the index templates for runtime fields, the `script` field for a runtime field will need to differ based on whether or not the field is indexed. If `index: false` is specified or the field is not mapped at all, the field will not be available in `doc['foo']` but will be available in `params._source['foo']`.

We can probably support a design like this for both use cases:
```
- name: foo.bar
  type: long
[...]

PR #505, opened 2023-04-19
Add support to define runtime fields

mrodm:
## What does this PR do?

Add support to define [runtime fields](https://www.elastic.co/guide/en/elasticsearch/reference/8.7/runtime-mapping-fields.html).

In Kibana versions that do not support runtime fields, it is created a mapping with the given type. Because of that, as there is no failure when installing a package that contains a runtime field in those kibana versions, a new validation rule has been introduced to ensure that the packages must have at least a given kibana version.

Here it is the issue to add this support in Kibana https://github.com/elastic/kibana/issues/155255

## Checklist

- [x] Add validation rule to allow this new `runtime` key to be used from kibana.version 8.9?

## Related issues

- Relates #39


code/go/pkg/validator/validator_test.go
```diff
@@ -393,6 +403,45 @@ func TestValidateWarnings(t *testing.T) {
 		"sql_input": []string{
 			"conditions.kibana.version must be ^8.8.0 or greater for non experimental input packages (version > 1.0.0)",
 		},
+		"bad_runtime_kibana_version": []string{
+			"conditions.kibana.version must be ^8.9.0 or greater to include runtime fields",
+		},
+	}
```
mrodm, line 409: Created a new test function to check minimum kibana versions.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -52,11 +58,25 @@ func validateMinimumKibanaVersion(packageType string, packageVersion semver.Vers
 		return nil
 	}
 
-	if kibanaVersionConditionIsGreaterThanOrEqualTo8_8_0(kibanaVersionCondition) {
+	if kibanaVersionConditionIsGreaterThanOrEqualTo(kibanaVersionCondition, "8.8.0") {
 		return nil
 	}
 
-	return errors.New("Warning: conditions.kibana.version must be ^8.8.0 or greater for non experimental input packages (version > 1.0.0)")
+	return errors.New("conditions.kibana.version must be ^8.8.0 or greater for non experimental input packages (version > 1.0.0)")
```
mrodm, line 65: Removed `Warning: ` prefix since this error is always raised and fails the validation.

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
```diff
@@ -52,11 +58,25 @@ func validateMinimumKibanaVersion(packageType string, packageVersion semver.Vers
...
+}
+
+func validateMinimumKibanaVersionRuntimeFields(fsys fspath.FS, packageVersion semver.Version, kibanaVersionCondition string) error {
+
+	errs := validateFields(fsys, validateNoRuntimeFields)
+	if len(errs) == 0 {
+		return nil
+	}
+
+	if kibanaVersionConditionIsGreaterThanOrEqualTo(kibanaVersionCondition, "8.9.0") {
```
mrodm, line 75: As there is no failure if a package that contains a runtime field is installed in kibana versions that do not support those kind of fields, this validation rule ensures that the packages must have at least a given kibana version.

What version should I set here @kpollich ? I was thinking that maybe it is 8.9 ?

code/go/internal/validator/semantic/types.go
```diff
@@ -21,6 +23,74 @@ const dataStreamDir = "data_stream"
 
 type fields []field
 
+type runtime struct {
```
jsoriano, line 26: Nit. This is a field, not a runtime :)
```suggestion
type runtimeField struct {
```

code/go/internal/validator/semantic/types.go
```diff
@@ -21,6 +23,74 @@ const dataStreamDir = "data_stream"
...
+	switch v := alias.(type) {
+	case bool:
+		r.enabled = v
+	case string:
+		r.enabled = true
+		r.script = v
+	default:
+		// JSONSchema already checks about the type of this field (e.g. int or float)
+		r.enabled = true
+		r.script = string(data)
```
jsoriano, line 83: I think we should be on the safe side and return an error here for unexpected types, without assuming that this has always passed first through the json schema validator.

If we want to support numbers, I would add a case for them.
```suggestion
	case string:
		r.enabled = true
		r.script = v
	case float64:
		r.enabled = true
		r.script = string(data)
	default:
		return fmt.Errorf("unexpected type....
```
(json package always uses float64 for numbers, except when `UseNumber()` is used).

jsoriano (approved):
A couple of nitpicks, but looks good :+1:

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -416,6 +430,27 @@ spec:
...
+          properties:
+            type:
+              enum:
+                - boolean
+                - date
+                - double
+                - geo_point
+                - ip
+                - keyword
+                - long
```
mrodm, line 443: This list is not the complete list available.
According to [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-mapping-fields.html), the complete list is:
- boolean
- composite
- date
- double
- geo_point
- ip
- keyword
- long
- [lookup](https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-retrieving-fields.html#lookup-runtime-fields)

In this condition, just the [types that we support in the `type` field](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/fields/fields.spec.yml#L42) are added. Missing types:
- composite
- lookup

code/go/internal/validator/semantic/types.go
mrodm: TBH not sure if we need to support numbers (e.g. float64), just added as a test case for completeness.

Returning that error in the default case makes the two implementations for unmarshaling YAML and JSON different, at least as they are now. `UnmarshalYAML` should be able to differentiate between strings and float64, int... etc, but it only receives a `value.Value` that is a string.

It could be added to the UnmarshalYAML something like (or viceversa allowing it):
```
       if _, err := strconv.ParseFloat(text, 8); err == nil {
               return fmt.Errorf("unexpected type found float64")
       }
```
And other should also be checked I think.

A side-effect of introducing that error, it also adds an error for each JSON schema error too:
```
    validator_test.go:173: 
        	Error Trace:	/home/mariorodriguez/Coding/work/package-spec/code/go/pkg/validator/validator_test.go:173
        	Error:      	"found 10 validation errors:
        	            	   1. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: field 0: Must not be present
        	            	   2. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: field 1.runtime: Invalid type. Expected: string, given: integer
        	            	   3. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: field 2: Must not be present
        	            	   4. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: field 3: Must not be present
        	            	   5. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: can't unmarshal fields: yaml.Unmarshal failed (path: data_stream/foo/fields/fields.yml): unexpected type found float64
        	            	   6. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: can't unmarshal fields: yaml.Unmarshal failed (path: data_stream/foo/fields/fields.yml): unexpected type found float64
        	            	   7. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: can't unmarshal fields: yaml.Unmarshal failed (path: data_stream/foo/fields/fields.yml): unexpected type found float64
        	            	   8. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: can't unmarshal fields: yaml.Unmarshal failed (path: data_stream/foo/fields/fields.yml): unexpected type found float64
        	            	   9. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: can't unmarshal fields: yaml.Unmarshal failed (path: data_stream/foo/fields/fields.yml): unexpected type found float64
        	            	  10. file "../../../../test/packages/bad_runtime_fields/data_stream/foo/fields/fields.yml" is invalid: can't unmarshal fields: yaml.Unmarshal failed (path: data_stream/foo/fields/fields.yml): unexpected type found float64
        	            	" should have 4 item(s), but has 10
        	Test:       	TestValidateFile/bad_runtime_fields
```


It looks that it would be too many errors to be shown to the user/developer for the same field wrong type:
- There are 4 wrong fields in the `bad_runtime_fields` package: `data_stream/foo/fields/fields.yml`.
- The error is shown for every time that YAML file is parsed.

An option for that would be check for duplicated errors and just show once that error:
```diff
--- code/go/internal/validator/spec.go
+++ code/go/internal/validator/spec.go
@@ -93,16 +93,24 @@ func processErrors(errs ve.ValidationErrors) ve.ValidationErrors {
                "Must validate all the schemas (allOf)",
                "Must validate at least one schema (anyOf)",
        }
+
+       duplicated := make(map[string]struct{})
        for _, e := range errs {
                for _, msg := range msgTransforms {
+                       if _, ok := duplicated[e.Error()]; ok {
+                               continue
+                       }
                        if strings.Contains(e.Error(), msg.original) {
-                               processedErrs = append(processedErrs, errors.New(strings.Replace(e.Error(), msg.original, msg.new, 1)))
+                               transformedErr := errors.New(strings.Replace(e.Error(), msg.original, msg.new, 1))
+                               processedErrs = append(processedErrs, transformedErr)
+                               duplicated[transformedErr.Error()] = struct{}{}
                                continue
                        }
                        if substringInSlice(e.Error(), redundant) {
                                continue
                        }
                        processedErrs = append(processedErrs, e)
+                       duplicated[e.Error()] = struct{}{}
                }
        }
```

Given that, I have some doubts @jsoriano :
- should we just accept strings for now?
- if other types are not allowed, in `unmarshalYAML` should we check for int, unsigned, ... too ? 
- If that error is raised, WDYT about removing duplicate error messages?

jsoriano: Ok, I wouldn't add complexity here to handle special cases. I would be fine with accepting only strings and bools. There would be any valid cases for numbers here?

mrodm: I think that the scripts defined in runtime fields usually won't set just a number. Adding just a number for a field it could be also done in the ingest pipeline too.

Just tested adding runtime field with just a number in the script field using this `index_template`:
```yaml
elasticsearch:
  index_template:
    mappings:
      runtime:
        foo:
          type: keyword
          script:
            source: "42"
``` 

And there is an error when trying to install the package:
```
2023/04/25 16:35:49 DEBUG POST https://127.0.0.1:5601/api/fleet/epm/packages
Error: can't install the package: can't install the package: could not zip-install package; API status code = 500; response body = {"statusCode":500,"error":"Internal Server Error","message":"mapper_parsing_exception\n\tCaused by:\n\t\tscript_exception: compile error\n\tRoot causes:\n\t\tscript_exception: compile error"}
```

Tested with the Elastic stack versions 8.8.0-SNAPSHOT and 8.0.0 

As there will be compile errors, I guess we could just check bools or everything else being a string, couldn't we?

jsoriano: Yes, let's consider that this is a bool or a script, and rely on script compilation for the rest.

jsoriano (approved):

code/go/internal/validator/semantic/validate_minimum_kibana_version.go
mrodm: Talked offline with @kpollich , Elastic stack / Kibana `8.9.0` version is going to be set as minimum version for runtime fields.

jsoriano (approved):

kpollich (approved):

PR #507, opened 2023-04-20
Mark profiling type as technical preview

jsoriano:
New data stream type added in https://github.com/elastic/package-spec/pull/503 is intended for a package that may never be released as GA. Set this type as technical preview so we can also remove it if not needed anymore in the future.


test/packages/profiling_symbolizer/data_stream/example/manifest.yml
```diff
@@ -1,8 +1,13 @@
...
+  - input: logfile
+    title: Sample logs
+    description: Collect sample logs
+    vars:
+      - name: paths
+        type: text
+        title: Paths
+        multi: true
+        default:
+          - /var/log/*.log
```
inge4pres, line 13: Not sure why we are removing this?

mrodm (approved):
LGTM

test/packages/profiling_symbolizer/data_stream/example/manifest.yml
jsoriano: I recreated the package from scratch to fill some missing files and forgot to recover this content. Fixed.

mrodm (approved):

PR #512, opened 2023-04-24
Extend system benchmarks spec

marc-gr:
## What does this PR do?

After tinkering with the spec, I found the original definition was limited, so this is an iteration. It also changes naming from `e2e` to `system` for consistency with tests.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-package/pull/1232
- Relates https://github.com/elastic/package-spec/issues/483


spec/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/509
+  - description: Improve system benchmarks definition
```
jsoriano, line 10: Could we describe here how the definition is being improved? It seems that relocates some files, could this be mentioned here?

spec/integration/_dev/benchmark/spec.yml
```diff
@@ -2,9 +2,9 @@ spec:
   additionalContents: false
   developmentFolder: true
   contents:
-    - description: Folder containing end-to-end benchmarks scenarios.
+    - description: Folder containing system benchmarks scenarios.
       type: folder
-      name: e2e
+      name: system
```
jsoriano, line 7: This change will be breaking, did we mark this feature as beta/experimental? Is it currently used?
To avoid breaking changes we could duplicate the definition and keep both working...

jsoriano (commented):
We should be more specific in the changelog about the improvements done. Also, this change seems to be breaking. I guess it is ok though because this hasn't been implemented yet in elastic-package and nobody will be using it.

spec/integration/_dev/benchmark/spec.yml
marc-gr: No currently used, it should be alright to break it as it is first used in https://github.com/elastic/elastic-package/pull/1232

spec/integration/_dev/benchmark/scenario.spec.yml
```diff
@@ -0,0 +1,134 @@
...
+          required:
+            - name
+          properties:
+            name:
+              description: The name of the input service.
+              type: string
+            signal:
+              description: The signal to send to the service once the benchmark is ready to start.
+              type: string
+              example: SIGHUP
```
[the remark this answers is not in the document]
marc-gr: It defaults to empty, it is just an example

spec/integration/_dev/benchmark/scenario.spec.yml
```diff
@@ -0,0 +1,134 @@
...
+      $ref: "../../manifest.spec.yml#/definitions/version"
+    input:
+      description: The input of the package to benchmark.
+      type: string
+    vars:
+      description: The package level variables.
+      type:
+        - object
+        - "null"
+      additionalProperties: true
```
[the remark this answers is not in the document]
marc-gr: Isn't `null` the default by default? 😸  If in doubt I can make it explicit in both `vars` and `data_stream.vars`.

mrodm (approved):
LGTM

mrodm (approved):
:+1:

Issue #499, opened 2023-04-12
Add `Advanced Analytics (UEBA)` sub-category to integrations


susan-shu-c:
## Overview

For customer navigation, we'd like to add a new subcategory to integrations, since having all these packages under `Security` (which has 175 packages) makes it hard to discover.

Category name: `Security` > `Advanced Analytics (UEBA)` 

Related packages 
* ProblemChild (Living off the Land Detection)
* DGA
* DED (Data Exfiltration Detection)
* LMD (Lateral Movement Detection)

We can make the PRs for these in the `integrations` repo after this issue/ticket is resolved.

# Business & User Value:

- [ ] Users need a convenient view of all ML-based advanced detection packs within their space. We are building additional integration packages under this category. The category will significantly reduce user efforts, reduce user error, and aid in feature discovery.
- [ ] Elastic security (Entity Analytics Onboarding) workflow will leverage this filtered view in the user journey for discovering these packages. Without a reconciled view, the onboarding workflow will remain incomplete.

# Implementation tasks
- [x] Add to package-spec ([ref PR](https://github.com/elastic/package-spec/pull/471))
* https://github.com/elastic/package-spec/pull/515
- [x] Add to package-registry ([ref PR](https://github.com/elastic/package-registry/pull/953))
* https://github.com/elastic/package-registry/pull/997
- [x] Add to kibana ([ref PR](https://github.com/elastic/kibana/pull/135758))
* https://github.com/elastic/kibana/pull/155935
- [x] Release new version of package spec (contact @mrodm)
- [x] Release new version of elastic-package (contact @mrodm)
- [x] Release new version of package registry (contact @mrodm)

# Mockup:
![Screenshot 2023-04-12 at 9 22 19 AM](https://user-images.githubusercontent.com/23287722/231551128-93ffecb4-2d4d-4a18-822d-54451487b883.png)

## Related tickets

* https://github.com/elastic/integrations/pull/5123

susan-shu-c:
@jen-huang mentioned that this should be a subcategory under Security, which would then look like this ([source](https://github.com/elastic/ingest-dev/issues/1349)):

<img width="1030" alt="Screenshot 2023-04-12 at 3 30 50 PM" src="https://user-images.githubusercontent.com/23287722/231564947-a5087ec1-3568-45f5-90f3-78ae758e0b6d.png">

[...]

PR #515, opened 2023-04-26
Add `Advanced Analytics (UEBA)` package subcategory under `Security` category

susan-shu-c:
## What does this PR do?

* https://github.com/elastic/package-spec/issues/499

## Checklist

## Related issues

-

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -7,6 +7,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/509
+  - description: Add new subcategories
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/515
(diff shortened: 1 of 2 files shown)


spec/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/509
+  - description: Add new subcategories
```
jsoriano, line 10: Given that we are adding a single subcategory, could we be more specific here?
```suggestion
  - description: Add new subcategory "advanced_analytics_ueba"
```

susan-shu-c: Thanks for the suggestion!

jsoriano (approved):
:rocket:

PR #513, opened 2023-04-25
Disallow external fields without _dev/build/build.yml file

mrodm:
## What does this PR do?

Add a new semantic validation to avoid fields with the external key defined if there is no `_dev/build/build.yml` file (where the ECS reference is defined).

If any field has defined the `external` key, this new validation rule:
- checks that `_dev/build/build.yml` exists
- checks that `_dev/build/build.yml` contains a definition for the external value set in the definition:
    - currently, the only supported value is `ecs`

Example:
1. Go to some package that contains external fields (e.g. elastic_package_registry)
2. Remove `_dev/build/build.yml` file
3. Run `elastic-package build -v`
4. First, it fails becuase it cannot render the README file due to the `{{ fields metrics }}`
5. Remove that line from `_dev/build/docs/README.md`
   ```diff
   diff --git packages/elastic_package_registry/_dev/build/docs/README.md packages/elastic_package_registry/_dev/build/docs/README.md
   index 84a915bd8..c7190745f 100644
   --- packages/elastic_package_registry/_dev/build/docs/README.md
   +++ packages/elastic_package_registry/_dev/build/docs/README.md
   @@ -45,7 +45,6 @@ Elastic Package Registry can provide Prometheus metrics in the `/metrics` endpoi
    You can verify that metrics endpoint is enabled by making an HTTP request to
    `http://localhost:9000/metrics` on your package registry instance.
    
   -{{ fields "metrics" }}
    
    #### Example
    
   ```
7. Run `elastic-package build -v`
```shell
 $ elastic-package build -v
2023/04/26 16:17:00 DEBUG Enable verbose logging
...
Error: building package failed: invalid content found in built zip package: found 8 validation errors:
   1. file "/home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip/data_stream/metrics/fields/ecs.yml" is invalid: field ecs.version with external key defined ("ecs") but no _dev/build/build.yml found
   2. file "/home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip/data_stream/metrics/fields/ecs.yml" is invalid: field error.message with external key defined ("ecs") but no _dev/build/build.yml found
   3. file "/home/mariorodriguez/Coding/work/integrations/build/packages/elastic_package_registry-0.0.6.zip/data_stream/metrics/fields/ecs.yml" is invalid: field event.duration with external key defined ("ecs") but no _dev/build/build.yml found
[...]

PR #516, opened 2023-04-28
Validate fields input packages

### LINKED ISSUE #511: fields in input packages are not validated
Currently, the validation helper to check fields relies on datastreams and therefore just checks integration packages:

- https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/types.go#L38
- https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/types.go#L86

It expects that fields files are under `/package/datastream/name/fields/*.yml`. However , input packages define the fields file in the package root folder directly: `/package/fields/*.yml`

For instance, validation does not fail when adding a field definition in a input package (e.g. `test/packages/good_input`) with an error like:
```yaml
- name: date_example
  type: boolean
  date_format: yyyy-MM-dd
```

This is used in other validation rules like:
- [validate fields limits](https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/validate_fields_limits.go)
- [validate required fields](https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/validate_required_fields.go)
- [validate unique fields](https://github.com/elastic/package-spec/blob/main/code/go/internal/validator/semantic/validate_unique_fields.go)

mrodm:
## What does this PR do?

This PR adds field validation also for input packages.

- Added key `fieldsPerDataStreamLimit` into input package spec.
- Filter rules applied depending on what type of package is being validated.
- Add the same required fields for input packages as in integration packages:
    - `data_stream.type`
    - `data_stream.dataset`
    - `data_stream.namespace`
    - `@timestamp`

## Why is it important?

Currently, input packages are missing those checks/validation rules that are performed using `validateFields` helper. This helper needs to manage both fields files defined under data_stream folders and also fields files in input packages.

Example using `sql_input` package from integrations:
- Before
  ```
   $ cd packages/sql_input
   $ elastic-package check 
  Format the package
  Done
  Lint the package
  Done
  Build the package
  2023/04/28 09:52:41  INFO License text found in "/home/mariorodriguez/Coding/work/integrations/LICENSE.txt" will be included in package
  Package built: /home/mariorodriguez/Coding/work/integrations/build/packages/sql-0.0.1.zip
  Done
  ``` 
- Now
  ```shell
   $ cd packages/sql_input
   $ elastic-package check
  Format the package
  Done
  Lint the package
  Error: checking package failed: linting package failed: found 4 validation errors:
     1. expected field "data_stream.type" with type "constant_keyword" not found in input package
     2. expected field "data_stream.dataset" with type "constant_keyword" not found in input package
     3. expected field "data_stream.namespace" with type "constant_keyword" not found in input package
     4. expected field "@timestamp" with type "date" not found in input package
  ```

## Checklist

## Related issues

- Closes #511


mrodm:
As expected, there are some input packages that fail because of this. Errors can be checked in this PR https://github.com/elastic/package-spec/pull/516

Mainly the errors is because of the validation rule that checks for required fields.

Example of failure in `gcp_metrics`:

```
[2023-04-28T08:44:23.242Z] + ../../build/elastic-package check -v
[2023-04-28T08:44:23.343Z] elastic-package has been installed.
[2023-04-28T08:44:23.343Z] 2023/04/28 08:44:23 DEBUG Enable verbose logging
[2023-04-28T08:44:23.443Z] Format the package
[2023-04-28T08:44:23.444Z] Done
[2023-04-28T08:44:23.444Z] Lint the package
[2023-04-28T08:44:23.444Z] Error: checking package failed: linting package failed: found 4 validation errors:
[2023-04-28T08:44:23.444Z]    1. expected field "data_stream.type" with type "constant_keyword" not found in input package
[2023-04-28T08:44:23.444Z]    2. expected field "data_stream.dataset" with type "constant_keyword" not found in input package
[2023-04-28T08:44:23.444Z]    3. expected field "data_stream.namespace" with type "constant_keyword" not found in input package
[2023-04-28T08:44:23.444Z]    4. expected field "@timestamp" with type "date" not found in input package
[2023-04-28T08:44:23.444Z] 
```

For instance, these input packages are defining those mappings:
- [custom_log](https://github.com/elastic/integrations/blob/main/packages/log/fields/base-fields.yml)
- [cel](https://github.com/elastic/integrations/blob/main/packages/cel/fields/input.yml)

Should these mappings be required also in input packages ? @hop-dev @kpollich @jsoriano 
I would say that they should also be required for input packages, because at the end those fields would be in the documents. WDYT ?

jsoriano:
> Should these mappings be required also in input packages ? @hop-dev @kpollich @jsoriano
> I would say that they should also be required for input packages, because at the end those fields would be in the documents. WDYT ?

+1, we can fix these packages when the integrations repo is updated with an elastic-package that uses this spec.

kpollich:
> Should these mappings be required also in input packages ? @hop-dev @kpollich @jsoriano
I would say that they should also be required for input packages, because at the end those fields would be in the documents. WDYT ?

Yes I think input packages should still define these mappings in a `fields.yml` file even if they're sourced from the policy editor. They'll still be present in documents ingested, so we should assert that they're mapped. 

@hop-dev feel free to correct me if I'm off base with my understanding of input packages here.

code/go/internal/validator/spec.go
```diff
@@ -138,12 +139,26 @@ func (s Spec) rules(rootSpec spectypes.ItemSpec) validationRules {
...
+			continue
+		}
+
 		validationRules = append(validationRules, rule.fn)
 	}
 
 	return validationRules
 }
 
+func contains(elems []string, v string) bool {
```
jsoriano, line 153: Nit. Call this like the type it checks.
```suggestion
func stringSliceContains(elems []string, v string) bool {
```
Or use generics :slightly_smiling_face: 
```suggestion
func sliceContains[T any](elems []T, v T) bool {
```

Or maybe we can use this library that has chances of ending up in the stdlib https://pkg.go.dev/golang.org/x/exp/slices#Contains

jsoriano (dismissed):
:+1:

code/go/internal/validator/spec.go
mrodm: Trying the generics option, there is an error that I don't see why it is a "set":
> internal/validator/spec.go:155:6: invalid operation: a == v (incomparable types in type set)

I will go for the keep using strings (first option).

jsoriano: Ah, it seems it shouldn't be `any`, but `comparable`, sorry.
```
func sliceContains[T comparable](elems []T, v T) bool {
```

jsoriano (dismissed):

mrodm:
Update the PR testing in integrations with the latest changes 
https://github.com/elastic/integrations/pull/6026

Due to the update with the upstream `main` branch, it will include also validation in input packages to disallow external fields without `_dev/build/build.yml` (https://github.com/elastic/package-spec/pull/513).

jsoriano (commented):
This "id" concept looks a bit weird to me, and I think that maybe we don't need it. For input packages we could very well use the empty string in `metadata.dataStream` for grouping, and rely on it to format the errors.

code/go/internal/validator/semantic/validate_fields_limits.go
```diff
@@ -20,18 +20,15 @@ func ValidateFieldsLimits(limit int) func(fspath.FS) ve.ValidationErrors {
...
 
-		dataStream, err := dataStreamFromFieldsPath(fsys.Path(), fieldsFile)
-		if err != nil {
-			return ve.ValidationErrors{err}
-		}
-		count, _ := counts[dataStream]
-		counts[dataStream] = count + 1
+		id := metadata.ID()
+		count, _ := counts[id]
+		counts[id] = count + 1
```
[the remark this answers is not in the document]
mrodm: Sure, no need to use `iD()` here. Thanks!

code/go/internal/validator/semantic/types.go
```diff
@@ -33,41 +34,50 @@ type field struct {
...
 	}
 
 	var vErrs ve.ValidationErrors
-	for _, fieldsFile := range fieldsFiles {
-		unmarshaled, err := unmarshalFields(fsys, fieldsFile)
+	for _, metadata := range fieldsFilesMetadata {
+		unmarshaled, err := unmarshalFields(fsys, metadata.filePath)
 		if err != nil {
-			vErrs = append(vErrs, errors.Wrapf(err, `file "%s" is invalid: can't unmarshal fields`, fsys.Path(fieldsFile)))
+			vErrs = append(vErrs, errors.Wrapf(err, `file "%s" is invalid: can't unmarshal fields`, fsys.Path(metadata.filePath)))
```
jsoriano, line 58: Could we use `metadata.fullFilePath` here?
```suggestion
			vErrs = append(vErrs, errors.Wrapf(err, `file "%s" is invalid: can't unmarshal fields`, metadata.fullFilePath))
```

code/go/internal/validator/semantic/types.go
```diff
@@ -33,41 +34,50 @@ type field struct {
 	Fields fields `yaml:"fields"`
 }
 
-type validateFunc func(fieldsFile string, f field) ve.ValidationErrors
+type fieldFileMetadata struct {
+	packageType  string
+	packageName  string
```
jsoriano, line 39: `packageType` and `packageName` are not used at the end, right? Do we need them? And `filePath` is only used once, and maybe we can use `fullFilePath` there.

jsoriano (dismissed):

code/go/internal/validator/semantic/types.go
mrodm: At first, I wanted to ensure that the same errors were provided.
Now, that it has been checked that, I think this could be updated as you suggest.

code/go/internal/validator/semantic/types.go
mrodm: Sure, I'll check that, and I'll remove them if they are not needed now. :+1:

code/go/internal/validator/semantic/types.go
mrodm: Doing some tests, `metadata.filePath` is required to unmarshal the fields. It requires the path inside the package to the fields file.

```go
		unmarshaled, err := unmarshalFields(fsys, metadata.filePath)
		if err != nil {
			vErrs = append(vErrs, errors.Wrapf(err, `file "%s" is invalid: can't unmarshal fields`, metadata.fullFilePath))
		}
```

code/go/internal/validator/semantic/types.go
mrodm: Removed `pacakgeType` and `packageName` fields

jsoriano (approved):

PR #524, opened 2023-05-15
add rally benchmkars spec

aspacca:
## What does this PR do?
Add spec for rally benchmark, following the spec of #512 

## Checklist

## Related issues

-


aspacca:
@mrodm 
I have a note regarding: https://github.com/elastic/package-spec/blob/8a67b8465f6949dfb0dae5eb4abb0a52da46096b/spec/integration/_dev/corpora.spec.yml#L22-L25

I've aligned with several stakeholders for the corpus generator (including @marc-gr) and I'm going to change the way the total amount of events will be generated: not anymore based on size (as string) but with the exact number of events (as integer)

What is the best approach here from your point of view?
Include the changes in this PR, or merge this as it this and then create a new one after implementing the changes in the corpus generator?

I suspect that `package-spec` -> `elastic-package` (meaning definition of the specs and any change to them will come before they will supported in elastic package), but what about breaking specs? there will be a period where `elastic-package` will not be in line with the new `package-spec`. is it everything fine as long we don't cut a precise version for both projects?

please, let me know. thanks

mrodm (commented):
This PR. Just some comments since it looks like some properties could be defined and then reused in the new file for rally benchmarks.

mrodm:
> What is the best approach here from your point of view? Include the changes in this PR, or merge this as it this and then create a new one after implementing the changes in the corpus generator?
> 
> I suspect that `package-spec` -> `elastic-package` (meaning definition of the specs and any change to them will come before they will supported in elastic package), but what about breaking specs? there will be a period where `elastic-package` will not be in line with the new `package-spec`. is it everything fine as long we don't cut a precise version for both projects?

As this PR does not introduce any breaking change, I think it cold be merged as it is. 

In a following PR, those changed that you mention could be introduced.
About breaking changes, in a previous PR (https://github.com/elastic/package-spec/pull/512) , there was already a breaking change.

Thinking about options: 
- Could it be used another field name for that new integer ? or do you want to keep `size` but with type `integer` ? 

Assuming there will be a new field for the number of events, once the new implementation is done, create another PR here in package-spec where:
    1. `size` field is set as deprecated
    2. the new field with type `integer` is added 

That would allow to be backwards compatible.

Please @elastic/ecosystem  chime in if you see other options here.

spec/integration/_dev/benchmark/system.scenario.spec.yml
```diff
@@ -0,0 +1,65 @@
...
+      description: The package to benchmark.
+      type: string
+    description:
+      description: A description of the system benchmark scenario.
+      type: string
+    version:
+      $ref: "../../manifest.spec.yml#/definitions/version"
+    input:
+      description: The input of the package to benchmark.
+      type: string
```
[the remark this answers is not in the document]
aspacca: > Are they intended to be the same in both files ?

not intended as for the fact that they are the same.
I was not sure if extracting them on an extra spec file, or like you suggested here reference from the rally benchmark spec file to the system benchmark spec file

spec/integration/_dev/benchmark/system.scenario.spec.yml
```diff
@@ -0,0 +1,65 @@
...
+      default: ""
+      example: 10m
+    wait_for_data_timeout:
+      description: Timeout for waiting for data during a system benchmark.
+      type: string
+      default: 10m
+      example: 10m
+    corpora:
+      type: object
+      $ref: "../corpora.spec.yml#/properties/corpora"
```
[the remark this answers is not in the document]
aspacca: > To keep this, I would move this `corpora` object definition here in this file under the `definitions` key. And later, it can be referenced here and it can be referenced in the `rally.benchmark.spec.yml` file too

👍 on this

I knew the reference to `properties` was wrong and it should rather be to `definitions`, but looking quickly through the json schema docs I could not find the exact way for it

aspacca:
> Thinking about options:
> 
> * Could it be used another field name for that new integer ? or do you want to keep `size` but with type `integer` ?
> 
> Assuming there will be a new field for the number of events, once the new implementation is done, create another PR here in package-spec where: 1. `size` field is set as deprecated 2. the new field with type `integer` is added

there will be a new field, most probably `total_events`, the problem is that the v0.5.0 of the [dependency](https://github.com/elastic/elastic-integration-corpus-generator-tool) supports `size` and not `total_events`, and I was planning to bump to v0.6.0 that will do the contrary: supports `total_events` and not `size`.

if it's not a big deal I would avoid to have an intermediary version of the dependency that supports both field: I will take care of upgrading the relevant code in `elastic-package` for the migration. so fare there's only me and @marc-gr that added code that rely on the dependency

test/packages/integration_benchmarks/_dev/benchmark/rally/rally-benchmark.yml
```diff
@@ -0,0 +1,8 @@
+---
+description: Benchmark 20MiB of data ingested
+data_stream.name: test
```
ruflin, line 3: As this is a single value, does it mean the rally track will always be generated for a single dataset?

aspacca: the folder is per dataset, and the schemas are the same. so yes :)

aspacca: we could later figure out if there's the need to test "concurrent dataset ingestions" how to orchestrate them

ruflin: In the above, I assume `integration_benchmarks` is the package. It seems this is a `_dev` folder under the package level. When you say it is per dataset, is `rally` the dataset here? But then I don't get what there is a `data_stream.name`?

aspacca: @ruflin , yes, you are right, I mixed up the level of the `_dev` folder, sorry

to give more context you could reference @marc-gr's PR in `elastic-package`
https://github.com/elastic/elastic-package/pull/1232/files#diff-42b318802d9197a331edfa93fe80ac147ce9f76adbcc4aafa35a70930c3e20bdR59-R71

Like you will have from the spec already merged:
```
<package root>/
  _dev/
    benchmark/
      system/
        <scenario>.yml
```

You will have with my spec:
```
<package root>/
  _dev/
    benchmark/
      rally/
        <scenario>.yml
```

the content of every `<scenario>.yml` can vary, each of them targeting a specific `dataset` (imagine `system.cpu`). and potentially there could be multiple `<scenario>.yml` for the same `dataset` with different (well.. :)) scenarios

`rally` is not limited to a single dataset or even a single datastream, you can have a track with multiple of them, as far as I can see, but each corpus for each of them must be a separated file. specifically relating this to the corpus generator, the tool can generate one content for one single "dataset" (please, consider this as different schemas -A, B, C etc- for each dataset) at time


> But then I don't get what there is a `data_stream.name`?

in relation to the rally benchmark a `<scenario>.yml` will reference something like `data_stream.name: metrics-system.cpu-default`
it's the name of the data_stream. that's composed, according to the naming standard, by the dataset.

ruflin: So `rally` here is hardcoded and not related to package or dataset name. Inside is a list of scenarios. The part I stumbling over this that each scenario is for a single dataset: `data_stream.dataset`, so if the scenario would be inside, the config would not be needed. But having the scenario outside has the advantage, that a single scenario could be built for multiple dataset (which I think we should be able to eventually). Or is the idea here to run scenarios of scenarios?

I got confused by `data_stream.name`. First, that in your test scenario you call it `test` which is not a valid data stream naming scheme value. Second, `data_stream.name` doesn't really exist as part of the data stream naming scheme. I see this comes from the PR from marc. If we have it outside the package, shouldn't we specify `data_stream.dataset` here so we know which dataset it is for? I'll have a quick chat with @marc-gr also related to his PR, that might clarify things on my end.

aspacca: > The part I stumbling over this that each scenario is for a single dataset: `data_stream.dataset`, so if the scenario would be inside, the config would not be needed. But having the scenario outside has the advantage, that a single scenario could be built for multiple dataset (which I think we should be able to eventually). Or is the idea here to run scenarios of scenarios?

it's the same thing that confused me: indeed I'm not sure if a scenario could be built for multiple dataset, I don't see it feasible according to the current spec, and I honestly don't see a big advantage for that (it would be possible, but is it even useful? :)). so it could be the case that we can move the `rally`/`system` folders under `_dev` at dataset level.

I cannot guess why @marc-gr chose to use the package level, but I would probably have done the same, since that was already the case of the supported spec and it leads to less "folders pollution", and I would do only if the confusion you had as well cannot be avoided with better documentation :)

ruflin: Because of some historic renaming, `data_stream.name = data_stream.dataset`. We should rename it here to remove the confusion.

The discussion I had with @marc-gr is that for now we can keep it at the package level, we can always later also support it on the dataset level.

On the multi data set testing: I think there are use cases for this. Lets take k8s that has many dataset. I would want to have a rally track that ingests all dataset in parallel and not just one at the time. Just recently we have discovered that this can lead to a different behaviour in ES. I want the rally tracks to be as close to the real use case as possible.

mrodm:
> there will be a new field, most probably `total_events`, the problem is that the v0.5.0 of the [dependency](https://github.com/elastic/elastic-integration-corpus-generator-tool) supports `size` and not `total_events`, and I was planning to bump to v0.6.0 that will do the contrary: supports `total_events` and not `size`.

If `total_events` is going to be the new field, it could be added in this PR and set `size` as deprecated in this PR. cc @elastic/ecosystem 

> if it's not a big deal I would avoid to have an intermediary version of the dependency that supports both field: I will take care of upgrading the relevant code in elastic-package for the migration. so fare there's only me and @marc-gr that added code that rely on the dependency

FYI In elastic-package, updates of dependencies are managed by dependabot and merged automatically (if there are no test failures).

aspacca:
> FYI In elastic-package, updates of dependencies are managed by dependabot and merged automatically (if there are no test failures).

let me ensure that a test failure will occur, then :)

test/packages/integration_benchmarks/_dev/benchmark/rally/rally-benchmark.yml
aspacca: > Because of some historic renaming, `data_stream.name = data_stream.dataset`. We should rename it here to remove the confusion.

I'm not 100% sure about renaming in to `data_stream.dataset`
given `metrics-system.cpu-default`, `data_stream.dataset` should be `system.cpu`, while the "name of the data_stream" is indeed `metrics-system.cpu-default`

unless I missed something

mrodm:
> just the last `uint64` has a different meaning, so nothing will break: just the number of events generated will be different: I can add a unit test to enforce this so when dependabot will create the PR for the bump will fail, what do you think @mrodm ?
> 
> done at: [elastic/elastic-package#1267](https://github.com/elastic/elastic-package/pull/1267)

Yes, better to add that test so we can adapt the code in elastic-package , thanks!
Once the dependency is updated, the test should also be updated, shouldn't it ?

aspacca:
> Once the dependency is updated, the test should also be updated, shouldn't it ?

yes, the PR from dependabot will fail: I'll fetch locally the branch, do the proper changes (including the test) and push

spec/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/522
+  - description: Extend benchmarks spec to add rally scenario
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/524
```
mrodm, line 12: nit. this entry should be at the bottom

mrodm (commented):
Just one minor comment about the changelog

test/packages/integration_benchmarks/_dev/benchmark/rally/rally-benchmark.yml
ruflin: Are we putting the full name in? If yes, I kind of agree it is a different field. @marc-gr is this also the case on your end?

ruflin (approved):
Change LGTM. Approving this under the assumption we can still make "breaking" changes to this spec moving forward as this is experimental.

mrodm (approved):
LGTM :+1:

mrodm (approved):

PR #526, opened 2023-05-17
Add variants and terraform and k8s input service deployers

marc-gr:
## What does this PR do?

- Adds support for variants and terraform/k8s input service deployers
- Rename data_stream.name to data_stream.dataset

## Checklist

### DIFF
--- a/spec/integration/_dev/benchmark/spec.yml
+++ b/spec/integration/_dev/benchmark/spec.yml
@@ -6,12 +6,17 @@ spec:
       type: folder
       name: system
       required: false
-      additionalContents: true
+      additionalContents: false
       contents:
         - description: Folder containing deployable services for benchmarks.
           type: folder
           name: deploy
           $ref: "./deploy/spec.yml"
+        - description: A folder containing resources used by one or more benchmarks.
+          type: folder
+          pattern: '^[a-z0-9]+-benchmark$'
+          additionalContents: true
+          required: false
         - description: A system benchmark scenario.
           type: file
           pattern: '^.+\.yml$'


spec/integration/_dev/benchmark/deploy/tf/spec.yml
```diff
@@ -0,0 +1,26 @@
...
+    - description: Terraform template files
+      pattern: '^*.tftpl$'
+      type: file
+      required: false
+    - description: Environment variables exposed for Terraform
+      type: file
+      name: "env.yml"
+      contentMediaType: "application/x-yaml"
+      required: false
+      $ref: "./env.spec.yml"
```
mrodm, line 21: Could it be used this specification that is already in place for `_dev/deploy/tf` ?

spec/integration/_dev/benchmark/spec.yml
```diff
@@ -6,12 +6,32 @@ spec:
...
+          required: false
+          contents:
+            - description: Folder containing deployment files for a Docker-based service deployment
+              type: folder
+              name: docker
+              $ref: "./deploy/docker/spec.yml"
+            - description: Folder containing deployment files for a Terraform-based service deployment
+              type: folder
+              name: tf
+              $ref: "./deploy/tf/spec.yml"
```
mrodm, line 24: It looks the same specification than the one in `_dev/deploy/tf`.
Could it be used as a ref this one ? 
https://github.com/elastic/package-spec/blob/main/spec/integration/_dev/deploy/tf/spec.yml

spec/integration/_dev/benchmark/spec.yml
```diff
@@ -6,12 +6,32 @@ spec:
...
+            - description: Folder containing deployment files for a Kubernetes service deployment
+              type: folder
+              name: k8s
+              $ref: "./deploy/k8s/spec.yml"
```
mrodm, line 28: This definition also looks like to the one used in `_dev/deploy/k8s`. 
Could it be used this file in the `$ref` key ? 
https://github.com/elastic/package-spec/blob/main/spec/integration/_dev/deploy/k8s/spec.yml

spec/integration/_dev/benchmark/spec.yml
```diff
@@ -6,12 +6,32 @@ spec:
...
+            - description: Configuration file describing variations available for deployment
+              type: file
+              name: "variants.yml"
+              contentMediaType: "application/x-yaml"
+              required: false
+              $ref: "./deploy/variants.spec.yml"
```
mrodm, line 34: Same for this `variants.yml`. Could it be set the `$ref` key to be https://github.com/elastic/package-spec/blob/main/spec/integration/_dev/deploy/variants.spec.yml ?

spec/integration/_dev/benchmark/deploy/spec.yml
```diff
@@ -1,15 +0,0 @@
-spec:
```
mrodm, line 1: This file is still being used in the rally specification:
https://github.com/elastic/package-spec/blob/f2406bfb68971ed88c8e569cbc143c713b8bad6f/spec/integration/_dev/benchmark/spec.yml#L50

```
[2023-05-24T07:23:15.094Z]         	Error:      	Received unexpected error:
[2023-05-24T07:23:15.094Z]         	            	open integration/_dev/benchmark/deploy/spec.yml: file does not exist
[2023-05-24T07:23:15.094Z]         	            	could not open folder specification file
```

As this same specification is used for system and rally benchmarks, could it be recovered and just the same definition is used in both places?

spec/integration/_dev/benchmark/spec.yml
```diff
@@ -6,11 +6,13 @@ spec:
...
       required: false
-      additionalContents: false
+      additionalContents: true
       contents:
         - description: Folder containing deployable services for benchmarks.
           type: folder
           name: deploy
+          additionalContents: false
+          required: false
           $ref: "./deploy/spec.yml"
```
mrodm, line 16: In  `deploy/spec/.yml`, additionalContents and required are already defined.

```
spec:
  additionalContents: false
  required: true
```

what is the final value for `required` here?

test/packages/integration_benchmarks/_dev/benchmark/system/tcp-benchmark/config.yml
```diff
@@ -0,0 +1,40 @@
+- name: IP
```
mrodm, line 1: Should this file `config.yml` and `fields.yml` be added to the specification ?

Are those files required if the folder is present?

spec/integration/_dev/benchmark/spec.yml
```diff
@@ -6,11 +6,13 @@ spec:
       type: folder
       name: system
       required: false
-      additionalContents: false
+      additionalContents: true
```
mrodm, line 9: It looks like this is set to true because it needs to be added the `tcp-benchmark` folder, is that right ? 

Could there be more folders like this one under `system` or just one? 
Could it be added somehow to the specification? Maybe renaming it to a more generic name if it is just one folder expected for that.

mrodm: Probably, it could be used `pattern` to define the folders. Something like
https://github.com/elastic/package-spec/blob/a105f6aaec9460cfa5e601e7dac853871b36e6d4/spec/integration/data_stream/_dev/test/system/spec.yml#L6

test/packages/integration_benchmarks/_dev/benchmark/system/tcp-benchmark/config.yml
marc-gr: The folder is supposed to be a placeholder for any resources required by the benchmark, the idea of not making it based on the benchmark name was based on the idea that different benchmarks may want to use the same resources. So both the folder names and its contents were just not enforced any particular structure.

spec/integration/_dev/benchmark/spec.yml
marc-gr: See my previous comment

test/packages/integration_benchmarks/_dev/benchmark/system/tcp-benchmark/config.yml
mrodm: Would that mean that there could be any file inside those folders ?

marc-gr: even though I see this not being so common, yes it could be any number of them

mrodm: Talked offline, a new entry will be added to define resource folders explicitly. And those folder could contain any resource.

mrodm (approved):
:+1:

PR #530, opened 2023-05-25
Add configuration to onboard repository into Buildkite and Backstage

mrodm:
## What does this PR do?

This PR adds the needed configuration to start using Buildkite as a CI framework for this repository.

## Why is it important?

At some point, all pipelines should be running in Buildkite.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,65 @@
+# Declare a Backstage Component that represents your application.
+---
+# yaml-language-server: $schema=https://json.schemastore.org/catalog-info.json
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: package-spec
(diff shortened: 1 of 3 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,65 @@
...
+metadata:
+  name: package-spec
+  description: EPR package specifications
+  links:
+    - title: Developer Documentation
+      icon: file-doc
+      url: https://www.elastic.co/guide/en/integrations-developer/current/package-spec.html
+
+spec:
+  type: tool
```
mrodm, line 15: @alexsapran @jsoriano What should be used here ? library? spec?
https://backstage.io/docs/features/software-catalog/descriptor-format/#kind-component

alexsapran (approved):

PR #535, opened 2023-06-12
Add support to define routing rules in data streams

### LINKED ISSUE #514: [Fleet] Add support for routing rules in integrations
In order to support document-based routing in Fleet, integrations need to expose their routing rules as part of their data stream manifest files.

These rules would be translated to [`reroute` processors](https://www.elastic.co/guide/en/elasticsearch/reference/master/reroute-processor.html) appropriately by Fleet during integration installation.

A routing rule is composed of a few pieces of data:
- "Source" dataset: in which dataset should Fleet place this rule?
- "Destination" dataset: where should documents be routed by this rule?
- Condition: what logic determines when a document is routed?
- Namespace: under what namespace should this document be written once routed?

We'll need to support two types of routing rules defined by an integration:
- "Local" routing rules that route from a dataset on a given integration to other datasets on that same integration
- "Injected" routing rules that route from a dataset on a _external_ integration back to the given integration

As far as integrations are concerned, though, there is no meaningful difference between writing a "local" routing rule and an "injected" routing rule in a data stream manifest. Fleet will be responsible for generating the appropriate `processors` in the appropriate ingest pipelines based on these rules. So, the implementation on the package-spec side will be a generic `routing_rules` object at the data stream manifest level.

Typically, routing rules will be defined on a "catch-all" or "data sink" style dataset like `kubernetes.router` that is essentially a passthrough to more specific data streams.

For example, we might have an `nginx` catch-all dataset that routes Nginx logs to more specific data sets like `nginx.error` and `nginx.access` based on the logfile path reported in each document.

Here's a proposed example of the above in action. Please see the annotative comments for more details:

```yml
# nginx/data_stream/nginx/manifest.yml
title: Nginx logs
type: logs

# This is a catch-all "sink" data stream that routes documents to 
# other datasets based on conditions or variables
dataset: nginx

# Ensures agents have permissions to write data to `logs-nginx.*-*`
elasticsearch.dynamic_dataset: true
elasticsearch.dynamic_namespace: true

routing_rules:
  # "Local" routing rules are included under this current dataset, not a special case
  nginx:
    # Route error logs to `nginx.error` when they're sourced from an error logfile
    - dataset: nginx.error
      if: "ctx?.file?.path?.contains('/var/log/nginx/error')"
      namespace:
        - {{labels.data_stream.namespace}}
        - default

    # Route access logs to `nginx.access` when they're sourced from an access logfile
    - dataset: nginx.access
      if: "ctx?.file?.path?.contains('/var/log/nginx/access')"
      namespace:
        - {{labels.data_stream.namespace}}
        - default
  
  # Route K8's container logs to this catch-all dataset for further routing
  k8s.router: 
    - dataset: nginx
      if: "ctx?.container?.image?.name == 'nginx'"
      namespace:
        - {{labels.data_stream.namespace}}
        - default
   
  # Route syslog entries tagged with nginx to this catch-all dataset
  syslog:
    - dataset: nginx
      if: "ctx?.tags?.contains('nginx')"
      namespace:
        - {{labels.data_stream.namespace}}
        - default

```

Fleet support will be implemented as follow:
- Phase 1: https://github.com/elastic/kibana/issues/155910
- Phase 2: https://github.com/elastic/kibana/issues/157422

mrodm:
## What does this PR do?

This PR adds support for routing rules for data streams.
If routing_rules are defined, then `dataset` field becomes mandatory to be set too.

This new feature will be part of the following package-spec version 2.9.0 (updated changelog accordingly).

These routing rules definitions are going to be defined in a new file under each data stream folder. The file name would be `routing_rules.yml`

Example of the contents/definitions of this new file (following example in #514):
```yaml
  # "Local" routing rules are included under this current dataset, not a special case
- source_dataset:  nginx
  # Route error logs to `nginx.error` when they're sourced from an error logfile
  - target_dataset: nginx.error
    if: "ctx?.file?.path?.contains('/var/log/nginx/error')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default

  # Route access logs to `nginx.access` when they're sourced from an access logfile
  - target_dataset: nginx.access
    if: "ctx?.file?.path?.contains('/var/log/nginx/access')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default
  
# Route K8's container logs to this catch-all dataset for further routing
- source_dataset: k8s.router
  - target_dataset: nginx
    if: "ctx?.container?.image?.name == 'nginx'"
    namespace:
      - {{labels.data_stream.namespace}}
      - default
   
  # Route syslog entries tagged with nginx to this catch-all dataset
- source_dataset: syslog
  - target_dataset: nginx
    if: "ctx?.tags?.contains('nginx')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default
```

## Why is it important?

Integration packages need to expose their routing rules as part of their data stream manifest files.

## Checklist

    - Added a new data-stream in `good_v2`
    - Added a new test package `bad_routing_rules`

## Related issues

- Closes #514 

## How to test

1. Build `elastic-package` with the latest changeset from this branch. Example:
   ```
   go mod edit -replace github.com/elastic/package-spec/v2=github.com/mrodm/package-spec/v2@4c8a7e2
   go mod tidy
   make build
   ``` 
3. Update a package including routing_rules key and defining `dataset` field.
4. Build package, start an Elastic stack and installs that new package vesrion
   ```
   elastic-package build -v
   elastic-package stack down ; elastic-package stack up -v -d
   # install package using the CLI or through the UI
   elastic-package install -v 
   ```


jlind23:
@mrodm is this one ready to be reviewed?

mrodm:
> @mrodm is this one ready to be reviewed?

Not yet, pending some changes. I'm working to include the `routing_rules` key in the same data stream manifest. So it can be removed for spec versions before 2.9.0

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -307,6 +332,29 @@ spec:
...
+        required:
+          - dataset
+          - rules
+  allOf:
+    - if:
+        required:
+          - routing_rules
+      then:
+        required:
+          - dataset
```
mrodm, line 357: In case routing_rules is defined, dataset becomes a required field.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -307,6 +332,29 @@ spec:
         dynamic_namespace:
           description: When set to true, agents running this integration are granted data stream privileges for all namespaces of its type
           type: boolean
+    routing_rules:  # technical preview
```
mrodm, line 335: Tried to install a package with this new routing_rules key in 8.7.1 and it is installed successfully. No error and no reference to those processors neither.

Should we add some validation rule for this ?

spec/changelog.yml
```diff
@@ -2,17 +2,17 @@
...
-    link: https://github.com/elastic/package-spec/pull/522
   - description: Add validation for data types of metrics
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/525
   - description: Extend benchmarks spec to add rally scenario
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/524
+  - description: Add support to define routing rules in data streams (technical preview)
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/535
```
mrodm, line 15: Added `technical preview` since it is documented also that `reroute processors` is in technical preview too. https://www.elastic.co/guide/en/elasticsearch/reference/current/reroute-processor.html

Is that ok ? @jsoriano @juliaElastic

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -307,6 +332,29 @@ spec:
...
+      description: Routing rules set.
+      type: array
+      items:
+        type: object
+        additionalProperties: false
+        properties:
+          dataset:
```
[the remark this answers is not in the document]
jsoriano: The option in the processor is called `dataset`, right? Though it seems it could accept a package name.
I would use `dataset`, to be coherent with the processor, or `name` if we want to be generic enough, but maybe `name` is too generic (would it be the option passed to the processor or would it be a name for the route itself?).

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -169,6 +169,31 @@ spec:
             hidden:
               description: Makes the data stream hidden
               type: boolean
+    route:
+      description: Routing rule definition
+      type: object
+      properties:
+        dataset:
+          type: string
```
jsoriano, line 177: If this corresponds to `dataset` in the reroute processor, this can be also a list: https://www.elastic.co/guide/en/elasticsearch/reference/current/reroute-processor.html

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -169,6 +169,31 @@ spec:
             hidden:
               description: Makes the data stream hidden
               type: boolean
+    route:
```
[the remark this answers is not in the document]
mrodm: As I wanted to remove this new key `routing_rules` in case spec version is < 2.9.0, I had to do in this way (adding it to the manifest).

Adding it as a new file, I would have to remove the entry also in the `spec.yml` but I think it is not supported json patches in those files.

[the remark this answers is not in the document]
jsoriano: Ah yes, but well, we should give priority to how we want packages to look like. If this is the only reason to keep this in the manifest, lets add support for json patches in the folder spec files.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -169,6 +169,31 @@ spec:
...
+        dataset:
+          type: string
+          examples:
+          - nginx.error
+          - nginx
+        if:
+          type: string
+          examples:
+          - "ctx?.file?.path?.contains('/var/log/nginx/error')"
+          - "ctx?.container?.image?.name == 'nginx'"
```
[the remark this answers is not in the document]
mrodm: I've been doing some tests and it works as expected. 
To be considered that `if` as an `if-then-else` block, I would say it has to be at the same level (indentation) as properties key.

spec/integration/data_stream/manifest.spec.yml
mrodm: Updated.
Reading that definition, I set both string and array of strings for both `dataset` and `namespace` keys.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -307,6 +332,29 @@ spec:
         dynamic_namespace:
           description: When set to true, agents running this integration are granted data stream privileges for all namespaces of its type
           type: boolean
+    routing_rules:  # technical preview
+      description: Routing rules set.
+      type: array
+      items:
```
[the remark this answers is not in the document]
mrodm: Applying this suggestion:
```diff
--- spec/integration/data_stream/manifest.spec.yml
+++ spec/integration/data_stream/manifest.spec.yml
@@ -358,20 +358,11 @@ spec:
           type: boolean
     routing_rules:  # technical preview
       description: Routing rules set.
-      type: array
-      items:
-        type: object
-        additionalProperties: false
-        properties:
-          dataset:
-            type: string
-          rules:
-            type: array
-            items:
-              $ref: "#/definitions/routing_rule"
-        required:
-          - dataset
-          - rules
+      type: object
+      additionalProperties:
+        type: array
+        items:
+          $ref: "#/definitions/routing_rule"
   allOf:
     - if:
         required:
```

works and it would be like in the example given in the issue.

But it cannot be applied, since some elements under `routing_rules` could contain dots, for instance `k8s.router`. As it is implemented now package-spec, every dot in these keys are expanded as objects ([link](https://github.com/elastic/package-spec/blob/5e86c23605e14f5800a966a7ab535ef106b5f0b8/code/go/internal/yamlschema/loader.go#L92))

I think we would need to keep this a list of elements

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
mrodm: I'll check how to take into account JSON patches in folder spec files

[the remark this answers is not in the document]
mrodm: There is a possible drawback of moving this to its own file.
If `routing_rules` are defined in it own file, the condition to set as required the `dataset` field , I think it cannot be achieved. They would be in different files.

Referring to this block:
```yaml
  allOf:
    - if:
        required:
          - routing_rules
      then:
        required:
          - dataset
```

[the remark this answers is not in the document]
mrodm: Added a new validation rule in code for this requirement, since it cannot be achieved with if-then-else block in JSON schema.

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
mrodm: If I understand correctly the docs and the example in the issue https://github.com/elastic/package-spec/issues/514, this field would define the source dataset to be used by the reroute processor to check for documents to be routed to another dataset. This latter dataset would be the ones defined in each element under `rules` key.

Being that, would `dataset` be a good name? another options that I can think of `source`, `source_dataset`? @jsoriano @juliaElastic 

If so, I would add these description to these fields
```diff
--- spec/integration/data_stream/manifest.spec.yml
+++ spec/integration/data_stream/manifest.spec.yml
@@ -364,8 +364,10 @@ spec:
         additionalProperties: false
         properties:
           dataset:
+            description: Source dataset to be used for this reroute processsor
             type: string
           rules:
+            description: List of routing rules
             type: array
             items:
               $ref: "#/definitions/routing_rule"
```

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    # Route access logs to `routing_rules.access` when they're sourced from an access logfile
+    - dataset: routing_rules.access
+      if: "ctx?.file?.path?.contains('/var/log/routing_rules/access')"
+      namespace:
+        - "{{labels.data_stream.namespace}}"
+        - default
+
+  # Route K8's container logs to this catch-all dataset for further routing
+  - dataset: k8s.router
```
jsoriano, line 19: We cannot make this a map because of support for dots, right?

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
```diff
@@ -0,0 +1,35 @@
+routing_rules:
```
jsoriano, line 1: Is this key redundant with the name of the file? Could we have directly a list?

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
```diff
@@ -0,0 +1,35 @@
...
+  # "Local" routing rules are included under this current dataset, not a special case
+  - dataset: routing_rules
+    rules: # Route error logs to `routing_rules.error` when they're sourced from an error logfile
+    - dataset: routing_rules.error
```
jsoriano, line 5: It may be confusing to have two `dataset` attributes at different levels, could we be more specific about what each of them represent?

jsoriano (commented):
Looks good, waiting for the other PR to be merged.

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
mrodm: Exactly, I tried to set a map for these definitions. However, that `dot` causes that the keys are expanded and therefore it is interpreted as an object.

Here it is what I tried https://github.com/elastic/package-spec/pull/535#discussion_r1229428169

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
mrodm: This is related to the question I posted here https://github.com/elastic/package-spec/pull/535#discussion_r1234204623

From what I understand, I think the first `dataset` (at line 3) could be interpreted as "source dataset". Is that right? 

If so, maybe it could be renamed both fields to be `source_dataset` and `target_dataset` ?

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
mrodm: Yes, it looks like this key could be removed since this is now in its own file. I'll update the spec.

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
mrodm: Options that I could think of @juliaElastic @jsoriano :
- source_dataset and target_dataset
```yaml
# "Local" routing rules are included under this current dataset, not a special case
- source_dataset:  nginx
  # Route error logs to `nginx.error` when they're sourced from an error logfile
  - target_dataset: nginx.error
    if: "ctx?.file?.path?.contains('/var/log/nginx/error')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default

  # Route access logs to `nginx.access` when they're sourced from an access logfile
  - target_dataset: nginx.access
    if: "ctx?.file?.path?.contains('/var/log/nginx/access')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default
```
- source_dataset and dataset
```yaml
# "Local" routing rules are included under this current dataset, not a special case
- source_dataset:  nginx
  # Route error logs to `nginx.error` when they're sourced from an error logfile
  - dataset: nginx.error
    if: "ctx?.file?.path?.contains('/var/log/nginx/error')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default

  # Route access logs to `nginx.access` when they're sourced from an access logfile
  - dataset: nginx.access
    if: "ctx?.file?.path?.contains('/var/log/nginx/access')"
    namespace:
      - {{labels.data_stream.namespace}}
      - default
```

spec/integration/data_stream/manifest.spec.yml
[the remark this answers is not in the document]
juliaElastic: I think `source_dataset` is the clearest name, I vote for that.

test/packages/good_v2/data_stream/routing_rules/routing_rules.yml
mrodm: Talked offline with @jsoriano 
Updated fields to be `source_dataset` and `target_dataset`, so those field names are more explicit.

spec/integration/data_stream/manifest.spec.yml
mrodm: Probably, it could be kept as it is without validation rule

code/go/internal/validator/spec.go
```diff
@@ -131,6 +131,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateILMPolicyPresent, since: semver.MustParse("2.0.0"), types: []string{"integration"}},
 		{fn: semantic.ValidateProfilingNonGA, types: []string{"integration"}},
 		{fn: semantic.ValidateKibanaObjectIDs, types: []string{"integration"}},
+		{fn: semantic.ValidateRoutingRulesAndDataset, types: []string{"integration"}, since: semver.MustParse("2.9.0")},
```
mrodm, line 134: Added here the `since`, to avoid checking for routing rules before 2.9.0.

It has been added a JSON Patch to remove it before 2.9.0.

spec/integration/data_stream/routing_rules.spec.yml
```diff
@@ -0,0 +1,74 @@
...
+       source_dataset:
+         description: >
+           Source dataset to be used by this reroute processsor.
+           If applicable, documents from this dataset will be routed according to the rules defined.
+         type: string
+       rules:
+         description: List of routing rules
+         type: array
+         items:
+           $ref: "#/definitions/routing_rule"
```
jsoriano, line 71: Nit. Should we require this to have at least one rule?
```suggestion
         type: array
         minItems: 1
         items:
           $ref: "#/definitions/routing_rule"
```

spec/integration/data_stream/spec.yml
```diff
@@ -78,3 +78,17 @@ spec:
       required: false
       visibility: private
       $ref: "./_dev/spec.yml"
+    - description: File containing routing rules definitions (technical preview)
+      type: file
+      contentMediaType: "application/x-yaml"
+      sizeLimit: 5MB
```
jsoriano, line 84: Nit. Are you setting an explicit size limit here for some reason? I think that for `application/x-yaml` files the `configurationSizeLimit` applies, that is already 5MB.
```suggestion
      contentMediaType: "application/x-yaml"
```

spec/integration/data_stream/routing_rules.spec.yml
```diff
@@ -0,0 +1,74 @@
...
+     additionalProperties: false
+     properties:
+       source_dataset:
+         description: >
+           Source dataset to be used by this reroute processsor.
+           If applicable, documents from this dataset will be routed according to the rules defined.
+         type: string
+       rules:
+         description: List of routing rules
+         type: array
```
jsoriano, line 69: Nit. Maybe add a comment here mentioning that this is not an object because using the source dataset as key would require to support keys with dots.

jsoriano (approved):
LGTM, some nitpicking but can be ignored :slightly_smiling_face:

spec/integration/data_stream/spec.yml
mrodm: Just copied from the manifest.yml definition. I'll remove it :+1:

spec/integration/data_stream/routing_rules.spec.yml
mrodm: Yes, that would be better to avoid empty files or empty arrays.

jsoriano (approved):

spec/integration/data_stream/routing_rules.spec.yml
mrodm: Added, better to keep it as a comment

spec/integration/data_stream/routing_rules.spec.yml
mrodm: There is an error that could be related to the conversion it is done between YAML and JSON

> 1. could not read root folder spec file: could not load spec for "integration/data_stream/spec.yml": could not load schema for "integration/data_stream": failed to load schema for "integration/data_stream/routing_rules.spec.yml": minItems must be of an integer


I'll remove this `minItems`

juliaElastic (approved):
LGTM

PR #545, opened 2023-06-20
Fix changelog entries

mrodm:
## What does this PR do?

Fix changelog entries to be in the following version to be released.

Follow-up https://github.com/elastic/package-spec/pull/526


jsoriano (approved):
The "Prepare for next version" change can be removed.

mrodm:
> The "Prepare for next version" change can be removed.

I'll remove that entry when preparing the following version 2.9.0 :+1:

PR #546, opened 2023-06-22
Prepare version 2.9.0

mrodm:
## What does this PR do?

This PR prepares the changelog to release package-spec 2.9.0


jlind23 (approved):

jsoriano (approved):
Perhaps the support for JSON patches wouldn't need to be in the changelog, as is not an user facing change, but is fine, as you prefer.

mrodm:
> Perhaps the support for JSON patches wouldn't need to be in the changelog, as is not an user facing change, but is fine, as you prefer.

If it is not a problem, I would keep that changelog entry even it is not a user facing change

PR #550, opened 2023-06-27
migrate to buildkite

sharbuz:

The partial migration of the Jenkins pipeline to the Buildkite pipeline. 

Preparing the Buldkite pipeline instead Jenkins pipeline

It doesn't be checked locally, that's why it's a partial "step by step" migration

## Checklist

## Related issues

-


sharbuz:
buildkite test integrations

PR #552, opened 2023-06-28
The migration of the Jenkins pipeline to the Buildkite pipeline

sharbuz:

The migration of the Jenkins pipeline to the Buildkite pipeline. 

Preparing the Buldkite pipeline instead Jenkins pipeline

It doesn't be checked locally, that's why it's a partial "step by step" migration

## Checklist

## Related issues

-

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -37,12 +37,12 @@ steps:
           provider: "gcp"
           image: "${GO_WINDOWS_AGENT_IMAGE}"
         artifact_paths:
-          - build/*.xml
+          - "*.xml"
 
       - label: ":junit: Junit annotate"
         plugins:
           - junit-annotate#v2.4.1:
-              artifacts: "build/*.xml"
+              artifacts: "*.xml"
               fail-build-on-error: true
         agents:
           provider: "gcp"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,52 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: 'v0.5.0' # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
+  SETUP_MAGE_VERSION: "latest"
+  GH_CLI_VERSION: "2.29.0"
+  JQ_VERSION: "1.6"
+  GO_LINUX_AGENT_IMAGE: "golang:${GO_VERSION}"
+  GO_WINDOWS_AGENT_IMAGE: "family/ci-windows-2022"
+
+steps:
+  - label: ":go: Run check-static"
+    key: check
+    command: "make check"
+    agents:
+      image: "${GO_LINUX_AGENT_IMAGE}"
+      cpu: "8"
+      memory: "4G"
+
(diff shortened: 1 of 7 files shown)


.buildkite/pipeline.yml
```diff
@@ -37,12 +37,12 @@ steps:
...
           image: "${GO_WINDOWS_AGENT_IMAGE}"
         artifact_paths:
-          - build/test-results/test-windows-*.xml
+          - build/test-results/*.xml
 
       - label: ":junit: Junit annotate"
         plugins:
           - junit-annotate#v2.4.1:
-              artifacts: "build/test-results/test-*.xml"
+              artifacts: "bbuild/test-results/*.xml"
```
jlind23, line 45: was the double "b" added on purpose here?

jlind23 (approved):

.buildkite/pipeline.yml
sharbuz: no, by mistake.
I already got the error in the last stage ))
fixed.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,52 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: 'v0.5.0' # https://github.com/andrewkroh/gvm/issues/44#issuecomment-1013231151
+  SETUP_MAGE_VERSION: "latest"
+  GH_CLI_VERSION: "2.29.0"
```
mrodm, line 6: Do we use gh cli here ? IIRC it is not, and I could be removed.

mrodm (commented):
we could add as part of this release a change in the `catalog-info.yaml` to add the needed environment variable to add comments with the status of the build:

```diff
diff --git catalog-info.yaml catalog-info.yaml
index ba2b11a..58dcb88 100644
--- catalog-info.yaml
+++ catalog-info.yaml
@@ -56,6 +56,8 @@ spec:
       cancel_intermediate_builds_branch_filter: '!main'
       skip_intermediate_builds: true
       skip_intermediate_builds_branch_filter: '!main'
+      env:
+        ELASTIC_PR_COMMENTS_ENABLED: 'true'
       teams:
         ecosystem:
           access_level: MANAGE_BUILD_AND_READ
```

.buildkite/pipeline.yml
sharbuz: removed

sharbuz:
Regarding the comment:
https://github.com/elastic/package-spec/pull/552#pullrequestreview-1504763888
If this PR is fine, probably it's better to create one more PR with changes only in the catalog-info.yml: 
1. your suggestion 
2. add Slack notifications about fails.
@mrodm WDYT?

mrodm:
> Regarding the comment: [#552 (review)](https://github.com/elastic/package-spec/pull/552#pullrequestreview-1504763888) If this PR is fine, probably it's better to create one more PR with changes only in the catalog-info.yml:
> 
> 1. your suggestion
> 2. add Slack notifications about fails.
>    @mrodm WDYT?
@sharbuz It can be added in this PR the update for catalog-info.yaml.

About slack notifications when it fails, currently Jenkins is not configured to do so. Therefore I would keep that way and not add those notifications

mrodm (approved):
LGTM

PR #555, opened 2023-06-30
Debug issue

sharbuz:

Error getting the GO version
https://buildkite.com/elastic/package-spec/builds/15#018907c7-f525-4bd0-a626-82bc76360fcf

Moved the exportin of the variable in the post-checkout script

It worked in the previous PR because it was PR

## Checklist

## Related issues

-


jlind23 (approved):

jlind23:
@mrodm we went down the path of manually defining the go version which was the easiest solution here and the same approached taken here: https://github.com/elastic/elastic-package/blob/main/.buildkite/pipeline.yml

PR #561, opened 2023-07-05
Add condition to define stack capability requirements

### LINKED ISSUE #560: [Change Proposal] Add new field for dependency on stack capabilities
There are some packages that are only useful if certain Elastic features are available. Some of these features are not going to be available in all the serverless offerings, according to [these configurations](https://github.com/elastic/kibana/tree/bfab1b0659269cf67d6a864d6a80ee29632b4b4c/config). Packages requiring these features are going to need to declare it.

Add a new field for dependencies on stack capabilities. We can try to add them as a new condition:
```
conditions:
  kibana.version: '^8.10.0'
  elastic.subscription: 'basic'
  elastic.capabilities:
    - apm
    - uptime
```

But so far we don't have lists in conditions. If this is a problem, we can use comma-separated lists:
```
  elastic.capabilities: 'apm,uptime'
``` 

Or add a completely new field `stack_capabilties`.

We need to confirm the final list of capabilities, but so far we can start with:
- `apm`
- `uptime`
- `enterprise_search`
- `serverless_search`
- `observability`
- `logs_ui`
- `serverless` (TBD, for packages intended only for serverless)
- `ilm`/`dlm` (TBD, only if there are packages that only work with one of the lifecycle managers)

jsoriano:
## What does this PR do?

Add condition to define stack capability requirements.

## Why is it important?

There are some features that may be disabled on some deployments, and there are packages that need these features. We need something to indicate the required features so we can prevent users to try to use these packages on deployments where they are not going to work properly.

## Checklist

## Related issues

- Fixes #560.

### DIFF
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -107,6 +107,7 @@ spec:
                 The package should not be used in deployments without the indicated features.
                 Packages that don't indicate any capability condition can be used on any deployment.
               type: array
+              uniqueItems: true
               items:
                 type: string
                 enum:
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -101,6 +101,21 @@ spec:
               default: basic
               examples:
               - basic
+            capabilities:
+              description: |-
+                Stack features that are required by the package to work properly.
+                The package should not be used in deployments without the indicated features.
+                Packages that don't indicate any capability condition can be used on any deployment.
+              type: array
+              items:
+                type: string
+                enum:
+                  - apm
+                  - enterprise_search
+                  - observability
+                  - security
+                  - serverless_search
+                  - uptime
         kibana:
           description: Kibana conditions
           type: object
@@ -379,6 +394,10 @@ spec:
 
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 2.10.0
+    patch:
+      - op: remove
+        path: "/definitions/conditions/properties/elastic/properties/capabilities"
   - before: 2.3.0
     patch:
       - op: add
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -2,11 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom of each in-development version.
 ##
-- version: 2.9.1-next
+- version: 2.10.0-next
   changes:
-  - description: Prepare for next version
+  - description: Packages can define required stack capabilities
     type: enhancement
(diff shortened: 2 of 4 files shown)


spec/integration/manifest.spec.yml
```diff
@@ -101,6 +101,21 @@ spec:
...
+              type: array
+              items:
+                type: string
+                enum:
+                  - apm
+                  - enterprise_search
+                  - observability
+                  - security
+                  - serverless_search
+                  - uptime
```
mrodm, line 118: Maybe it is worth to try to add `uniqueItems: true` field here (https://json-schema.org/understanding-json-schema/reference/array.html#uniqueness), to ensure that there are no duplicated items.
```suggestion
              type: array
              uniqueItems: true
              items:
                type: string
                enum:
                  - apm
                  - enterprise_search
                  - observability
                  - security
                  - serverless_search
                  - uptime
```

Never tested in any field this setting, but it could be helpful even with those few elements. WDYT ?

Tested with this change
```diff
--- test/packages/good_v2/manifest.yml
+++ test/packages/good_v2/manifest.yml
@@ -11,6 +11,9 @@ conditions:
   elastic.subscription: 'basic'
   elastic.capabilities:
     - observability
+    - observability
+    - apm
+    - apm
 vars:
   - name: package_password
     type: password
```
shows these errors:
```
        	Error Trace:	/home/mariorodriguez/Coding/work/package-spec/code/go/pkg/validator/validator_test.go:599
        	Error:      	Received unexpected error:
        	            	found 2 validation errors:
        	            	   1. file "../../../../test/packages/good_v2/manifest.yml" is invalid: field conditions.elastic.capabilities: array items[0,1] must be unique
        	            	   2. file "../../../../test/packages/good_v2/manifest.yml" is invalid: field conditions.elastic.capabilities: array items[2,3] must be unique
        	Test:       	TestValidateRoutingRules/good_v2
```

mrodm (approved):
LGTM
Just a suggestion to avoid duplicated values when defining this new field in a manifest.

mrodm (approved):

spec/integration/manifest.spec.yml
```diff
@@ -101,6 +101,22 @@ spec:
               default: basic
               examples:
               - basic
+            capabilities:
+              description: |-
+                Stack features that are required by the package to work properly.
```
andrewkroh, line 106: @jsoriano The wording implies to me that _all_ the specified capabilities are required to run a package. Is that a correct interpretation? In other words, these capabilities are `AND`ed together to determine eligibility. If that's not the case then I think the wording should be clarified.

PR #567, opened 2023-07-19
Add saved object tags definitions file under kibana folder

### LINKED ISSUE #396: [Change Proposal] Add asset tags to package level manifest.yml
## What problem the proposal is solving?
Dashboard from different packages should be able to have the same saved object tag applied without creating multiple saved object tags that have the same name/description/color but different IDs.

## Where the solution will need to be implemented?

Package-spec should support specifying `tags` with `text` and `asset_types` or `asset_ids` in the package manifest:

```yml
# integration/manifest.yml
tags:
- text: SecurityExternalIntegrations
  asset_types:
  - dashboard
- text: ThreatDetectionDashboard
  asset_ids:
  - dashboard3
  - dashboard4
```

These tags will be created by Fleet during package installation time.

## Related Issues

- https://github.com/elastic/kibana/issues/152814

<details><summary>Original description</summary>
## Proposal

Allow an optional list of tags to be specified in the package's `manifest.yml` that are to be applied to all dashboards contained in the integration.

```yaml
# manifest.yml
dashboard:
  tags:
    - name: Threat Intel # Required
      color: '#FFFFFF'  # Required
      description: Relates to security threat intelligence data. # Optional
```

## What problem the proposal is solving?

Dashboard from different packages should be able to have the same saved object tag applied without creating multiple saved object tags that have the same name/description/color but different IDs.

Today, it is not possible to apply the same saved object tag across multiple packages. This is because the package-spec requires [^1] that all object ID's begin with the package name. This prevents using the same object ID for a tag across multiple packages. There is a good reason to prevent object ID collisions between packages -- it allows for clear ownership of the object such that it can be upgraded or removed with the owning package.

By shifting responsibility of tag creating into Fleet it should be possible for Fleet to reuse the existing tag object when the name, description, and color match. This greatly improves the user experience within Kibana since there will be fewer (or none) tags with the same name. This screenshot shows what happens today if multiple packages create a tag with the same name.

Screen Shot 2022-08-10 at 11 12 23

[^1]: https://github.com/elastic/package-spec/blob/40b2a04971e90c59f651044a4c396e3fdde39817/code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go#L19-L23

## Where the solution will need to be implemented?

Fleet (the part in Kibana) will need to be updated to create tags and add them to dashboards. If the saved object ID for the tags is derived from the name, color, and description attributes then it can ensure that only one saved object exists with those attributes (e.g. `fleet-tag-<hash(name, color, description)>`).

## Related Issues

- https://github.com/elastic/integrations/issues/3951#issuecomment-1215695606
</details>

mrodm:
## What does this PR do?

Add support to add a new tags file under `kibana` folder: `kibana/tags.yml`

Example of the contents:
```yaml
- text: mycustomtag
  asset_types:
    - dashboard
- text: tag42
  asset_ids:
    - id1
    - id2
```

This will be just supported if all these conditions are fulfilled:
- **kibana.version** is equal or greater to **8.10.0**
- **format_version** is at least **2.10.0**

## Why is it important?

Tags defined in this file are going to be used to create shared tags between different packages.

## Checklist

- [x] Added JSON Patches to remove the support for previous versions: 2.10.0
- [x] Added validation check to ensure a minimum Kibana Version: 8.10.0

## Related issues

- Closes #396 
- Relates https://github.com/elastic/kibana/issues/152814


spec/integration/kibana/tags.spec.yml
```diff
@@ -0,0 +1,51 @@
...
+             - search
+             - map
+             - lens
+             - index_pattern
+             - security_rule
+             - csp_rule_template
+             - ml_module
+             - tag
+             - osquery_pack_asset
+             - osquery_saved_query
```
mrodm, line 32: These are the asset types that are allowed to be under `kibana` folder:
https://github.com/elastic/package-spec/blob/7990807d58465a669fac5ce95f8857fdbc232128/spec/integration/kibana/spec.yml

P1llus (dismissed):
LGTM, any reason we decided to go for a new file rather than just manifest.yml?
Thanks for the quick work!

spec/integration/kibana/spec.yml
```diff
@@ -1,121 +1,133 @@
...
+    required: false
+    contents:
+    - description: An osquery saved query file
+      type: file
+      contentMediaType: "application/json"
+      pattern: '^{PACKAGE_NAME}-.+\.json$'
+  - description: File containing saved object tag definitions for assets
+    type: file
+    contentMediaType: "application/x-yaml"
+    name: "tags.yml"
```
juliaElastic, line 125: The original request was to add tags to the main `manifest.yml`, do we recommend to add to a separate file?

spec/integration/kibana/tags.spec.yml
```diff
@@ -0,0 +1,51 @@
...
+         type: array
+         items:
+           type: string
+     if:
+       required:
+         - asset_types
+     then:
+       not:
+         required:
+           - asset_ids
```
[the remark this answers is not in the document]
juliaElastic: I don't see this in the requirements, I think we could allow setting both, and Fleet would add the tag on all types and ids that are configured.

spec/integration/kibana/tags.spec.yml
```diff
@@ -0,0 +1,44 @@
...
+             - dashboard
+             - visualization
+             - search
+             - map
+             - lens
+             - index_pattern
+             - security_rule
+             - csp_rule_template
+             - ml_module
+             - tag
```
jsoriano, line 30: Umm, how are these tag assets different to the ones we are adding here?

spec/integration/kibana/spec.yml
jsoriano: > @elastic/ecosystem WDYT about creating a new file for these tag definitions ?

I'd prefer to have these definitions separated from the manifest file, yes.

P1llus: Let's do that then 👍

spec/integration/kibana/tags.spec.yml
mrodm: True, I just filled this enum list with all the types available in kibana folders. But that's true that `tag` would not be added other tags... I'll remove it from the enum list.

mrodm: About how these `tags` are different from the ones adding here. The tags added in this PR/issue are aimed to be shared tags between different dashboards/assets. So all of them will have the same tag (with the same id).

spec/integration/kibana/tags.spec.yml
mrodm: Just removed `tag` from the list. I think tags would not be assigned other tags.

spec/integration/kibana/tags.spec.yml
```diff
@@ -0,0 +1,43 @@
...
+  description: Tags automatically added to saved object assets.
+  items:
+     type: object
+     additionalProperties: false
+     properties:
+       text:
+         description: Tag name.
+         type: string
+       asset_types:
+         description: Asset types where this tag is going to be added.
```
jsoriano, line 16: So the tag will be added to all the assets of these types included in the package?

Maybe I would rephrase this then to something like:
```suggestion
         description: This tag will be added to all the assets of these types included in the package.
```

jsoriano (approved):

juliaElastic (approved):
LGTM

spec/integration/kibana/tags.spec.yml
mrodm: Yes, that would be more explicit.

@juliaElastic just to confirm, if the same tag is defined in other package. Assets from both packages would be using the same `tag`?

mrodm: According to the original isuse https://github.com/elastic/kibana/issues/152814 it looks like that tag could be the same (reused) in other packages.

> When Fleet installs a package that includes these top-level tags, it should generate the tag + asset references using a unique ID for the tag object. This ID should be a unique identifier based on the downcased text value to avoid collisions and duplications. For example, if two packages define a SecurityExternalIntegrations and securityexternalintegrations tag, these would result in only one tag object being created in Kibana.

mrodm: Updated `asset_types` and `asset_ids` to include this reference

mrodm: @jsoriano @juliaElastic please check if that description makes sense, thanks in advance!

juliaElastic (approved):

jsoriano (approved):
:+1:

PR #568, opened 2023-07-20
Add owner.type field to indicate package support level

taylor-swanson:
## What does this PR do?

- Add new owner.type field to indicate the support level of a package in the manifest.yml file.
- The 'elastic' value indicates the package is built and supported by Elastic
- The 'partner' value indicates the package is built and supported by a partner/vendor and may include involvement from Elastic
- The 'community' value indicates the package is built and supported by non-Elastic community members
- The field is currently not required

## Why is it important?

Identifies the expected support owner and level of our packages. Other areas, such as the Kibana Integrations UI, can display notices on the support level or provide for filtering to only show Elastic-supported packages, for instance.

## Checklist

## Related issues

- Relates elastic/integrations#6569


spec/integration/manifest.spec.yml
```diff
@@ -240,6 +240,13 @@ spec:
           description: GitHub repository name of package maintainer.
           type: string
           pattern: '^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\/[a-zA-Z0-9-]+))$'
+        type:
+          description: Describes who owns the package and the level of support that is provided.
```
jsoriano, line 244: We should probably define what to expect when this value is not set.

In principle I see two options:
* `elastic` as default, as most existing packages are ours. This may give problems if publishers forget to set the value in the future.
* Add this field starting on next package-spec version (2.10.0). Assume `community` as default from this version of the spec, and `elastic` for current packages. We could add something in our CI to ensure that we set the field.

ebeahan: ++ to defining a default.

I prefer the second option of having `elastic` needing set explicitly for future packages we release.

taylor-swanson: I definitely agree on setting a default and also agree with defaulting to `community` for the next package-spec version. For existing packages, I think I saw a `before` constraint that we can set to satisfy this:

> `elastic` for current packages.

I saw this, for example:

```yaml
  - before: 2.3.0
    patch:
      - op: add
        path: "/properties/release"
        value:
          description: The stability of the package (deprecated, use prerelease tags in the version).
          deprecated: true # See https://github.com/elastic/package-spec/issues/225
          type: string
          enum:
          - experimental
          - beta
          - ga
          default: ga
          examples:
          - experimental
```

Would I do something similar for `owner.type` and default it to `elastic` when the version is before `2.10.0`?

taylor-swanson: Basically, this:
```diff
 # JSON patches for newer versions should be placed on top
 versions:
   - before: 2.10.0
     patch:
       - op: remove
         path: "/definitions/conditions/properties/elastic/properties/capabilities"
+      - op: add
+        path: "/definitions/owner/properties/type"
+        value:
+          description: Describes who owns the package and the level of support that is provided.
+          type: string
+          default: elastic
+          enum:
+            - elastic
+            - partner
+            - community
+          examples:
+            - community
   - before: 2.3.0
```

jsoriano: Well, you could replace only the default, with something like:
```
+      - op: replace
+        path: "/definitions/owner/properties/type/default"
+        value: elastic
```

Or directly remove the definition so it cannot be used in older versions to avoid confusion:
```
+      - op: remove
+        path: "/definitions/owner/properties/type"
```

taylor-swanson: I'm leaning towards the "replacing the default" option so that way a value is always present, rather than the definition being absent in some cases. If you don't think that'd be a concern, then I'm okay with removing the definition as well. Either change seems pretty straightforward to add.

taylor-swanson:
@jsoriano, looks like `2.10.0` was just released. I'd imagine we'd want this going into `2.11.0` now and not `2.10.1`?

jsoriano:
> @jsoriano, looks like `2.10.0` was just released. I'd imagine we'd want this going into `2.11.0` now and not `2.10.1`?

Hi, I am very sorry that we missed this change. Yes, In principle I am afraid that it will have to wait for 2.11.0.

spec/integration/manifest.spec.yml
```diff
@@ -240,6 +240,16 @@ spec:
           description: Github team name of the package maintainer.
           type: string
           pattern: '^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\/[a-zA-Z0-9-]+))$'
+        type:
+          description: Describes who owns the package and the level of support that is provided.
```
jsoriano, line 244: Nit. Please add to the description what each possible value is intended for.

spec/changelog.yml
```diff
@@ -2,6 +2,11 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom of each in-development version.
 ##
+- version: 2.11.0-next
+  changes:
+  - description: Add owner.type field to indicate package support level
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/568
 - version: 2.10.1-next
```
jsoriano, line 10: Please, remove this block as it won't be needed.

spec/integration/manifest.spec.yml
```diff
@@ -241,7 +241,13 @@ spec:
...
           pattern: '^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\/[a-zA-Z0-9-]+))$'
         type:
-          description: Describes who owns the package and the level of support that is provided.
+          description: >
+            Describes who owns the package and the level of support that is
+            provided. The 'elastic' value indicates that the package is built
+            and maintained by Elastic. The 'partner' value indicates that the
+            package is built and maintained by a partner vendor and may include
+            involvement from Elastic. The 'community' value indicates the package
+            is built and maintained by non-Elastic community members.
```
jsoriano, line 250: :+1: thanks

jsoriano (approved):

spec/changelog.yml
taylor-swanson: Not quite sure I understand, do you want me to put the changelog entry under `2.10.1-next` instead?

jsoriano: No, we have to remove at some point the `2.10.1-next` block, but no worries, we can also remove it later.

taylor-swanson: I'll put it under `2.10.1-next` so at least the entry is in there, but we can remove/rearrange later like you say.

jsoriano (approved):

jsoriano:
Let's merge this before we forget again :slightly_smiling_face:

Issue #349, opened 2022-06-02
Add support for `subobjects: false`


ruflin:
In https://github.com/elastic/elasticsearch/pull/86166 Elasticsearch introduced the `subobjects: false` parameter. This allows ingest flat key/value pairs that had in the past conflicts like `host` and `host.name`.

package-spec and Fleet should support this option in templates. 

At first, the feature should be used with care in packages as the objects must be flattened before ingestion (see also https://github.com/elastic/apm/issues/347 for more details). At the moment ingestion tools like Beats often follow the opposite approach and expand all objects. There are potential future improvements that solve this problem: https://github.com/elastic/apm/issues/347#issuecomment-1140947927

From https://github.com/elastic/package-spec/issues/425, we need to ensure that at least one of the following mappings produce expected results, so this is supported.

------
```
- name: prometheus.labels.*
  type: object
  object_type: keyword
  subobjects: false
```
Should produce:
* A dynamic mapping for `prometheus.labels.*`, with type keyword.
* A static mapping for `prometheus.labels`, with `type: object` and `subobjects: false`.
-------
```
- name: prometheus.labels
  type: object
  subobjects: false
```

Should produce a mapping for `prometheus.labels` with `type: object` and `subobjects: false`, independently of the existence of other related dynamic mappings.

------

Maybe this already works, but we would need to check it.
```[tasklist]
### Tasks
- [ ] https://github.com/elastic/package-spec/pull/573
- [ ] https://github.com/elastic/kibana/pull/171826
- [ ] https://github.com/elastic/package-spec/pull/727
```

ruflin:
Having `subobject: false` could have a broader impact on how we do things. Few thoughts:

* For fields.yml, should we flatten these by default?
* Should we flatten fields for component templates in Elasticsearch loaded by fleet for every object that does not have properties? @kpollich 
* Should we ship data as mostly key/value pairs to Elasticsearch from Elastic Agent instead of the nested objects @ph @cmacknz 
* Can we remove most of the `common.MapStr` magic inside Beats / Agent?

We don't need to necessarily have all these discussions here but worth referencing. I'm excited about the potential future opportunities this offers. (@joshdover )

jsoriano:
This is great, and I think that in many of the cases where we would take more advantage of this, basically where we dedot now, the field is already flattened, because we send it as is read from origin.

I think we could add support for this, and introduce it eventually in a case per case basis.

[...]

PR #573, opened 2023-08-01
Add support for subobjects

zmoog:
## What does this PR do?

Add subobjects item to the fields spec.

Field definition supports a new `boolenan` item named `subojects` that allow user to turn off the expansions of field names containing dots into actual tree of objects.

For example if this is set to `true`, a field named `foo.bar` will be expanded into an object with a field named `bar` inside an object named `foo`.

## Why is it important?

Give end user and integration developers control on expansion of field names with dots into sub objects. 

Metric names often contains names with dots, so we need to disable this behaviour in some contexts.

## Checklist

## Related issues

- Relates #349


felixbarny:
This adds support for setting `subobjects` on a per-field basis. I think the level of granularity is too fine here. There should be a per-data stream setting that controls the `subobjects` setting at the top-level of the mapping.

In the future, the goal is to set `subobjects` to `false` by default in Fleet, unless an integration has explicitly set it to `true`. For example, integrations that use the `nested` field type won't work with `subobjects: false` and will therefore need to opt-out. For these integrations, it may still be useful to be able to set `subobjects: false` on a per-field level. However, as the expectation is to have the default be `subobjects: false`, most integrations won't need that. Therefore, I think we should focus on having a top-level setting here.

I suppose, we'd add this to [data_stream/manifest.spec.yml](https://github.com/elastic/package-spec/blob/main/spec/integration/data_stream/manifest.spec.yml) in `elasticsearch.index_template.mappings.subobjects`.

jsoriano (commented):
Let's don't forget about adding a changelog entry, if packages using this are going to require the change in kibana we would need to include this in 3.1.0.

Also, following this comment https://github.com/elastic/package-spec/pull/573#issuecomment-1725339642, do we need the property at the data stream level? And if we add this one, do we need `subobjects` per field?

zmoog:
> Also, following this comment https://github.com/elastic/package-spec/pull/573#issuecomment-1725339642, do we need the property at the data stream level? 

Yep, we planned to add support for subobjects at both data stream and field level. 

However, while working on the Fleet plugin in Kibana to set subobjects the data stream level, I found issues in setting the subobjects for the whole mapping. IIRC, we need something in Elasticsearch to make this happen. @felixbarny, am I remembering this right, and what's missing to make this happen?

So, in this phase, we decided to move forward with subobjects at the field level.

> And if we add this one, do we need subobjects per field?

I believe that having a subobjects per field would make the adoption of the feature easier. Flipping the switch for the whole data stream is riskier than setting it to false on specific fields that really need this. 

I would use a per field subobjects today on integrations like Azure Logs or Azure Metrics to deal with user provided metadata or a few Azure services that sends weird content (for example, an embedded JSON used as a key in a map).

felixbarny:
I agree with all of the above.

The issue is that Elasticsearch requires the mappings to be flat when subobjects is set to `false` at the root level.

For example, this would work:
```json
{
  "subobjects": false,
  "properties": {
    "foo.bar": {
      "type": "keyword"
    }
  }
}
```


But this wouldn't work:
```json
{
  "subobjects": false,
  "properties": {
    "foo": {
      "properties": {
        "bar": {
          "type": "keyword"
        }
      }
    }
  }
}
```

We have plans to automatically flatten the mappings in Elasticsearch: https://github.com/elastic/elasticsearch/issues/99860. But it will take us a while to get there. In the meantime, I think it makes a lot of sense that we can add subobjects: false to some properties, such as `labels`. This will allow us to adopt that feature gradually and de-risk the full adoption. The eventual goal is to disable subobjects at the root level everywhere.

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -402,6 +402,15 @@ spec:
...
         default: false
+      
+      subobjects:
+        description: >
+          Specifies if field names containing dots should be expanded into subobjects.
+          For example, if this is set to `true`, a field named `foo.bar` will be
+          expanded into an object with a field named `bar` inside an object
+          named `foo`.
+        type: boolean
+        default: true
```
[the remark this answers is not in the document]
zmoog: Oh! 🤦 

Now I get how the patches work. Adding it!

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -548,12 +557,26 @@ spec:
...
 
     required:
     - name
 
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.1.0
+    patch:
+      - op: remove
+        path: "/items/allOf/8" # removing subobjects when type is object
```
jsoriano, line 579: Remove `subobjects` property too?
```suggestion
        path: "/items/allOf/8" # removing subobjects when type is object
      - op: remove
        path: "/items/properties/subobjects"
```

spec/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/671
+  - description: Add support for subobjects in fields
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/573
```
jsoriano, line 12: According to the patch, this version should be `3.1.0-next`.

test/packages/good_v3/data_stream/subobjects/fields/some_fields.yml
```diff
@@ -0,0 +1,13 @@
...
+  type: group
+  fields:
+    - name: a.labels.*
+      type: object
+      object_type: keyword
+      subobjects: false
+    - name: b.labels
+      type: object
+      object_type: keyword
+      subobjects: false
```
jsoriano, line 13: :+1:

code/go/pkg/validator/validator_test.go
```diff
@@ -173,6 +173,12 @@ func TestValidateFile(t *testing.T) {
 				"field streams.0.vars.3: Must not be present",
 			},
 		},
+		"bad_subobjects": {
+			"data_stream/rules/fields/base-fields.yml",
+			[]string{
+				`field 5.type: 5.type must be one of the following: "object"`,
```
jsoriano, line 179: Looks like CI is failing because of this, the wrong field is in sixth position now.
```suggestion
				`field 6.type: 6.type must be one of the following: "object"`,
```

PR #581, opened 2023-08-29
[Buildkite] Use Go version from file

mrodm:
## What does this PR do?

This PR ensures that all the steps in Buildkite use the same version as the one defined in .go-version file.

### DIFF
--- a/.buildkite/scripts/run-win-tests.ps1
+++ b/.buildkite/scripts/run-win-tests.ps1
@@ -8,7 +8,7 @@ function fixCRLF {
     git reset --quiet --hard
 }
 function withGolang($version) {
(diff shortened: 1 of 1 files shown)


.buildkite/hooks/pre-command
```diff
@@ -6,6 +6,8 @@ source .buildkite/scripts/tooling.sh
 
 GITHUB_TOKEN_VAULT_PATH=kv/ci-shared/platform-ingest/github_token
 
+export GO_VERSION=$(cat .go-version)
```
jsoriano, line 9: Nice to be able to use the same source of truth for this :+1:

jsoriano (approved):

PR #585, opened 2023-09-01
Add support for flattened type dimensions

### LINKED ISSUE #584: Add support for flattened type dimensions
Currently, `flattened` type cannot be a dimension.  This limitation restricts us from effectively utilizing flattened fields within TSDB enabled data streams. The point of this is enhancing the existing functionality by allowing the declaration of multiple values as dimensions, even when they are flattened.

```yaml
- name: dimensions
  type: flattened
  dimensions: ["a", "a.b", "c.d"]
```

Support for flattened multiple dimensions was added in ES in `8.8` https://github.com/elastic/elasticsearch/pull/94842

ghost:
## What does this PR do?

Adds support for `flattened` type dimensions.

## Why is it important?

Currently, `flattened` type cannot be a dimension. This limitation restricts us from effectively utilizing flattened fields within TSDB enabled data streams. 

## Checklist

## Related issues

- Closes #584

### DIFF
--- a/code/go/internal/validator/semantic/types.go
+++ b/code/go/internal/validator/semantic/types.go
@@ -92,13 +92,14 @@ func (r *runtimeField) UnmarshalYAML(value *yaml.Node) error {
 }
 
 type field struct {
-	Name       string `yaml:"name"`
-	Type       string `yaml:"type"`
-	Unit       string `yaml:"unit"`
-	DateFormat string `yaml:"date_format"`
-	MetricType string `yaml:"metric_type"`
-	Dimension  bool   `yaml:"dimension"`
-	External   string `yaml:"external"`
+	Name       string   `yaml:"name"`
+	Type       string   `yaml:"type"`
+	Unit       string   `yaml:"unit"`
+	DateFormat string   `yaml:"date_format"`
+	MetricType string   `yaml:"metric_type"`
+	Dimension  bool     `yaml:"dimension"`
+	Dimensions []string `yaml:"dimensions"`
+	External   string   `yaml:"external"`
 
 	Runtime runtimeField `yaml:"runtime"`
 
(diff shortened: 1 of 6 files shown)


code/go/internal/validator/semantic/types.go
```diff
@@ -92,13 +92,14 @@ func (r *runtimeField) UnmarshalYAML(value *yaml.Node) error {
...
-	MetricType string `yaml:"metric_type"`
-	Dimension  bool   `yaml:"dimension"`
-	External   string `yaml:"external"`
+	Name       string   `yaml:"name"`
+	Type       string   `yaml:"type"`
+	Unit       string   `yaml:"unit"`
+	DateFormat string   `yaml:"date_format"`
+	MetricType string   `yaml:"metric_type"`
+	Dimension  bool     `yaml:"dimension"`
+	Dimensions []string `yaml:"dimensions"`
```
agithomas, line 101: The resemblance of `dimension` and `dimensions` are very close. Could there be a better name so that developers attempt for an incorrect usage? Unfortunately, presently, i can't think of a better name either.

agithomas:
Please update the [TSDB developer's guideline document](https://github.com/elastic/integrations/blob/main/docs/developer_tsdb_migration_guidelines.md) with a separate PR.

jsoriano:
Discussed about this offline, and I think I am missing a good use case for this feature in packages.

I see these use cases:
* Sets of unknown sub fields -> Use the `flattened` type.
  * If they all need to be dimensions -> Obtain a `fingerprint` from all of them, and use this as dimension.
  * If only some of them need to be dimensions -> Move the known ones to well-known fields, and setup them as dimensions.
* Sets of known sub fields -> Define all them as normal keyword fields, as dimensions if needed.

I see this could help for the case of an unknown set of sub fields where some of them need to be dimensions, but if these known fields are relevant, why not putting them in their own fields, where they can be documented, known beforehand and so on?

In any case, maybe I am missing an specific use case for this.

Issue #559, opened 2023-07-04
[Change Proposal] Identify and check additional features that may change between stack versions


jsoriano:
To be able to use checks based on the package spec versions, we need to be able to model as many features as possible in each package.

Try to identify and check the use of additional features, so we can control what version of the package spec starts including them:
- [x] Ingest processors used in ingest pipelines.
  - https://github.com/elastic/package-spec/pull/586
- [ ] ~~Aggregations used in dashboard queries.~~
  - Would be to find operationTypes... not sure if very reliable.
- [x] Elasticsearch settings.
  - https://github.com/elastic/package-spec/pull/587

This may need to be part of package spec v3 as may be breaking for some cases.

jsoriano:
Closing this by now, though we may add more checks for dashboards in the road to v3.

Issue #539, opened 2023-06-14
Package Spec v3


jsoriano:
We are starting to discuss about some potentially breaking changes that we would like to include, let's start to list them to prepare a new major. 

```[tasklist]
### Tasks
- [ ] https://github.com/elastic/elastic-package/issues/1435
- [ ] https://github.com/elastic/package-spec/issues/538
- [x] Validation of dangling kibana objects https://github.com/elastic/package-spec/blob/450bd34f8bf3bb1274dd01a0c2ff4d33ad4e90d0/code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go#L15 https://github.com/elastic/package-spec/pull/589
- [ ] https://github.com/elastic/package-spec/issues/559
- [x] Try to fix consistent rendering of empty groups in docs, see https://github.com/elastic/elastic-package/pull/1335#discussion_r1248039405
- [ ] https://github.com/elastic/package-spec/issues/317
- [ ] https://github.com/elastic/package-spec/issues/596
- [ ] https://github.com/elastic/ingest-dev/issues/2174
- [ ] https://github.com/elastic/elastic-package/pull/1464
- [x] Remove ValidationErrors from public validator package https://github.com/elastic/package-spec/pull/625
- [ ] https://github.com/elastic/elastic-package/issues/1470
- [ ] https://github.com/elastic/package-spec/issues/624
- [ ] https://github.com/elastic/package-spec/pull/629
- [ ] https://github.com/elastic/package-spec/pull/626
- [ ] https://github.com/elastic/package-spec/pull/635
- [ ] https://github.com/elastic/package-spec/pull/636
- [ ] https://github.com/elastic/package-spec/pull/613
```

jsoriano:
test integrations

jlind23:
@jsoriano now that https://github.com/elastic/elastic-package/issues/1470 is on its way, what else would be missing here to consider this as done?

jsoriano:
@jlind23 we have a couple of small last-minute PRs, and we have to do the actual release.

PR #586, opened 2023-09-06
Add validation for allowed processors in ingest pipelines

jsoriano:
## What does this PR do?

Restrict the processors allowed in ingest pipelines.

## Why is it important?

Package compatibility with the stack is going to be controlled by the package spec version. Each version of Kibana will support a range of versions of the spec, so the spec needs to be specific about the features each package can use.

## Checklist

- [x] I have added a condition so this validation is only enabled on v3.

## Related issues

- Part of https://github.com/elastic/package-spec/issues/539
- Part of https://github.com/elastic/package-spec/issues/559


jsoriano:
test integrations

mrodm (approved):
Great! :+1:

PR #587, opened 2023-09-07
Stricter validation of data stream settings

jsoriano:
## What does this PR do?

Limit the index settings that can be defined in data stream manifests.

## Why is it important?

Package compatibility with the stack is going to be controlled by the package spec version. Each version of Kibana will support a range of versions of the spec, so the spec needs to be specific about the features each package can use.

## Checklist

- [x] I have added a patch so this only applies to v3.

## Related issues

- Part of https://github.com/elastic/package-spec/issues/539
- Part of https://github.com/elastic/package-spec/issues/559


jsoriano:
test integrations

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -159,9 +159,142 @@ spec:
...
+                type: string
+                examples:
+                  - ["strict_date_optional_time","yyyy/MM/dd HH:mm:ss Z||yyyy/MM/dd Z"] # This is the default if not set.
+                  - ["MM/dd/yyyy"]
+
+            dynamic_templates:
+              type: array
+              items:
+                type: object
+                # maxProperties: 1 # No idea why this doesn't work. But this will be also checked by Elasticsearch.
```
mrodm, line 265: Tried to use this value, but I get this error:
> could not read root folder spec file: could not load spec for "integration/data_stream/spec.yml": could not load schema for "integration/data_stream": failed to load schema for "integration/data_stream/manifest.spec.yml": maxProperties must be of an integer

I didn't find why it could be detected as a different type than integer :/

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -159,9 +159,142 @@ spec:
...
+                        type:
+                          $ref: "./fields/fields.spec.yml#/items/properties/type"
+                        scaling_factor:
+                          $ref: "./fields/fields.spec.yml#/items/properties/scaling_factor"
+                        metrics:
+                          $ref: "./fields/fields.spec.yml#/items/properties/metrics"
+                        default_metric:
+                          $ref: "./fields/fields.spec.yml#/items/properties/default_metric"
+                        ignore_above:
+                          $ref: "./fields/fields.spec.yml#/items/properties/ignore_above"
```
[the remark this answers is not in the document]
jsoriano: Added exception for these fields in https://github.com/elastic/package-spec/pull/587/commits/d946169cfcc9772a023f198c3e8b4ea20c70ee20.

spec/integration/data_stream/manifest.spec.yml
jsoriano: Workaround added in https://github.com/elastic/package-spec/pull/587/commits/2d31fd8249471108e49d3710754e1816e0a0bf92

mrodm: Thanks for adding this!

mrodm (approved):
Thanks!

:rocket:

PR #589, opened 2023-09-11
Add validation dangling objects

mrodm:
## What does this PR do?

This PR adds a new validation rule to be applied in v3 to check dangling object IDs.

## Why is it important?

This new validation rule ensures that all references listed in dashboards or other assets are valid ones. That means that there is an asset file with the same ID and type in the package.

## Checklist

## Related issues

- Part of #539


mrodm:
test integrations

jsoriano:
Is it possible that it is ok to reference these fleet managed tags?

code/go/internal/validator/spec.go
```diff
@@ -133,6 +133,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateProfilingNonGA, types: []string{"integration"}},
 		{fn: semantic.ValidateKibanaObjectIDs, types: []string{"integration"}},
 		{fn: semantic.ValidateRoutingRulesAndDataset, types: []string{"integration"}, since: semver.MustParse("2.9.0")},
+		{fn: semantic.ValidateKibanaNoDanglingObjectIDs, since: semver.MustParse("3.0.0")},
```
mrodm, line 136: Just to be applied since v3.0.0

mrodm:
> Is it possible that it is ok to reference these fleet managed tags?

@jsoriano Not sure about this, in case of elastic-package when it exports dashboards it was set to remove those references from the dashboard too in this PR https://github.com/elastic/elastic-package/pull/1206.

Currently, if it is tried to export those dashboards (for instance from airflow package), it is raised this error (as a note the installation does not fail):

```shell
 $ elastic-package-v0.86.1 export dashboards --allow-snapshot
Export Kibana dashboards
Warning: exporting dashboards from a SNAPSHOT version of Kibana (8.10.0-SNAPSHOT) is discouraged. It could lead to invalid dashboards (for example if they use features that are reverted or modified before the final release)
? Which dashboards would you like to export? [Metrics Airflow] Overview (ID: airflow-a3aa42d0-a465-11ed-9ff0-ab4dd59e4c75)
Error: dashboards export failed: exporting dashboards using Kibana client failed: at least Kibana object returned an error: [0] Saved object [tag/airflow-fleet-managed-default] not found
[1] Saved object [tag/airflow-fleet-pkg-airflow-default] not found
```

Probably because elastic-package also adds the package name as part of one of their transform operations: https://github.com/elastic/elastic-package/blob/e1157add2447f28f135fdb055aa396fce8f9c3b5/internal/export/transform_standardize.go#L128-L141

Those fleet tags are always created when the package is installed ("managed" and "<package_name>").

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
```diff
@@ -4,14 +4,170 @@
...
+		for _, installed := range installedIDs {
+			if reference.objectID != installed.objectID {
+				continue
+			}
+			if reference.objectType != installed.objectType {
+				continue
+			}
+
+			found = true
+		}
```
jsoriano, line 79: Nit, as the struct seems to be comparable, maybe you can use [`slices.Contains`](https://pkg.go.dev/slices#Contains).
```suggestion
		found := slices.Contains(installedIDs, reference)
```

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
```diff
@@ -4,14 +4,170 @@
...
+	if err != nil {
+		return []reference{}, fmt.Errorf("unable to convert references: %w", err)
+	}
+
+	if len(allReferences) == 0 {
+		return []reference{}, nil
+	}
+
+	var references []reference
+	for _, reference := range allReferences {
```
jsoriano, line 157: Nit. Also here maybe you can use some `slices` helper.

test/packages/good/kibana/dashboard/good-dashboard-abc-1.json
```diff
@@ -1,3 +1,4 @@
 {
-  "id": "good-dashboard-abc-1"
+  "id": "good-dashboard-abc-1",
+  "type": "dashboard"
```
jsoriano, line 3: Why these files didn't have `type`s? Were they crafted files for tests?

jsoriano (approved):
:+1:

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
mrodm: I'll check it, but there is a third parameter `filePath` in the struct that has not the same value between installedIDs and referencedIDs

test/packages/good/kibana/dashboard/good-dashboard-abc-1.json
mrodm: Yes I think those assets were crafted with the minimum values for the tests. At that moment, it was just the `id` field.

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
mrodm: Changed to use `slices.Contains` :+1:

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
mrodm: Changed to use `slice.ContainsFunc` to use those specific conditions.

Using `slice.Conditions` make the tests fail.

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
```diff
@@ -12,6 +12,7 @@ import (
 	ve "github.com/elastic/package-spec/v2/code/go/internal/errors"
 	"github.com/elastic/package-spec/v2/code/go/internal/fspath"
 	"github.com/elastic/package-spec/v2/code/go/internal/pkgpath"
+	"golang.org/x/exp/slices"
```
jsoriano, line 15: Nit. Import order.

jsoriano (dismissed):
:+1:

code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go
mrodm: I trusted the IDE xD 
I'll check it, sure!

PR #591, opened 2023-09-12
Disallow dotted fields

### LINKED ISSUE #538: [Proposal] Disallow dotted YAML key names
I propose that package-spec disallows dotted key names in the YAML used to define packages. Allowing dotted key names leads to ambiguous specifications. It also increases the complexity of _properly_ parsing the specification because the parse needs to account for fields that maybe defined in more than one way. In the two ambiguous examples below the specification is clear in that both `kibana` and `elasticsearch` are to be objects. [^1] [^2] But from a validation point of view these are accepted by `elastic-package`.

```yaml
conditions:
  kibana.version: '^8.8.0'

  kibana:
    version: '^8.10.0'
```

```yaml
elasticsearch.dynamic_dataset: true

elasticsearch:
  dynamic_dataset: false
```

There are many examples of this occurring in packages:
 - https://github.com/elastic/integrations/blob/835c3c7cee4bef9aea55e831541d22803125f292/packages/system/data_stream/syslog/manifest.yml#L40-L42
- https://github.com/elastic/integrations/blob/835c3c7cee4bef9aea55e831541d22803125f292/packages/system/manifest.yml#L11-L12

[^1]: https://github.com/elastic/package-spec/blob/18849a4d552c24ed5fc4626365f6d8d13501bec9/spec/integration/manifest.spec.yml#L104-L107
[^2]: https://github.com/elastic/package-spec/blob/54365210c7d800b662c520668fb79a12d1ee26ef/spec/integration/data_stream/manifest.spec.yml#L264-L267

mrodm:
## What does this PR do?

This PR disallows to use fields with dots. For instance:
```yaml
conditions:
  kibana.version: ^8.8.0
```

Those files should be updated to:
```yaml
conditions:
  kibana:
    version: ^8.8.0
```

## Why is it important?

Allowing dotted fields leads to ambiguous specifications.

## Checklist

## Related issues

- Closes #538 
- Part of #539


code/go/internal/yamlschema/loader.go
```diff
@@ -72,24 +73,26 @@ func (s *FileSchema) Validate(fsys fs.FS, filePath string) ve.ValidationErrors {
 	return nil // item content is valid according to the loaded schema
 }
 
-func loadItemSchema(fsys fs.FS, path string, contentType *spectypes.ContentType) ([]byte, error) {
+func loadItemSchema(fsys fs.FS, path string, contentType *spectypes.ContentType, specVersion *semver.Version) ([]byte, error) {
```
jsoriano, line 76: No need to pass by pointer here, right?
```suggestion
func loadItemSchema(fsys fs.FS, path string, contentType *spectypes.ContentType, specVersion semver.Version) ([]byte, error) {
```

code/go/internal/yamlschema/loader.go
```diff
@@ -72,24 +73,26 @@ func (s *FileSchema) Validate(fsys fs.FS, filePath string) ve.ValidationErrors {
...
 	data, err := fs.ReadFile(fsys, path)
 	if err != nil {
-		return nil, ve.ValidationErrors{errors.Wrap(err, "reading item file failed")}
+		return nil, ve.ValidationErrors{fmt.Errorf("reading item file failed: %w", err)}
 	}
 	if contentType != nil && contentType.MediaType == "application/x-yaml" {
-		return convertYAMLToJSON(data)
+		return convertYAMLToJSON(data, specVersion.LessThan(semver.MustParse("3.0.0")))
```
jsoriano, line 82: Nit. Consider extracting the parsed version to a constant.

jsoriano (dismissed):
Looks good, added a couple of nitpicks.

jsoriano (approved):
:boat:

andrewkroh:
Awesome! 🏆

Issue #537, opened 2023-06-13
Add a "Need to run as root" option in package spec


jlind23:
We are working towards providing an option  for Elastic Agent to run without superuser privileges. we stumbled over a couple of use cases where integrations need to be run as root in order to access some system metrics and others admin settings.

In order to automatically know what integrations need root permission we should add an option in the package spec manifest just as follow:

```format_version: 1.0.0
name: system
title: System
version: 1.33.0
license: basic
description: Collect system logs and metrics from your servers with Elastic Agent (TSDB Beta).
type: integration
categories:
  - os_system
release: ga
conditions:
  kibana.version: '^8.8.0'
  **need.root: 'true'**
screenshots:
  - src: /img/kibana-system.png
 ```

jsoriano:
Discussed about this with Julien, and we see this more like an agent confinguration that could go in a new block in the manifest, rather than a condition.

jlind23:
@nimarezainia Do we already have a kibana follow up issue for this? As we discussed we should at least display a a information message when users install integrations that needs "root" privileges.

jsoriano:
@nimarezainia @jlind23 for integration packages, is this needed at the package level, or at the data stream level? For example the cpu and memory data streams of the system packages don't require root, and can be valuable on their own on unprivileged deployments.

jlind23:
@jsoriano I wonder how users will react if some datastreams are not working/populated? I think this should be applied at the package level but shouldn't prevent the user from installing a package.

cmacknz:
> So it is more a question about the granularity we want to have, and if we want to support use cases like running data streams that don't require root privileges, but are included in packages that contain data streams that require them.

I think we need both package level and data_stream granularity. For Elastic Defend we want to require root at the package level, for the system integration we'd want it at the data stream level.

jsoriano:
@cmacknz sounds good :+1: Thanks.

nimarezainia:
> @nimarezainia Do we already have a kibana follow up issue for this? As we discussed we should at least display a a information message when users install integrations that needs "root" privileges.

looks like we don't. I created https://github.com/elastic/kibana/issues/166784
had no choice but to include in sp18 should be a small change.

PR #605, opened 2023-09-18
Add definition to declare if a package or data stream uses root privileges

jsoriano:
## What does this PR do?

Add a new setting for packages and data streams to declare that they use root privileges.

Packages or data streams that need to be run as root, would need to add this to their manifest:
```
agent:
  privileges:
    root: true
```

## Why is it important?

This is intended to inform users that a package or data stream needs to be run with root privileges, and/or to indicate the agent to grant these privileges.

## Checklist

## Related issues

- Close https://github.com/elastic/package-spec/issues/537.


jlind23 (dismissed):
LGTM

test/packages/good_v2/manifest.yml
```diff
@@ -1,4 +1,4 @@
-format_version: 2.10.0
+format_version: 2.12.0
```
mrodm, line 1: Should it be added here in `good_v2` also the new `agent` block ?

```yaml
agent:
  privileges:
    root: true
```

mrodm (approved):

test/packages/good_v2/manifest.yml
jsoriano: Added :+1:

jlind23 (approved):

Issue #596, opened 2023-09-12
[Change Proposal] Check that dashboards have some filter defined


jsoriano:
We [recommend to include filters in visualizations](https://github.com/elastic/integrations/blob/888f0df174e7e881a684f593f244b4aa8ffb2cab/docs/dashboard_guidelines.md#visualizations-should-contain-a-filter), so only data related to an integration is queried when using dashboards included in their packages.

Most of the packages could filter based on their dataset, but there are some integrations that allow to configure it. But in a general sense it is probably safe to assume that all dashboards could include a filter that could benefit its use.

PR #607, opened 2023-09-20
Require filters in dashboards

jsoriano:
## What does this PR do?

Fails validation if a dashboard doesn't have a filter.

## Why is it important?

See https://github.com/elastic/package-spec/issues/596

Dashboards without filters may be doing queries on unnecessarily large sets of data. Requiring
filters in dashboards will produce dashboards that limit the queries to the data of the data sets
collected by the package where they are included.

We don't require an specific filter to still give some liberty on the kind of filter that better suite each
dashboard included in a package.

## Checklist

## Related issues

- Closes https://github.com/elastic/package-spec/issues/596

### DIFF
--- a/code/go/internal/validator/semantic/validate_kibana_filter_present.go
+++ b/code/go/internal/validator/semantic/validate_kibana_filter_present.go
@@ -1,0 +1,62 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package semantic
+
+import (
+	"errors"
+	"fmt"
+	"path"
+
+	"github.com/mitchellh/mapstructure"
+
+	ve "github.com/elastic/package-spec/v2/code/go/internal/errors"
+	"github.com/elastic/package-spec/v2/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v2/code/go/internal/pkgpath"
+)
+
+// ValidateKibanaFilterPresent checks that all the dashboards included in a package
+// contain a filter, so only data related to its datasets is queried.
+func ValidateKibanaFilterPresent(fsys fspath.FS) ve.ValidationErrors {
+	var errs ve.ValidationErrors
+
+	filePaths := path.Join("kibana", "dashboard", "*.json")
+	dashboardFiles, err := pkgpath.Files(fsys, filePaths)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("error finding Kibana dashboard files: %w", err))
+		return errs
+	}
+	for _, file := range dashboardFiles {
+		err = checkDashboardHasFilter(file)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("file \"%s\" is invalid: expected filter in dashboard: %w", fsys.Path(file.Path()), err))
+		}
+	}
+
+	return errs
+}
+
+func checkDashboardHasFilter(file pkgpath.File) error {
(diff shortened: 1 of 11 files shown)


code/go/internal/validator/semantic/validate_kibana_filter_present_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+package semantic
+
+import (
+	"testing"
+
+	"github.com/elastic/package-spec/v2/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v2/code/go/internal/pkgpath"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
mrodm, line 13: Nit. Should these two imports be up ?

code/go/internal/validator/semantic/validate_kibana_filter_present.go
```diff
@@ -0,0 +1,62 @@
...
+		} `mapstructure:"query"`
+	}
+	err = mapstructure.Decode(searchJSON, &search)
+	if err != nil {
+		return fmt.Errorf("unable to decode search definition: %w", err)
+	}
+
+	if len(search.Filter) == 0 && len(search.Query.Query) == 0 {
+		return errors.New("no filter found")
+	}
```
mrodm, line 59: So at least a filter or a query should be present in the dashboard, is that right?

In an issue related, it was mentioned that it would be preferable to keep the search bar for the users: https://github.com/elastic/integrations/issues/4936#issuecomment-1387095936

But if we do not allow any query neither, maybe would the number of packages failing be greater ? Would it be too restrictive ?

Any recommendation about that @drewdaemon ?

code/go/internal/validator/spec.go
```diff
@@ -135,6 +135,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateKibanaObjectIDs, types: []string{"integration"}},
 		{fn: semantic.ValidateRoutingRulesAndDataset, types: []string{"integration"}, since: semver.MustParse("2.9.0")},
 		{fn: semantic.ValidateKibanaNoDanglingObjectIDs, since: semver.MustParse("3.0.0")},
+		{fn: semantic.ValidateKibanaFilterPresent, since: semver.MustParse("3.0.0")},
```
mrodm, line 138: For curiosity, if this condition about `since: semver.MustParse("3.0.0")` is removed, are there many packages failing the check ?

code/go/internal/validator/semantic/validate_kibana_filter_present.go
jsoriano: So I should allow only the filters and not the query? Sounds good to me. It will imply changes in more packages though.

jsoriano: Changed to report a message also if there is a query, but not a filter. With different error messages.

mrodm (approved):

code/go/internal/validator/semantic/validate_kibana_filter_present.go
drewdaemon: Kibana's UI presents both a query bar and filter pills, but they accomplish basically the same thing. A Kibana query is simply converted into a `filter` clause in the Elasticsearch DSL.

If we're adding this check to encourage more performant dashboards by prompting integration authors to scope down the underlying Elasticsearch queries, then I think that either a query or a filter would be acceptable.

jsoriano: With current code we would only allow a filter, and the query is expected to be empty. So users can use the queries for quick filtering without risk of altering the provided filters. And this way we will also provide a more consistent experience between different dashboards.

Issue #558, opened 2023-07-04
[Change Proposal] Support disabling validation rules per package


jsoriano:
Add support to disable validation rules per package. This will help to disable certain validations on exceptional cases, or till some feature is fully supported.

Related to #342.

mrodm:
Changes introduced in package-spec to support this must be added into:
- [ ] package-storage-infra: add support to skip the required errors defined in each package when running the publishing job.
- [ ] elastic-package: add support to validate and filter errors according to the errors configuration file.

mrodm:
Closing this issue since elastic-package and package-storage-infra are already using filtering based on codes.

PR #609, opened 2023-09-21
Add support to filter errors using codes

mrodm:
## What does this PR do?

This PR adds the required support to add processors to filter errors.

Currently, this filtering takes into account just codes assigned to some specific errors. Those errors that could be skipped will show the code to be used in the message itself. Example of these error messages:
```
1. file "/home/mariorodriguez/Coding/work/integrations/packages/coredns/kibana/dashboard/coredns-8a1ad3ad-2439-452f-ae43-9e10fc076b7c.json" is invalid: expected filter in dashboard: no filter found (SVR00002)
```

Example of errors configuration that a package could define at the root of the repository (`errors.yml`):
```yaml
errors:
  exclude_checks:
    - SVR00001
    - EP0001
```

In this PR it has been added codes for the errors raised by the semantic validation rules:
- ValidateKibanaFilterPresent - https://github.com/elastic/package-spec/blob/617a1a150b2a00405b4907adb45609a1fada6c42/code/go/internal/validator/semantic/validate_kibana_filter_present.go#L21
- ValidateKibanaNoDanglingObjectIDs - https://github.com/elastic/package-spec/blob/617a1a150b2a00405b4907adb45609a1fada6c42/code/go/internal/validator/semantic/validate_kibana_no_dangling_object_ids.go#L33
## Why is it important?

This way of managing the errors gives us the opportunity to take actions over the errors, and probably dismiss some errors.

## Checklist

## Related issues

- Relates https://github.com/elastic/package-spec/issues/558


spec/integration/spec.yml
```diff
@@ -66,3 +66,9 @@ spec:
     name: elasticsearch
     required: false
     $ref: "./elasticsearch/spec.yml"
+  - description: Errors configuration file
+    type: file
+    contentMediaType: "application/x-yaml"
+    name: "errors.yml"
```
jsoriano, line 72: I don't like the name "errors" so much, as this is going to end up in built packages. Would it be an option something like `lint.conf/yaml`? Or `validation.yml`?

spec/integration/spec.yml
```diff
@@ -66,3 +66,9 @@ spec:
     name: elasticsearch
     required: false
     $ref: "./elasticsearch/spec.yml"
+  - description: Errors configuration file
```
jsoriano, line 69: Nit. Add a more complete description, and maybe mention that the file is only intended for package validation and should be ignored when installing or using the package.

That they are "errors" is more a Go implementation detail, but they are more like the validation result for a user.

code/go/pkg/errors/processors/processor.go
```diff
@@ -0,0 +1,13 @@
...
+
+package processors
+
+import "github.com/elastic/package-spec/v2/code/go/pkg/errors"
+
+// Processor is the interface for every lint processor.
+type Processor interface {
+	Process(errors.ValidationErrors) (errors.ValidationErrors, errors.ValidationErrors, error)
+	Name() string
+}
```
jsoriano, line 13: Nit. Use names here for self-documentation.
```suggestion
type Processor interface {
	Process(errors.ValidationErrors) (processed errors.ValidationErrors, removed errors.ValidationErrors, error)
	Name() string
}
```

Or create an specific struct to return as we discussed at some point:
```suggestion
type Processor interface {
	Process(errors.ValidationErrors) (ProcessResult, error)
	Name() string
}

type ProcessResult struct {
	Processed errors.ValidationErrors
	Removed errors.ValidationErrors
}
```

I would use the `Processed`/`Removed` wording if we keep this as a general `Processor`. If we change it to something like a filter, we could talk about `Included`/`Excluded`.

jsoriano (commented):
General logic looks great, added some suggestions about the interfaces.

code/go/internal/validator/folder_item_spec.go
```diff
@@ -9,8 +9,8 @@ import (
 	"io/fs"
 	"regexp"
 
-	ve "github.com/elastic/package-spec/v2/code/go/internal/errors"
 	"github.com/elastic/package-spec/v2/code/go/internal/spectypes"
+	ve "github.com/elastic/package-spec/v2/code/go/pkg/errors"
```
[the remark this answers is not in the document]
jsoriano: Or, given that these errors are usually used in the context of the public `validation` package, and are prefixed by `Validation`, wdyt about adding all these errors and associated logic to the public `validation` package, and remove the `Validation` prefixes?

Or could this produce some dependency loop?

code/go/pkg/errors/processors/processor.go
mrodm: I'll go for the second option, creating an specific struct for these results :+1:

spec/integration/spec.yml
mrodm: Updated description with more details.

spec/integration/spec.yml
mrodm: I don't find any other better names. I opted for `validation.yml`

code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go
```diff
@@ -33,33 +33,33 @@ func ValidateKibanaObjectIDs(fsys fspath.FS) ve.ValidationErrors {
...
 			if err != nil {
-				errs = append(errs, fmt.Errorf("unable to get rule ID in file [%s]: %w", fsys.Path(filePath), err))
+				errs = append(errs, ve.NewStructuredError(fmt.Errorf("unable to get rule ID in file [%s]: %w", fsys.Path(filePath), err), ve.UnassignedCode))
 				continue
 			}
 
 			objectIDValue, ok := objectID.(string)
 			if !ok {
-				errs = append(errs, fmt.Errorf("expect object ID to be a string: %w", err))
+				errs = append(errs, ve.NewStructuredError(fmt.Errorf("expect object ID to be a string: %w", err), ve.UnassignedCode))
```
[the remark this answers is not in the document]
mrodm: If we add new errors in following PRs, those could not use that way of defining errors
```golang
func Errorf(format string, a ...any) *StructuredError {
    return StructuredErrorf(UnassignedCode, format, a...)
}
```

but I guess as this `StructuredError` is going to be the generic error, it's ok that the Errorf() method would return one of these, would that be ok ? @jsoriano

code/go/internal/validator/folder_item_spec.go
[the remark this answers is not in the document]
mrodm: If it is moved `errors.go`  with the logic and structs to the `validator` package, `import cycle` errors are raised.

It could be moved the whole `errors` package to the `validator` package. But in that case, we maybe have to keep the Validation prefixes in the structs?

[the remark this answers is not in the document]
mrodm: Rename package to be `specerrors`

code/go/internal/validator/semantic/validate_kibana_matching_object_ids.go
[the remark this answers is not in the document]
mrodm: Added the first option `NewStructuredErrorf` for now. It helps to remove a lot of usages of UnassignedCode, thanks!

code/go/pkg/specerrors/errors.go
```diff
@@ -0,0 +1,78 @@
...
+type ValidationSeverityError interface { // TODO no validation error using this interface yet
+	// File returns the file path where the error was raised.
+	Severity() string
+}
+
+// ValidationErrors is an error that contains an iterable collection of validation error messages.
+type ValidationErrors []ValidationError
+
+// Collect filters the validation errors using the function given as a parameter.
+func (ve ValidationErrors) Collect(collect func(elem ValidationError) bool) (ValidationErrors, ValidationErrors) {
```
jsoriano, line 36: Nit. Use names here to document what is each ValidationErrors, or use a struct similar to ProcessResult.
```suggestion
func (ve ValidationErrors) Collect(collect func(elem ValidationError) bool) (collected ValidationErrors, excluded ValidationErrors) {
```

jsoriano (dismissed):
:+1: :+1:

jsoriano (approved):
:+1:

PR #613, opened 2023-09-26
Fix validation of filters in dashboards with encoded search

jsoriano:
Dashboards in built packages have the search configuration encoded in JSON,
support looking for filters also in this format.

This fixes issues like the following one:
```
file "..." is invalid: expected filter in dashboard: unable to decode search definition: '' expected a map, got 'string'
```

Thanks @mrodm for catching this one!


spec/changelog.yml
```diff
@@ -24,9 +24,9 @@
     link: https://github.com/elastic/package-spec/pull/607
 - version: 2.12.1-next
   changes:
-  - description: Prepare for next version
-    type: enhancement
-    link: https://github.com/elastic/package-spec/pull/612
+  - description: Fix validation of filters in built dashboards
+    type: bugfix
+    link: https://github.com/elastic/package-spec/pull/613
```
mrodm, line 29: As this is for v3.0.0 , should it be moved to the 3.0.0-next block ?
In v2.12 or previous versions, that rule is not checked.

mrodm (dismissed):
LGTM
Just a comment about where to add the changelog entry

spec/changelog.yml
jsoriano: You are right, and as this hasn't been released yet I won't add any changelog entry. Thanks!

mrodm (approved):

PR #610, opened 2023-09-22
Disallow legacy visualizations

drewdaemon:
## What does this PR do?

Fails when any legacy Kibana visualizations are detected in a package.

## Why is it important?
We would like to migrate away from legacy visualizations to give our customers a better experience and decrease the maintenance burden of so many editors.

## Checklist

- [x] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.
- [x] I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).

## Related issues

- Closes https://github.com/elastic/package-spec/issues/317

PR #619, opened 2023-09-28
allow legacy input controls

drewdaemon:
## What does this PR do?

#610 disallowed legacy visualizations. We're making an exception for input controls because they aren't quite ready to be deprecated and because some package-spec v3 integrations are already using them.

## Why is it important?

To get package-spec 3 released!

## Checklist


mrodm:
test integrations

mrodm:
Created manually PR https://github.com/elastic/integrations/pull/8012 to test integrations

mrodm (approved):
LGTM
Triggered a PR in integrations to check that those packages are not failing now.

Issue #624, opened 2023-09-29
[Validation] Detect incorrectly defined dynamic mappings


andrewkroh:
Detect cases where a developer thinks they have declared a dynamic mapping, but have not. [Example](https://github.com/elastic/integrations/blob/3c88555054fd24a461babebe624d8645b5feea6c/packages/linux/data_stream/raid/fields/fields.yml#L35-L38):

```
    - name: disks.states.*
      type: object
      description: |
        map of raw disk states
```

The rule would detect fields where

- `name` contains an asterisk
- `type == "object"`
- `object_type` is not set

Without specifying an `object_type` the result index template will declare a mapping for a literal name containing `*` asterisks and that is unexpected/surprising/wrong.

```json
            "disks": {
              "properties": {
                "states": {
                  "properties": {
                    "*": {
                      "type": "object"
                    }
                  }
                }
              }
            },
```

andrewkroh:
As of https://github.com/elastic/integrations/commit/eaa7acff04f4f62d6343d542fbef4e3e9f220994 this problem occurs 61 times across 12 integrations.

jsoriano:
@andrewkroh thanks for reporting, this actually looks like a definition we shouldn't allow. I wonder if it was assumed that this should behave as if `object_type` would be `keyword`.

I have started https://github.com/elastic/package-spec/pull/628 to fix this.

I am disallowing any case of `type: object` without `object_type`. There can be fields with type object and without wildcards, on these cases it is assumed that the name is the base of the path and `.*` is appended to the `path_match`. Implemented in Fleet in https://github.com/elastic/kibana/pull/137772, in coherence with the behaviour in Beats.

andrewkroh:
I think we are missing a comprehensive guide to declaring dynamic mappings. We need a list of all the conditions that lead to a dynamic mapping being added by Fleet. This will help guide us in deciding what validations to apply. These are case I have observed, and I wonder if there are more.

1. Wildcard name and `type` (excluding `type: object`). In the case Fleet treats the `type` value like the `object_type`.
2. Wildcard name and `type: object` and  `object_type`

> I am disallowing any case of type: object without object_type. There can be fields with type object and without wildcards, on these cases it is assumed that the name is the base of the path and .* is appended to the path_match.

[...]

Issue #393, opened 2022-08-18
[Change Proposal] Properly support dynamic template mappings


jsoriano:
We lacked a proper definition for dynamic template mappings in the package spec, and package developers (and the initial import scripts) assumed to follow the same conventions used in beats. This was not properly supported by the spec or fleet, leading to unexpected mappings (https://github.com/elastic/kibana/issues/129344).

Beats-like behaviour was implemented in Fleet in https://github.com/elastic/kibana/pull/137772, to support the convention currently used.

Additionally, it was assumed that any field with wildcards in its name is also a dynamic template, even if it doesn't have `type: object`, to avoid issues like https://github.com/elastic/beats/issues/32577, more common in integrations than in beats. This was implemented in fleet in https://github.com/elastic/kibana/pull/137978.

Check that these conventions are properly defined in the package spec, and document them in the descriptions of the involved fields.

PR #628, opened 2023-10-02
Require object_type for object fields

jsoriano:
## What does this PR do?

Make `object_type` mandatory for fields of type `object` in Package Spec v3.

## Why is it important?

A field of `type` `object` without `object_type` probably assumes that the type should be `keyword`, but Fleet is producing
invalid mappings (see https://github.com/elastic/package-spec/issues/624). So better to disallow these mappings to remove ambiguities.

## Checklist

## Related issues

- Fixes https://github.com/elastic/package-spec/issues/624
- Closes https://github.com/elastic/package-spec/issues/393


mrodm (approved):
LGTM

PR #629, opened 2023-10-02
Require type group or nested for fields with subfields

jsoriano:
## What does this PR do?

Require type group for fields with subfields.

## Why is it important?

Fields with subfields should not have a definition on their own. This would lead to ambiguity if for
example an object would generate a dynamic mapping, but also definitions for their subelements.

The only cases where a field can contain subfields is when it is of type `group`, or `nested`.

## Checklist

## Related issues

- Requested on conversation with Andrew Kroh after https://github.com/elastic/package-spec/issues/624.


test/packages/bad_fields/data_stream/foo/fields/fields.yml
```diff
@@ -5,3 +5,10 @@
   date_format: yyyy-MM-dd
 - name: object_without_object_type.*
   type: object
+- name: object_with_subfields
+  description: This field should be of type group, because it contains subfields.
+  type: object
+  object_type: keyword
```
andrewkroh, line 11: `object_type` is not necessary here, right?

andrewkroh (approved):

mrodm:
test integrations

test/packages/bad_fields/data_stream/foo/fields/fields.yml
jsoriano: Yeah, it wouldn't be neccesary to produce this validation error, but I added it so it doesn't trigger also the error about missing `object_type` for `object`.

jsoriano:
Thanks Mario for fixing the integrations jobg! From the packages failed in https://github.com/elastic/integrations/pull/8065, noone fails by linting, so I will merge this.

PR #636, opened 2023-10-05
Report warnings as errors by default

jsoriano:
## What does this PR do?

Warnings that could be optionally enabled as errors are reported as errors now, and can be filtered-out. Starting on Package Spec v3.

## Why is it important?

To raise awareness on the issues reported by these warnings, so they are not so easily ignored, but can still be discarded.

## Checklist

## Related issues

- Related to https://github.com/elastic/package-spec/issues/539


mrodm (commented):
I like the idea to change these warnings to be errors that could be filtered.

But it will also require to add in most of the packages defined in the integrations repository the `validation.yml` file including these new codes. And probably (endpoint, apm...). I guess this is affordable.

code/go/internal/validator/folder_spec.go
```diff
@@ -68,8 +69,9 @@ func (v *validator) Validate() specerrors.ValidationErrors {
 			)
 		} else {
 			message := fmt.Sprintf("Warning: package with non-stable semantic version and active beta features (enabled in [%s]) can't be released as stable version.", v.pkg.Path(v.folderPath))
```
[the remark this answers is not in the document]
jsoriano: Correct, this is not run as part of the semantic validators and `WarnOn` only wraps some semantic validators. But what is true is that in the error it shouldn't include the `Warning` prefix. I will change this.

code/go/internal/validator/spec.go
```diff
@@ -145,15 +145,17 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateVersionIntegrity},
 		{fn: semantic.ValidateChangelogLinks},
 		{fn: semantic.ValidatePrerelease},
-		{fn: semantic.ValidateMinimumKibanaVersion},
+		{fn: semantic.WarnOn(semantic.ValidateMinimumKibanaVersion), until: semver.MustParse("3.0.0")},
+		{fn: semantic.ValidateMinimumKibanaVersion, since: semver.MustParse("3.0.0")},
```
mrodm, line 149: Ah! So until 3.0.0 is the same behaviour, that's perfect!
I didn't remember that `until` parameter :+1:

code/go/pkg/specerrors/constants.go
```diff
@@ -9,10 +9,13 @@ const (
 	UnassignedCode = ""
 
 	// PSR - Package Spec [General] Rule
-	CodeNonGASpecOnGAPackage = "PSR00001"
+	CodeNonGASpecOnGAPackage         = "PSR00001"
```
mrodm, line 12: I'd say this code can be removed, it looks like is not used right now

jsoriano: It is used in https://github.com/elastic/package-spec/blob/d0c66e84c6e2ccdfb9ea0811d2f4b85e1c388a96/code/go/internal/validator/spec.go#L78

mrodm: Ah it's from other file out of this PR , sorry. I thought it was a new error code :+1:

mrodm (approved):

PR #639, opened 2023-10-06
Remove deprecated `corpora.generator.size` in favour of `corpora.gene…

aspacca:
…rator.tot_events`

## What does this PR do?

Remove deprecated `corpora.generator.size` in favour of `corpora.generator.tot_events` for `benchmark/spec.yml`
## Why is it important?
Corpus generator tool remove since v0.6.0 `size` in favour of `totEvents`: this is now reflected in the `package-spec` so that we will be able to bump the library. Since it's a breaking-change it would be better to include in `3.0.0`

## Checklist

## Related issues

-


spec/integration/_dev/benchmark/system.scenario.spec.yml
```diff
@@ -65,16 +65,11 @@ spec:
           type: object
           additionalProperties: false
           required:
-            - size
+            - tot_events
```
jsoriano, line 68: Could this be called `total_events`? Or it is too late to rename?
```suggestion
            - total_events
```

spec/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Using non-GA versions of the spec in GA packages produces a filterable validation error instead of a warning.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/627
+  - description: Remove deprecated `corpora.generator.size` in favour of `corpora.generator.tot_events`
```
jsoriano, line 10: Maybe we can still include it in 3.0.0.

spec/integration/_dev/benchmark/system.scenario.spec.yml
aspacca: `tot_events` is already part of the spec, but it's not used anywhere yet but for this PR and one in `elastic-package`.
if we include in the 3.0.0 we can rename taking the opportunity of the breaking change, up to you

jsoriano: If this is not being used, let's rename to `total_events` if possible, as we are doing this a breaking change in any case.

aspacca: done!

spec/integration/_dev/benchmark/system.scenario.spec.yml
```diff
@@ -65,17 +65,12 @@ spec:
...
             - template
             - config
             - fields
           properties:
-            size:
-              description: A string describing the amount of data to generate.
-              type: string
-              example: 10MiB
-              deprecated: true # as for https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/94
-            tot_events:
```
mrodm, line 78: In order to not break current packages using these variables.
Should we add a json patch to at least add `size` parameter for packages using spec version <3.0.0 ?  cc @jsoriano 

Probably, this would be a similar example:
https://github.com/elastic/package-spec/blob/708b8a7049684091cb699fe840c3eb4ea54e59ba/spec/integration/manifest.spec.yml#L465

mrodm: I haven't found any usage of these `benchmark` folders in the integrations repository.

Commands used in the integrations repository:
```shell
 $ find . -type d -name system  | grep benchmark
 $ find . -type d -name benchmark
```

Maybe it's not needed to add a JSON Patch to re-add the `size` parameter.

WDYT ?

aspacca: @mrodm yes, to my knowledge we added the specs but the reference to them is still only it the specs itself and in the `elastic-package` repo

that's why I was positive to not deal with a migration path like you suggested

once we merge this and make part of v3 https://github.com/elastic/elastic-package/pull/1491 will be green and everything will be fine

what do you think?

jsoriano: Yeah, I agree that we don't need the migration path as this is not used yet.

mrodm: > @mrodm yes, to my knowledge we added the specs but the reference to them is still only it the specs itself and in the `elastic-package` repo
> 
> that's why I was positive to not deal with a migration path like you suggested

Perfect, thanks for checking this out. Agreed, it's not need to add the JSON patch :+1:

mrodm (approved):

aspacca:
@jsoriano . @mrodm thanks for the prompt review

do we have time to include this PR in v3.0.0? do we need to release a v3.0.0-rc2?

jsoriano:
> do we have time to include this PR in v3.0.0? do we need to release a v3.0.0-rc2?

Yes, we will include it in v3.0.0, that we will likely release today or tomorrow.

PR #650, opened 2023-10-24
Allow to set 'index: false'  in dynamic templates defined in data streams

jsoriano:
## What does this PR do?

Allow to set `index: false` in dynamic templates defined in data streams.

## Why is it important?

Package Spec v3 introduced validation for fields defined in manifests in data streams.  `index` was not added then, but this is [used by the APM package](https://github.com/elastic/apm-server/blob/5584a8aa2a999a3da3f656de66a15ace58f03665/apmpackage/apm/data_stream/app_metrics/manifest.yml#L33-L36).
In the case of the field in APM, it cannot be migrated to fields definitions because the definition is not based on the name of the field.

## Checklist

## Related issues

- Needed for https://github.com/elastic/apm-server/pull/11822.

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -7,6 +7,9 @@
   - description: Using non-GA versions of the spec in GA packages produces a filterable validation error instead of a warning
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/627
+  - description: 'Allow to set index: false in dynamic templates defined in data stream manifests'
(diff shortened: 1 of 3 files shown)


kyungeunni (approved):
LGTM!

juliaElastic (approved):
LGTM

mrodm (approved):

PR #654, opened 2023-10-25
Assign error code to path_match in manifests

jsoriano:
## What does this PR do?

Assign new error code `JSE00001` to validation errors raised when data streams define dynamic mappings
with path_match in data stream manifests.

## Why is it important?

Since Package Spec v3, it is mandatory to use field files to define fields that can be defined there. This
is the case of dynamic fields with `path_match`.
We have found that packages using these mappings and `dynamic: "runtime"` cannot ingest documents
with the error `Missing intermediate object labels`, likely caused because the parent object doesn't exist
in the mapping.
We should probably make Fleet to create these intermediate objects.
In the meantime, `JSE00001` may offer a escape hatch for packages needing this, like in the test package
added in this PR.

After this change, packages with this issue, can add a `validation.yml` file at the root of the package with
the following content to ignore this error:
```
errors:
  exclude_checks:
    - JSE00001 # Data stream defines a dynamic mapping in manifest that should be defined in field files.
``` 

## Checklist

## Related issues

- This issue blocks https://github.com/elastic/apm-server/pull/11822


mrodm (approved):

jsoriano:
Not needed by now, closing.

PR #662, opened 2023-11-09
Allow parquet files in terraform service deployer

bhapas:
## What does this PR do?

`parquet` files are only allowed for `amazon_security_lake` and to test the integration these files are to be allowed as S3 objects and hence allow them in `files` for terraform deployer.

## Checklist

## Related issues

- Relates elastic/integrations#8303


spec/integration/_dev/deploy/tf/files/spec.yml
```diff
@@ -16,3 +16,7 @@ spec:
     contentMediaType: "application/json"
     pattern: '^*.json$'
     required: false
+  - description: Parquet files to add into terraform resources
```
jsoriano, line 19: What are these parquet files? Could you provide an example in a test package on how it would be used?

bhapas: This is a file format that is used by [Amazon Security Lake](https://docs.aws.amazon.com/security-lake/latest/userguide/source-management.html) and I have added an example file in the test package.

jsoriano: But where is this file used? In terraform resources? Could you add an example about how the file would be used?

bhapas: This is a special format used by Amazon Security Lake and this is just another S3 Object. I added an [example terraform resource](https://github.com/elastic/package-spec/pull/662/files#diff-f1943dc87d54c608af7b9fc30bf056bd7f043601b10cd8385b42a4246d90239f) for this

bhapas: [This](https://github.com/elastic/package-spec/pull/662/files#diff-74a663008ae2600bdb37743ef34dbf6214c4fca82ee0b2ca235a68e2b08b5a42) is the test binary file in parquet format.

jsoriano (approved):
LGTM by now, let's keep an eye in case we need to add something more generic to allow any kind of file.

jsoriano (approved):

PR #665, opened 2023-11-14
Best-effort detection of potentially secret variables

jsoriano:
## What does this PR do?

Fails validation on variables that look like secrets and don't declare if they are secrets or not with the `secret` property.

This detection is based on two things:
* The name of the variable contains strings like `password`, `api_key` or `token`.
* The type of the variable is `password`.

## Why is it important?

To encourage adoption of this feature to secure secrets in package policies.

## Checklist

## Related issues

- Part of https://github.com/elastic/elastic-package/issues/1554


code/go/internal/validator/spec.go
```diff
@@ -118,18 +119,15 @@ func processErrors(errs specerrors.ValidationErrors) specerrors.ValidationErrors
...
 			}
-			if substringInSlice(e.Error(), redundant) {
-				continue
-			}
-			processedErrs = append(processedErrs, e)
 		}
+		if substringInSlice(e.Error(), redundant) {
+			continue
+		}
+		processedErrs = append(processedErrs, e)
```
jsoriano, line 130: Unrelated fix required for this change, but that I could move to a different PR.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -115,39 +115,51 @@ spec:
...
+                 min_items: 1
+           else:
+             not:
+               required:
+                 - options
+         - if:
+             anyOf:
+               - properties:
+                   name:
+                     pattern: "(password|api_key|access_key|token)"
```
jsoriano, line 156: @kpollich @nimarezainia this is the regexp pattern for variable names that we would detect as candidates to be secrets, is there any other string do you think we should add here?
Should we match with exact values instead of with substrings?

code/go/internal/validator/spec.go
jsoriano: Created also separate PR so we merge this fix in any case https://github.com/elastic/package-spec/pull/666

spec/integration/data_stream/manifest.spec.yml
kpollich: We could add the substring `secret` here as it will catch things like an AWS `secret_access_key` value. I think using substring matching is good to keep this generic.

kpollich (approved):
LGTM - thanks for getting this done quickly 🚀

spec/integration/data_stream/manifest.spec.yml
jsoriano: Added `secret` and `passphrase`, the latter is sometimes used for encrypted TLS keys.

jlind23: We should juste be cautious with values like "secret_path" which might be caught by this check though.

jlind23 (approved):

spec/integration/data_stream/manifest.spec.yml
jsoriano: > We should juste be cautious with values like "secret_path" which might be caught by this check though.

Yep, in these cases they will need to set `secret: false`.

kpollich (approved):

code/go/pkg/validator/validator_test.go
```diff
@@ -250,8 +257,13 @@ func TestValidateFile(t *testing.T) {
 
 	filter := specerrors.NewFilter(&specerrors.ConfigFilter{
 		Errors: specerrors.Processors{
-			// TODO: Actually fix the references instead of ignoring the error.
-			ExcludeChecks: []string{"SVR00004"},
+			ExcludeChecks: []string{
+				// Allow to test unreleased features in "good" packages.
+				"PSR00001",
```
mrodm: :+1:

mrodm (approved):
:+1:

code/go/pkg/validator/validator_test.go
```diff
@@ -193,6 +193,13 @@ func TestValidateFile(t *testing.T) {
 				"field vars.0: Additional property secret is not allowed",
 			},
 		},
+		"bad_secret_vars_v3": {
+			"manifest.yml",
+			[]string{
+				"field vars.0: variable identified as possible secret, secret parameter required to be set to true or false",
+				"field vars.1: variable identified as possible secret, secret parameter required to be set to true or false",
```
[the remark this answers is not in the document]
ebeahan: > This is the kind of errors package developers will see when upgrading to Package Spec 3.0.2.

Isn't this an enhancement? I'm surprised a new feature would fall into a patch version.

I'm also hesitant to jump right into _requiring_ `secret` when it's still new functionality  (sure, a dev can set `secret: false` but the way the error's presented, we're encouraging to set `secret: true`). Based on a recent issue I ran into trying to set `secret: true` (see https://github.com/elastic/kibana/issues/154715#issuecomment-1792869195), I think it's worth adopting on a smaller set of integrations first.

Should using `secret` be encouraged, especially with new integrations? Definitely! But _required_? I'm not so sure.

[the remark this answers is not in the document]
jsoriano: Hey @ebeahan,

> I'm surprised a new feature would fall into a patch version.

Starting on Package Spec 3 we are not fully following semver. I understand this can be surprising. We use minors now for compatibility with the stack, and this change doesn't affect the compatibility with the stack, it only enhances validations. Secrets were already supported in the spec since 2.9.0.

It is an enhancement in the sense that it can prevent security risks. We have also considered as enhancements other "breaking" validations that we have introduced in the past to avoid problematic situations. We can reconsider this in any case if needed.

> a dev can set `secret: false` but the way the error's presented, we're encouraging to set `secret: true`

I am open to suggestions regarding the error message.

> Based on a recent issue I ran into trying to set secret: true (see https://github.com/elastic/kibana/issues/154715#issuecomment-1792869195)

This is an issue found in an snapshot build, that was fixed. We can still delay the introduction of this change though, 3.0.2 hasn't been released yet.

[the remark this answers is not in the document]
jsoriano: > Starting on Package Spec 3 we are not fully following semver. I understand this can be surprising. We use minors now for compatibility with the stack, and this change doesn't affect the compatibility with the stack, it only enhances validations. Secrets were already supported in the spec since 2.9.0.

Updating versioning guidelines in the README taking recent changes into account https://github.com/elastic/package-spec/pull/667

Issue #557, opened 2023-07-04
Create compliance suite for implementations of the package spec


jsoriano:
Create a compliance suite that can be used to define expectations on implementations of the package spec. It must answer the question _Does version X of implementation Y support package spec Z?_
We can think on this as a tool to validate implementations in general, but we should target Kibana.

Its main purpose will be to test that a given build of Kibana properly supports a version of the package spec.

It should be possible to define expectations in a BDD-like fashion, so we could have for example:
```Gherkin
Feature: Runtime fields
  Configure fields in a package as runtime fields.

Scenario: Installer leverages runtime parameter
  Given "Installer" is candidate to support "2.8.0".
   When "package-with-runtime-fields" is installed.
   Then index template for "package-with-runtime-fields" includes runtime fields.
```
It should be possible to run these tests against Kibana endpoints, that would represent the role of "Installer".

These tests should reside within the package spec, so features and their expectations are implemented together, but it should be possible to easily run them in other repositories. For example to make them part of Kibana test suites.

As part of this task we should have the framework and an initial set of tests, with a focus on most recent features. It is not expected to have a compliance suite for the whole spec yet.

PR #667, opened 2023-11-15
Update Package Spec version guidelines

jsoriano:
## What does this PR do?

It updates the Package Spec versioning guidelines in the README.

## Why is it important?

We have started to use the Package Spec to determine the compatibility of packages and the stack where they are installed.
With this changes we are reserving compatibility constraints mostly for the compatibility with the stack, that means that a given minor will be always compatible with a version of the stack that was compatible with its 0 patch version, but we may still include new validation rules or other changes in patch versions.

After these changes, it seemed relevant to update the information included in the README about Package Spec versioning.

## Related issues

- https://github.com/elastic/kibana/issues/161210
- https://github.com/elastic/package-spec/issues/559
- https://github.com/elastic/package-spec/issues/557

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,34 +67,72 @@ Note that the specification files primarily define the structure (syntax) of a p
 
 # Specification Versioning
 
-Package specifications are versioned. Versions follow the [semantic versioning](https://semver.org/) scheme. In the context of package specifications, this means the following.
+Package Spec version follows [semantic versioning](https://semver.org) for its
+compatibility with the Stack and only partially for the packages format.
+That means that patch versions may include stricter validations for packages,
+but they should not include support for new features.
+Major versions are reserved for significant changes in the format of the files,
+the structure of packages or the interpretation of the Package Spec.
 
-* Packages must specify the specification version they are using. This is done via the `format_version` property in the package's root `manifest.yml` file. The value of `format_version` must conform to the semantic versioning scheme.
+Packages must specify the specification version they are using. This is done via
+the `format_version` property in the package's root `manifest.yml` file. The value
+of `format_version` must conform to the semantic versioning scheme.
 
-* Specifications are defined by schema files and semantic rules, some attributes or files will only be available since, or till a version.
+Specifications are defined by schema files and semantic rules, some attributes or
+files will only be available since, or till a version.
 
-* Note that the latest version of each major may include a pre-release suffix, `e.g. 1.4.0-alpha1`. This indicates that this version is still under development and may be changed multiple times. Once the pre-relase suffix is removed, however, the specification at that version becomes immutable. Further changes must follow the process outlined below in _Changing a Specification_.
+Note that some versions may include a pre-release suffix, `e.g. 1.4.0-alpha1`. This
+indicates that these versions are still under development and may be changed multiple
+times. These versions in development can be used in prerelease versions of
+packages, but breaking changes can still occur.
+Once the pre-relase suffix is removed, however, the specification at that version becomes
+immutable. Further changes must follow the process outlined below in _Changing a Specification_.
 
 ## Changing a Specification
 
-* Consider a proposal to change the specification in some way. The version number of the changed specification must be determined as follows:
-  * If the proposed change makes the specification stricter than it is at `x.y.z`, the new version number will be `X.0.0`, where `X = x + 1`. That is, we bump up the major version. 
-     * If the change is in a schema file, consider the `spec` the latest
-       version, and add a JSON patch in the `versions` section to support older
-       schemas.
-     * If the change is in semantic rules, add a constraint in the rule, so they only apply on
-       the indicated version range.
-     * Add a changelog entry in the `spec/changelog.yml` file in the section of this major.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,35 +67,73 @@ Note that the specification files primarily define the structure (syntax) of a p
...
-  * If the proposed change does not change the strictness of the specification at `x.y.z`, the new version number will be `x.y.Z`, where `Z = z + 1`. That is, we bump the patch version.
-     * Apply the proposed changes to the existing specification under the `spec` folder.
-     * Set the root-level `version` property in the specification's root `spec.yml` file to `x.y.Z`.
-     * Add a changelog entry in the `spec/changelog.yml` file in the section for
-       this version.
+Consider a proposal to change the specification in some way. The version number
+of the changed specification must be determined as follows:
+
+  * If the proposed change introduces changes in the format of the files that
+    require manual changes in packages, the new version number will be `X.0.0`,
```
mrodm, line 97: Maybe it can be rephrased to not used "changes" 

Maybe something like ? 

- If the proposed change introduces alters in the format of the files that require manual adjustments in packages, the new version
- If the proposed change introduces impacts the format of the files that require manual adjustments in packages, the new version

README.md
```diff
@@ -67,35 +67,73 @@ Note that the specification files primarily define the structure (syntax) of a p
...
+    where `X = x + 1`. That is, we bump up the major version.
+    * Exception: When the proposed change is intended to address existing issues
+      in packages like ambiguous mappings or security risks.
+    * Exception: When the proposed change affects a feature marked as technical
+      preview.
```
mrodm, line 102: These two exceptions would lead to a bump in the patch version, is that right?

What about adding what would be the expected version in those exceptions ?

README.md
```diff
@@ -67,35 +67,73 @@ Note that the specification files primarily define the structure (syntax) of a p
...
-
-* Specifications are defined by schema files and semantic rules, some attributes or files will only be available since, or till a version.
-
-* Note that the latest version of each major may include a pre-release suffix, `e.g. 1.4.0-alpha1`. This indicates that this version is still under development and may be changed multiple times. Once the pre-relase suffix is removed, however, the specification at that version becomes immutable. Further changes must follow the process outlined below in _Changing a Specification_.
+Package Spec version follows [semantic versioning](https://semver.org) for its
+compatibility with the Stack and only partially for the packages format.
+That means that patch versions may include stricter validations for packages,
+but they should not include support for new features.
+Major versions are reserved for significant changes in the format of the files,
+the structure of packages or the interpretation of the Package Spec.
```
mrodm, line 75: :+1:

README.md
```diff
@@ -67,35 +67,73 @@ Note that the specification files primarily define the structure (syntax) of a p
...
+apply on the indicated version range and package types.
+
+Remember to add a changelog entry in `spec/changelog.yml` for any change in the
+spec. If no section exist for the version determined by the above rules, please
+add the new section. Multiple `next` versions may exist at the same moment if
+multiple versions are in development.
 
 ## Version Compatibility between Packages and Specifications
 
 A package specifying its `format_version` as `x.y.z` must be valid against specifications in the semantic version range `[x.y.z, X.0.0)`, where `X = x + 1`.
```
mrodm, line 124: Just double checking about this section, this consideration is still valid about the ranges of valid specifications for a given package, no ? 

Would a package using 3.0.1 as format_version be still valid checking with package-spec 3.1.0 o 3.2.0 versions ? And it will be invalid for 4.0.0. Is that still right?

According to this item I would say there is no need to update this section:
> If the proposed change introduces support for a new feature that requires
    explicit support in the Stack, the new version will be `x.Y.0`, where
    `Y = y + 1`. That is, we bump up the minor version. See note below about
    compatibility between packages and the Stack.

README.md
jsoriano: Rephrased to "If the proposed change modifies the format of the files in a way that require manual changes in packages..."

README.md
jsoriano: >Would a package using 3.0.1 as format_version be still valid checking with package-spec 3.1.0 o 3.2.0 versions ? And it will be invalid for 4.0.0. Is that still right?

Yes, in principle it should be compatible with any 3.x version, but doesn't need to be compatible with 4.x. This is not completely true, neither has been, because we have introduced incompatible changes for bugfixes and so on (the exceptions mentioned above).

README.md
```diff
@@ -67,35 +67,74 @@ Note that the specification files primarily define the structure (syntax) of a p
...
+Consider a proposal to change the specification in some way. The version number
+of the changed specification must be determined as follows:
+
+  * If the proposed change modifies the format of the files in a way that
+    require manual adjustments in packages, the new version number will be `X.0.0`,
+    where `X = x + 1`. That is, we bump up the major version.
+    There are some exceptions, for changes that could be done in patch versions:
+    * When the proposed change is intended to address existing issues
+      in packages like ambiguous mappings or security risks.
+    * Exception: When the proposed change affects a feature marked as technical
```
mrodm, line 102: I think `Exception: ` can be removed from this second bullet item, as it refers also to the same exceptions to be considered as patch versions

```suggestion
    * When the proposed change affects a feature marked as technical
```

README.md
mrodm: Ok, agreed!

mrodm (approved):
LGTM
Just added a suggestion to be considered before merging

mrodm (approved):
:rocket:

PR #668, opened 2023-11-24
Update validation for github owner to allow _

shahzad31:
Update validation for github owner to allow _ !!

follow up to https://github.com/elastic/elastic-package/pull/1564

### DIFF
--- a/code/go/pkg/validator/validator_test.go
+++ b/code/go/pkg/validator/validator_test.go
@@ -78,10 +78,10 @@ func TestValidateFile(t *testing.T) {
 				"field policy_templates.2.data_streams.1: data stream doesn't exist",
 			},
 		},
-		"bad_github_owner": {
+		"bad_github_owner/": {
 			"manifest.yml",
 			[]string{
-				"field owner.github: Does not match pattern '^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\\/[a-zA-Z0-9-]+))$'",
+				"field owner.github: Does not match pattern '^(([a-zA-Z0-9-_]+)|([a-zA-Z0-9-_]+\\/[a-zA-Z0-9-_]+))$'",
 			},
 		},
 		"bad_owner_type": {
--- a/code/go/pkg/validator/validator_test.go
+++ b/code/go/pkg/validator/validator_test.go
@@ -78,7 +78,7 @@ func TestValidateFile(t *testing.T) {
 				"field policy_templates.2.data_streams.1: data stream doesn't exist",
 			},
 		},
-		"bad_github_owner/": {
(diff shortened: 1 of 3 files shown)


jsoriano (commented):
Could you please add a changelog entry in `spec/changelog.yml`, and an example of usage in a test package? Thanks!

jsoriano:
Oh, and looking to the failures in CI you will need to update the pattern that appears in error messages in tests.

code/go/pkg/validator/validator_test.go
```diff
@@ -78,10 +78,10 @@ func TestValidateFile(t *testing.T) {
 				"field policy_templates.2.data_streams.1: data stream doesn't exist",
 			},
 		},
-		"bad_github_owner": {
+		"bad_github_owner/": {
```
jsoriano, line 81: @shahzad31 is this change required?

spec/changelog.yml
```diff
@@ -636,3 +636,6 @@
   - description: Introduce "value" property for constant_keyword fields
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/194
+  - description: Allow _ in github owner field
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/668
```
jsoriano, line 641: This change should be moved to the section for the next version, top of the file.

spec/integration/manifest.spec.yml
```diff
@@ -251,7 +251,11 @@ spec:
         github:
           description: Github team name of the package maintainer.
           type: string
-          pattern: '^(([a-zA-Z0-9-]+)|([a-zA-Z0-9-]+\/[a-zA-Z0-9-]+))$'
+          pattern: '^(([a-zA-Z0-9-_]+)|([a-zA-Z0-9-_]+\/[a-zA-Z0-9-_]+))$'
+          examples:
+                - "elastic"
+                - "apm-agent-java"
+                - "ux_infra_team"
```
jsoriano, line 258: :+1: thanks for adding examples too.

jsoriano (changes requested):

code/go/pkg/validator/validator_test.go
shahzad31: bad_github_owner wasn't bad anymore :) 

/ makes it bad now :)

jsoriano: This is the name of the directory with the package, it should not make any difference to have the slash or not.
I would prefer not having it for consistency with the rest of packages listed here.

spec/changelog.yml
jsoriano: Fixed.

jsoriano (approved):

PR #679, opened 2023-12-08
Update go runtime to 1.21.5

jsoriano:
## What does this PR do?

Update go runtime to 1.21.5.

## Why is it important?

Keep the runtime updated. Not so important as this Go version only determines the version used in CI.

## Checklist

## Related issues

-


jsoriano:
Closing, will include this change in https://github.com/elastic/package-spec/pull/680

Issue #630, opened 2023-10-02
Empty arrays fail validation as null


jsoriano:
Reported in https://github.com/elastic/package-spec/pull/616#issuecomment-1739333137

When trying to define a setting with an empty array, it is validated as `null`, what can fail validation.

This may be related to the conversions from yaml to json that we do to use jsonschema and so on.

Once this is solved, check if elastic-package is able to override array fields with empty values when importing them from ECS (see https://github.com/elastic/elastic-package/blob/9d8ce35eb3051b832057b37e75ad454e28069ee6/internal/fields/dependency_manager.go#L349)

PR #683, opened 2023-12-12
Fix conversion to JSON of empty YAML arrays

jsoriano:
## What does this PR do?

Initialize arrays when converting them from YAML to JSON.

## Why is it important?

We were not initializing arrays when converting from YAML to JSON, so the ones that don't have elements were kept as null values. This lead to unexpected error messages about unexpected null values.

## Checklist

## Related issues

- Fixes https://github.com/elastic/package-spec/issues/630

### DIFF
--- a/code/go/pkg/validator/validator_test.go
+++ b/code/go/pkg/validator/validator_test.go
@@ -172,7 +172,6 @@ func TestValidateFile(t *testing.T) {
 			"data_stream/foo_stream/manifest.yml",
 			[]string{
 				"field streams.0.vars.1: options is required",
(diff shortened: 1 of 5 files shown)


code/go/pkg/validator/validator_test.go
```diff
@@ -172,7 +172,6 @@ func TestValidateFile(t *testing.T) {
 			"data_stream/foo_stream/manifest.yml",
 			[]string{
 				"field streams.0.vars.1: options is required",
-				"field streams.0.vars.2.options: Invalid type. Expected: array, given: null",
```
jsoriano, line 175: This was indeed wrong.

jsoriano:
test integrations

Issue #685, opened 2023-12-13
[Change Proposal] Take into account visualization filters for filter requirements on dashboards


jsoriano:
We had [some feedback](https://github.com/elastic/integrations/pull/6954#issuecomment-1828275810) on validations recently added, specifically:

> And what filters? Each visualization is filtered for the correct dataset. Is it wanting a filter on the dashboard itself? If so, that is a terrible thing to enforce. My users temporarily pin and remove filters all day long; often using the "remove all" option. If you put the filter on the dashboard itself, there is nothing preventing them from removing that "saved" filter and populating the dashboards with everything in logs-*.

It can be indeed a good idea to accept dashboards that don't have a filter, but where all its visualizations have one. In these cases we would also avoid searching in all indexes.

Regarding pinning, adding and removing filters, this is something an expected use case, but we have to take into account that "managed" dashboards cannot be modified since 8.11.

PR #687, opened 2023-12-20
Accept dashboards without filter that have filters in all panels

jsoriano:
## What does this PR do?

Checks if dashboards have filters in all panels, if they do, they don't require another filter at the dashboard level. 

## Why is it important?

Defining the filters in the panels can be more convenient in many use cases, so users are more flexible on the filters they can use, while still limiting searches.

## Checklist

## Related issues

- Fix https://github.com/elastic/package-spec/issues/685


code/go/internal/validator/semantic/validate_kibana_filter_present.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 var (
+	errDashboardPanelWithoutFilter   = errors.New("at least one panel does not have a filter")
 	errDashboardWithFilterAndNoQuery = errors.New("saved query found, but no filter")
```
mrodm, line 22: Not related to this PR, but this error should not be renamed as `errDashboardWithQueryAndNoFilter` ?

mrodm (dismissed):
LGTM

code/go/internal/validator/semantic/validate_kibana_filter_present.go
```diff
@@ -52,13 +53,74 @@ func ValidateKibanaFilterPresent(fsys fspath.FS) specerrors.ValidationErrors {
...
+		err = json.Unmarshal([]byte(panelsJSON), &panels)
+		if err != nil {
+			return fmt.Errorf("unable to decode embedded panels definition: %w", err)
+		}
+	default:
+		return fmt.Errorf("unexpected type for panels JSON: %T", panelsJSON)
+	}
+
+	for _, panel := range panels {
+		if panel.Type == "search" {
```
jsoriano, line 103: @drewdaemon are there other types of panels for which it would be ok if they don't have a filter?

(this is the panel why I added this https://github.com/elastic/package-spec/pull/687/files#diff-0e0b0481bb49a2f2675e5a87302801e3489b686114aef00040cae507f211a011R604)

mrodm (dismissed):
:+1:

code/go/internal/validator/semantic/validate_kibana_filter_present.go
drewdaemon: Unfortunately, it's a bit more nuanced than panel type. For example, most panels of type `visualization` make ES requests and would benefit from adding filters. But, markdown panels, which don't make any requests and don't support filters, are also of type `visualization`.

We had to deal with similar nuances/edge cases to reliably [detect legacy visualizations](https://github.com/elastic/kbncontent/blob/main/kbncontent.go#L77-L97).

I'm thinking that `kbncontent` might be the place to put this logic since there's already a groundwork there. Maybe we could add a `CanUseFilter()` method to the [visualization descriptor struct](https://github.com/elastic/kbncontent/blob/main/kbncontent.go#L19-L25) and use it here?

jsoriano: Good point, opened PR in https://github.com/elastic/kbncontent/pull/4, lets discuss there what visualizations we should check.

jsoriano:
Converting to draft till we decide about https://github.com/elastic/package-spec/pull/687#discussion_r1432671169.

drewdaemon (approved):
LGTM!

code/go/internal/validator/semantic/validate_kibana_filter_present_test.go
```diff
@@ -64,6 +64,9 @@ func TestCheckDashboardHasFilter(t *testing.T) {
 			require.Len(t, files, 1, "looking for %s", c.dashboard)
 
 			err = checkDashboardHasFilter(files[0])
+			if err != nil {
+				t.Log(err)
+			}
```
mrodm, line 69: Is this a leftover or is it intended to help checking test output ?

mrodm (approved):
:+1:

code/go/internal/validator/semantic/validate_kibana_filter_present_test.go
jsoriano: I found it useful when running the test with verbosity.

PR #689, opened 2024-01-04
Adds top level `volumes` property and `volume` definition

Tacklebox:
## What does this PR do?

This adds the properties and definitions required to support validating docker compose 2.3 volumes in the custom-agent.yml of an elastic integration.

## Why is it important?

The cloud_defend integration requires mounting a volume for its system test. The current schema validation fails due to not having definitions for volumes in the custom-agent.yaml file.

## Checklist

## Related issues

- Relates https://github.com/elastic/integrations/pull/8567

### DIFF
--- a/spec/integration/_dev/deploy/agent/custom-agent.spec.yml
+++ b/spec/integration/_dev/deploy/agent/custom-agent.spec.yml
@@ -53,18 +53,18 @@ spec:
         "^[a-zA-Z0-9._-]+$":
           "$ref": "#/definitions/volume"
       additionalProperties: false
-  services:
-      description: Service list
-      type: object
-      additionalProperties:
-        description: Additional services
-        "$ref": "#/definitions/service"
-      properties:
-        docker-custom-agent:
-          description: Custom agent service definition
+    services:
+        description: Service list
+        type: object
+        additionalProperties:
+          description: Additional services
           "$ref": "#/definitions/service"
-      required:
-        - docker-custom-agent
+        properties:
+          docker-custom-agent:
+            description: Custom agent service definition
+            "$ref": "#/definitions/service"
+        required:
+          - docker-custom-agent
   required:
     - version
     - services
--- a/spec/integration/_dev/deploy/agent/custom-agent.spec.yml
+++ b/spec/integration/_dev/deploy/agent/custom-agent.spec.yml
@@ -14,33 +14,7 @@ spec:
           - required:
             - hostname
     volume:
-      type:
-        - object
-        - "null"
-      properties:
-        name:
-          type: string
-        driver:
-          type: string
-        driver_opts:
-          type: object
-          patternProperties:
-            "^.+$": 
-              type:
-                - string
-                - number
-        external:
-          type:
-            - boolean
-            - object
-          properties:
-            name:
-              deprecated: true
-              type: string
-          additionalProperties: false
-          patternProperties: { "^x-": {} }
-      additionalProperties: false
-      patternProperties: { "^x-": {} }
+      type: "null"
   properties:
     version:
       description: Docker Compose version
--- a/spec/integration/_dev/deploy/agent/custom-agent.spec.yml
+++ b/spec/integration/_dev/deploy/agent/custom-agent.spec.yml
@@ -54,17 +54,17 @@ spec:
           "$ref": "#/definitions/volume"
       additionalProperties: false
     services:
-        description: Service list
-        type: object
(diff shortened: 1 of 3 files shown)


spec/integration/_dev/deploy/agent/custom-agent.spec.yml
```diff
@@ -13,12 +13,47 @@ spec:
...
+                - number
+        external:
+          type:
+            - boolean
+            - object
+          properties:
+            name:
+              deprecated: true
+              type: string
+          additionalProperties: false,
```
jsoriano, line 40: CI is failing due to something related to additionalProperties, it could be the comma here.
```suggestion
          additionalProperties: false
```

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
```diff
@@ -13,12 +13,47 @@ spec:
         anyOf:
           - required:
             - hostname
+    volume:
```
jsoriano, line 16: Do we want to allow all kind of volumes?

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
```diff
@@ -13,12 +13,47 @@ spec:
...
       pattern: '^2.3$'
-    services:
+    volumes:
+      description: Docker Compose volume declaration
+      type: object
+      patternProperties:
+        "^[a-zA-Z0-9._-]+$":
+          "$ref": "#/definitions/volume"
+      additionalProperties: false
+  services:
```
mrodm, line 56: This `services` key previously was at the same level as `version` property, but now it looks like there were removed two whitespaces.
Should those whitespaces be added ?
```suggestion
    services:
```

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
Tacklebox: I am only using a basic named volume in my PR. But I thought it would be easier and faster to just copy the complete schema from docker compose for volumes. Though I'm a little unclear on the motivation behind linting this file in the first place 🤷

jsoriano: The reason to validate this file would be to control the features that can be used, for example we don't want to allow to override the hostname, and it could be also problematic to support any network.

I would suggest to remove the properties that we don't plan to use by now.

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
```diff
@@ -13,23 +13,58 @@ spec:
...
-        "$ref": "#/definitions/service"
-      properties:
-        docker-custom-agent:
-          description: Custom agent service definition
+      patternProperties:
+        "^[a-zA-Z0-9._-]+$":
+          "$ref": "#/definitions/volume"
+      additionalProperties: false
+    services:
+        description: Service list
```
jsoriano, line 57: Unexpected indentation changes here.

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
```diff
@@ -13,23 +13,58 @@ spec:
...
+      additionalProperties: false
+      patternProperties: { "^x-": {} }
   properties:
     version:
       description: Docker Compose version
       type: string
       pattern: '^2.3$'
-    services:
-      description: Service list
+    volumes:
```
jsoriano, line 49: Could you add an example of a volume in a test package? For example in `./test/packages/deploy_custom_agent/_dev/deploy/agent/custom-agent.yml`.

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
```diff
@@ -13,11 +13,20 @@ spec:
         anyOf:
           - required:
             - hostname
+    volume:
+      type: "null"
```
jsoriano, line 17: Does it mean that we only allow empty definitions now? SGTM if this is the only thing we need now. Thanks for removing the definitions that we don't need by now.

jsoriano (approved):

spec/integration/_dev/deploy/agent/custom-agent.spec.yml
Tacklebox: Yep, that's all I needed to get my integration test to work. Just to have an empty directory mounted into the agent container. 👍

Issue #583, opened 2023-09-01
Add new validation rule to avoid rename processor without checking event.original


mrodm:
Originally posted in https://github.com/elastic/integrations/issues/3451#issuecomment-1693521101

Add a new validation rule to ensure that if there is a rename processor whose target_field is `event.original` and `field` is `message`, it needs also to check that `if`  key is defined checking about the presence of `event.original` in the document.

Moreover this new validation rule needs to check that a `remove` processor is defined to remove the `message` field, to avoid duplication.

As an example, if there is that rename processor in the ingest pipeline, a remove processor should exist and that rename processor should be like :
```yaml
    - rename:
        field: message
        target_field: event.original
        # ignore_missing: true 
        if: 'ctx.event?.original == null'
    - remove:
        field: message
        ignore_missing: true
        if: 'ctx.event?.original != null'
```

Similar to what it was done in https://github.com/elastic/integrations/pull/7026

ishleenk17:
@mrodm : The remove processor validation should not be limited to just the message field. 
So I am not sure if we should add this to the validation here. 
There should  be a generic validation that any field if renamed and not being used again should be removed.

jillguyonnet:
@ishleenk17 Apologies for not addressing your comment earlier. Renaming `message` to `event.original` is a common pattern (over 130 packages were affected in the [test branch](https://github.com/elastic/integrations/pull/8887)); it was straightforward enough to address through conditional schema, which also allowed stronger restrictions (such as the presence and value of the `if` property). Implementing a more general rule would probably open wider discussions and require a more involved approach. Please let us know if you have any concerns or feel free to open another issue to discuss this further.

Issue #676, opened 2023-12-04
Add minimal test to test validation of fields on specific spec versions


jsoriano:
We have many cases of good and bad fields, that in addition may depend on the package spec versions.
To test them we create whole test packages, see for example the test packages in https://github.com/elastic/package-spec/pull/675.

We could add a smaller test that only requires a spec version and a field definition, in `yaml`, and we would check the expected errors if any.

This test could work with `yamlspec`, but loading a reduced spec, that defines only a single file, with the definition of a single field, that can have subfields. The test would generate this file using the provided yaml, and would run validation on it. **Update:** In the implementation in https://github.com/elastic/package-spec/pull/700 we are validating whole packages, not only fields with a partial yamlspec, to validate also with semantic rules.

We could then add test cases to this test by adding only yaml definitions and its expected result.

And we could replace with this several test packages that we currently have.

PR #690, opened 2024-01-11
[Fleet] Add rename message processor validation

jillguyonnet:
## What does this PR do?

This PR adds the following validation rule for ingest pipeline processors.

If an ingest pipeline contains a `rename` processor with `field: message` and `target_field: event.original`, then:
- the `rename` processor should also contain `if: 'ctx.event?.original == null'`
- there should also exist a `remove` processor with `field: message`

### Examples

Valid pipeline definition:
```yaml
processors:
    - rename:
        field: message
        target_field: event.original
        if: 'ctx.event?.original == null'
    - remove:
        field: message
```

Invalid pipeline definition:
```yaml
processors:
    - rename:
        field: message
        target_field: event.original
        # missing `if: 'ctx.event?.original == null'`
    - remove:
        field: message
```

Invalid pipeline definition:
```yaml
processors:
    - rename:
        field: message
        target_field: event.original
        if: 'ctx.event?.original == null'
    # missing remove processor
```

### Skipping validation

The validation error has the `JSE00001` error code that allows packages to skip it (cf. `skip_pipeline_rename_validation` test package).

## Why is it important?

There is an existing issue for integrations using a `rename` processor when the target field already exists: https://github.com/elastic/integrations/issues/3451#issuecomment-1693521101. This validation aims to check that no field called `event.original` exists if there is a rename processor that renames `message` to `event.original` (what was done in https://github.com/elastic/integrations/pull/7026).

## Additional changes

- Enhanced error message transforming with regex
- Custom error code based on error message regex

## Checklist

## Related issues

- Closes https://github.com/elastic/package-spec/issues/583


jillguyonnet:
@jsoriano @mrodm This is a first pass at the validation rule proposed in https://github.com/elastic/package-spec/issues/583. Can you please comment on the following:
1. Assuming this validation is enough for now:
   1. Are there any missing rules? (`ignore_missing`, `if` in `remove`)
   2. Can we test it against existing integrations?
   3. Would it be considered a breaking change?
   4. Should it follow a version constraint?
2. Obviously the hard-coded field names are restrictive. I'm not sure, however, how https://github.com/elastic/package-spec/issues/583#issuecomment-1729188866 could be addressed properly: can we assume that a renamed field should always be removed?
3. Are there other avenues we want to explore? Jaime had mentioned the following options:
   1. Static pipeline analysis
   2. Black box testing with `elastic-package test pipelines`

mrodm:
Hi @jillguyonnet ! I'll try to answer some of these questions

> 1. Are there any missing rules? (`ignore_missing`, `if` in `remove`)

I think it would also be needed to check that the `remove` processor has defined the `ignore_missing` and `if` condition (reverse condition as in rename). That would avoid that the pipelines fails for instance if `message` does not exist when removing it.


> 2\. Can we test it against existing integrations?

It can be tested directly in this Pull Request, you could add the comment `test integrations` 
 each time you want to test with all the packages in integrations repo, and it will be created a new PR in the integrations repository updating package-spec to the latest changeset of this PR / branch. Here there are some docs about this feature https://github.com/elastic/package-spec/blob/main/CONTRIBUTING.md#testing-with-integrations-repository


> * Would it be considered a breaking change?
> * Should it follow a version constraint?

From my POV, probably it would be a breaking change, since there could be some packages that would start failing because of this new validation (to be checked when tested with integrations repository).

I would add a version constraint to these changes, so this new validation just take effect from a specific spec version. If it is achieved using the JSON schema itself, it would be needed a JSON patch to remove before the following version.


> 2\. Obviously the hard-coded field names are restrictive. I'm not sure, however, how [Add new validation rule to avoid rename processor without checking event.original #583 (comment)](https://github.com/elastic/package-spec/issues/583#issuecomment-1729188866) could be addressed properly: can we assume that a renamed field should always be removed? (Edit: perhaps it could be a warning)

If it is needed to consider more fields apart from `message`, would it be better to move this check to be a semantic validation rule? Would it be easier to implement this check ? There is also the possibility to just run this validation from a specific version adding the `since` parameter.
https://github.com/elastic/package-spec/blob/5943cc23c2e168f534ed761a7779d53dde93104f/code/go/internal/validator/spec.go#L151-L172

WDYT @jsoriano ? Should it be needed to consider any renamed field in this validation rule ?

> Are there other avenues we want to explore? Jaime had mentioned the following options:
> 1. Static pipeline analysis
> 2. Black box testing with `elastic-package test pipelines`

These are good questions, because I don't know if there could be some integration removing or renaming some fields using painless scripts ([script processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/script-processor.html)). Or if there is more than condition in the `if` field.
Currently, I don't have any other suggestion for this apart from the ones mentioned by Jaime.

jsoriano:
Nice, I didn't consider this could be done in the schema directly. I think this is fine this way, unless we need to handle lots of corner cases and the schema starts getting convoluted.

I think that next step would indeed be to test with current integrations, to see if this new rule actually detects valid cases. We should also take a look and try to find other cases that we should catch. Once we have results on this we can decide how to follow.

As Mario mentioned, this rule should only apply on packages using the next version of the schema, we can apply a patch in the schema for this, but lets confirm first how this works with existing packages.
I think it is ok to make it a "breaking change" because it fixes things, we have made this kind of changes in the past. 

Regarding black box testing, we would need to investigate if we can programatically forge events that can be used to validate this rule in a general way. It'd be nice to have these kind of checks, but not sure if possible. In this case it would imply to run the tests with events containing `event.original`.

jsoriano:
test integrations

jillguyonnet:
test integrations

jillguyonnet:
test integrations

jillguyonnet:
Hey @jsoriano - apologies, I pushed a commit after your comment yesterday (there was a needed fix).

There are many integrations that fail this validation as can be seen in https://github.com/elastic/integrations/pull/8887. The ones I've looked at seem expected: for instance, validation of `1password` [fails](https://buildkite.com/elastic/integrations/builds/8099#018d11c9-2225-4194-ac33-6ac0bea4980d) which makes sense because [this pipeline](https://github.com/elastic/integrations/blob/main/packages/1password/data_stream/audit_events/elasticsearch/ingest_pipeline/default.yml) doesn't meet the criteria (no matching `remove` processor). So there will be quite a few adjustments if we go ahead.

A couple of questions:
1. I had added a version constraint (removed in the last commit for testing purposes), but the `-next` version yields some validation failures, which seems like a catch-22 as I'd need at least the `good_v3` and `bad_ingest_pipeline` test packages to use this version to test the changes. How do I address this?
2. I wanted to put more validation tests as there are many ways to break the validation, but I struggled to test with more than one data stream because of the [test cases map format](https://github.com/elastic/package-spec/blob/main/code/go/pkg/validator/validator_test.go#L28) (which seems to take one package name and one file name, if I'm not mistaken). I added a `bad_ingest_pipeline_v2` test package as a workaround, but would there be a way to simply add more data streams to `bad_ingest_pipeline` and test those?

jillguyonnet:
FYI https://www.jsonschemavalidator.net/ is as always super useful to play with the schema definition, posting the test JSON here for reference.

<details>
<summary>Schema</summary>

```json
{
  "type": "object",
  "properties": {
    "processors": {
      "type": "array",
      "items": {
        "type": "object",
        "additionalProperties": false,
        "properties": {
          "rename": {
            "type": "object"
          },
          "remove": {
            "type": "object"
          }
        }
      },
      "if": {
        "contains": {
          "required": [
            "rename"
          ],
          "properties": {
            "rename": {
              "properties": {
                "field": {
                  "const": "message"
                },
                "target_field": {
                  "const": "event.original"
                }
              }
            }
          }
        }
      },
      "then": {
        "allOf": [
          {
            "contains": {
              "required": [
                "rename"
              ],
              "properties": {
                "rename": {
                  "required": [
                    "if"
                  ],
                  "properties": {
                    "if": {
                      "const": "ctx.event?.original == null"
                    }
                  }
                }
              }
            }
          },
          {
            "contains": {
              "required": [
                "remove"
              ],
              "properties": {
                "remove": {
                  "required": ["field", "ignore_missing", "if"],
                  "properties": {
                    "field": {
                      "const": "message"
                    },
                    "ignore_missing": {
                      "const": true
                    },
                    "if": {
                      "const": "ctx.event?.original != null"
                    }
                  }
                }
              }
            }
          }
        ]
      }
    }
  }
}
```
</details>

<details>
<summary>Test definition</summary>

```json
{
  "processors": [
    {
      "rename": {
        "field": "message",
        "target_field": "event.original",
        "if": "ctx.event?.original == null"
      }
    },
    {
      "remove": {
        "field": "message",
        "ignore_missing": true,
        "if": "ctx.event?.original != null"
      }
    }
  ]
}
```
</details>

jsoriano:
> There are many integrations that fail this validation as can be seen in elastic/integrations#8887.

There are indeed many of them. I wonder if any of them are doing the removal in a different way. With so many packages failing because of this we may need to think twice about the strategy to address this.

> So there will be quite a few adjustments if we go ahead.

This can be fine, we will add the new validation in a new version of the spec so packages will only need to be updated when updating to this version. We can also add an error code to the error message, so developers can skip this check if they don't want to fix it at the moment.

> 1\. I had added a version constraint (removed in the last commit for testing purposes), but the `-next` version yields some validation failures, which seems like a catch-22 as I'd need at least the `good_v3` and `bad_ingest_pipeline` test packages to use this version to test the changes. How do I address this?

I would suggest to continue without the version constraint while we confirm the implementation. For testing yes, we can use the next version in good_v3 and bad_ingest_pipeline.

> 2\. I wanted to put more validation tests as there are many ways to break the validation, but I struggled to test with more than one data stream because of the [test cases map format](https://github.com/elastic/package-spec/blob/main/code/go/pkg/validator/validator_test.go#L28) (which seems to take one package name and one file name, if I'm not mistaken). I added a `bad_ingest_pipeline_v2` test package as a workaround, but would there be a way to simply add more data streams to `bad_ingest_pipeline` and test those?

Yes, as this is now you would need to add one package per test case, this is similar to the issue that I plan to address in https://github.com/elastic/package-spec/issues/676.

For this case maybe you can maybe create a new test with multiple test cases, each case would reference a pipeline, and then before the execution you create a package by copying a base package to a tmp dir, and then copy the specific pipeline to this package.

jillguyonnet:
Hey @jsoriano 

> There are indeed many of them. I wonder if any of them are doing the removal in a different way. With so many packages failing because of this we may need to think twice about the strategy to address this.

134 packages are failing this validation. By far the most common scenario seems to be that the pipeline has
```yaml
  - rename:
      field: message
      target_field: event.original
      ignore_missing: true
      if: ctx.event?.original == null
```
but no processor to remove the `message` field. Instead, some have this:
```yaml
  - remove:
      field: event.original
      if: "ctx?.tags == null || !(ctx.tags.contains('preserve_original_event'))"
      ignore_failure: true
      ignore_missing: true
```

I think adding an error code in order to make this validation skippable would be a very good approach. I was playing with it based on [this example](https://github.com/elastic/package-spec/pull/654/files#diff-01c0062de741049c7dede60409d6227974f7b2f5b1b8f80fa5c40e725400be6aR120), but I'm not sure whether it could catch unwanted validation since it relies on regex. For example, the following 4 error messages could happen:
```
"field processors.1.rename: if is required"
"field processors.0: remove is required"
"field processors.2.remove.field: processors.2.remove.field does not match: \"message\""
"field processors.2.remove.if: processors.2.remove.if does not match: \"ctx.event?.original != null\""
```
WDYT?

> Yes, as this is now you would need to add one package per test case, this is similar to the issue that I plan to address in https://github.com/elastic/package-spec/issues/676.

> For this case maybe you can maybe create a new test with multiple test cases, each case would reference a pipeline, and then before the execution you create a package by copying a base package to a tmp dir, and then copy the specific pipeline to this package.

I've extracted the ingest pipeline validation tests in the latest commit. The only issue really was that the error message was wrong because in `TestValidateFile` there is only one file associated. The new test allows multiple files, so multiple pipelines per package. WDYT?

code/go/pkg/validator/validator_test.go
```diff
@@ -722,7 +715,72 @@ func TestValidateRoutingRules(t *testing.T) {
...
+				"field processors.2.foreach.processor: Additional property paint is not allowed",
+			},
+			"bad_rename_message": []string{
+				"field processors.1.rename: if is required (JSE00001)",
+				"field processors.0: remove is required (JSE00001)",
+			},
+			"bad_rename_message_2": []string{
+				"field processors.2.remove.field: processors.2.remove.field does not match: \"message\" (JSE00001)",
+				"field processors.2.remove.if: processors.2.remove.if does not match: \"ctx.event?.original != null\" (JSE00001)",
+			},
```
jsoriano, line 738: Would it make sense to replace these error messages, so they are more significant for this use case?

Something like this:
```suggestion
			"bad_rename_message": []string{
				"field processors.1.rename: at least one is required to move \"message\" to \"event.original\" (JSE00001)",
				"field processors.0: remove is required to remove \"message\" when \"event.original\" exists (JSE00001)",
			},
			"bad_rename_message_2": []string{
				"field processors.2.remove.field: processors.2.remove.field at least one remove processor should remove \"message\" (JSE00001)",
				"field processors.2.remove.if: processors.2.remove.if does not match: \"ctx.event?.original != null\" (JSE00001)",
			},
```

My only concern would be that maybe we replace the message for other cases, but I don't think we are enforcing these conditions in other cases.

code/go/pkg/validator/validator_test.go
```diff
@@ -722,7 +715,72 @@ func TestValidateRoutingRules(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestValidateIngestPipelines(t *testing.T) {
```
jsoriano, line 720: :+1: with this test we can test multiple cases

jsoriano (commented):
Looks good :+1: I think we can go on with this. @mrodm could you also take a look?

code/go/pkg/validator/validator_test.go
jillguyonnet: Good point 👍 This should be fairly safe as long as we don't add other similar conditions. I can add a comment in the code that these renamed error messages should be checked in case this changes.

jillguyonnet: I pushed a commit with renamed error messages, let me know what you think. I made the rules as strict as possible to avoid collisions with other cases. For example, just using `if is required` would also have affected [this test](https://github.com/elastic/package-spec/blob/b8224e23383208ce3688c1e63b1a4795b1037f3f/code/go/pkg/validator/validator_test.go#L699).

jillguyonnet:
test integrations

code/go/pkg/validator/validator_test.go
jillguyonnet: Actually, looking at the integrations tests, I realised that we can't hardcode the processor number so I would need to change that to a regex.

jsoriano: Oh I see. What do you think about adding support for regexps in the transforms?

jillguyonnet:
test integrations

code/go/pkg/validator/validator_test.go
jillguyonnet: Pushed a commit with that, let's see how it impact the integrations.

jillguyonnet:
test integrations

code/go/internal/validator/spec.go
```diff
@@ -102,36 +103,73 @@ func substringInSlice(str string, list []string) bool {
...
+			matcher: regexp.MustCompile(`Must not validate the schema \(not\)`),
+			new:     "Must not be present",
+		},
+		{
+			matcher: regexp.MustCompile("secret is required"),
+			new:     "variable identified as possible secret, secret parameter required to be set to true or false",
+		},
+		{
+			matcher: regexp.MustCompile(`field processors.[0-9]+.rename: if is required`),
+			new:     "rename \"message\" to \"event.original\" processor requires if: 'ctx.event?.original == null'",
```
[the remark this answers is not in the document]
jillguyonnet: I pushed a somewhat simpler version - after some testing, I _think_ it's sufficient for the current use case (although I also gave it the flexibility to not use the extracted variable). Let me know if I've missed anything.

spec/changelog.yml
```diff
@@ -2,6 +2,12 @@
 ## This file documents changes in the package specification. It is NOT a package specification file.
 ## Newer entries go at the bottom of each in-development version.
 ##
+- version: 3.2.0-next
+  changes:
+  - description: Add validation for ingest pipeline rename processor
+    type: breaking-change
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/690
```
mrodm, line 10: As `3.1.0` has not been released yet, probably it could be moved to the `3.1.0-next` changelog entry, couldn't it? Not sure if I could be missing something here.

code/go/internal/validator/spec.go
```diff
@@ -102,36 +103,77 @@ func substringInSlice(str string, list []string) bool {
...
+			} else if msg.matcher.MatchString(e.Error()) {
 				e = specerrors.NewStructuredError(
-					errors.New(strings.Replace(e.Error(), msg.original, msg.new, 1)),
+					errors.New(strings.Replace(e.Error(), msg.matcher.FindString(e.Error()), msg.new, 1)),
 					specerrors.UnassignedCode)
 			}
 		}
+		for _, transform := range addErrorCode {
+			if transform.matcher.MatchString(e.Error()) {
+				e = specerrors.NewStructuredError(e, transform.code)
```
mrodm, line 174: Should we create here an new error to avoid wrapping a StructuredError inside another structuredError ? 

StructuredError implements `UnWrap`, so not sure if this is a problem or not @jsoriano WDYT ?


If that could be a problem, this could be changed by something like:

```suggestion
				e = specerrors.NewStructuredError(error.New(e.Error()), transform.code)
```

mrodm (commented):
This PR looks really good! Added a few questions

code/go/internal/validator/spec.go
jsoriano: TLDR; Any option looks good to me for this PR, but let's keep the discussion for a follow up.

I see that in other transforms we are creating new errors. For consistency we should do the same thing in all cases.

Regarding wrapping structured errors inside structured errors, I don't see why this should be a problem, maybe we have to clarify the semantics in the `specerrors` package.

It would be probably better to wrap the error as is instead of creating new ones. By creating new ones we are losing information. For example an structured error could be wrapping an `os.ErrNotFound` error, and we would be losing this information with `error.New(e.Error())`.

So any option looks good to me for this PR, but we should follow up clarifying these cases. If wrapping structured errors is a problem, `NewStructuredError` could take care of this, perhaps with something like this:
```
// NewStructuredError creates a generic validation error by wrapping the given error. If the given error
// is already a StructuredError, the returned error wraps the error wrapped by the given error.
func NewStructuredError(err error, code string) *StructuredError {
        if se, ok := err.(*StructuredError) {
                result := *se
                result.code = code
                return &result
        }
        return &StructuredError{
                err:  err,
                code: code,
        }
}
```

spec/changelog.yml
jsoriano: Since v3 we are not strictly following semver for the package spec, I tried to explain this in https://github.com/elastic/package-spec#changing-a-specification. The summary would be that we only must use new minors for changes that require changes in the stack.

The reason for this change in v3 is that we are using this version for the compatibility of packages with the stack (since https://github.com/elastic/ingest-dev/issues/2185, internal only sorry), so we have to see if it is breaking specially regarding the stack, and not the packages themselves.

> The difference is that this change is breaking, while the one under `3.1.0-next` isn't. The PR for it has been merged, so maybe we could release 3.1.0?

The change about `subobjects` is in 3.1.0 because it needs support in the stack, even if it is indeed not breaking for packages. Before releasing 3.1.0 we need to be sure that this support has been implemented, and we will also need to update the max version supported by Kibana offerings ([these settings](https://github.com/elastic/kibana/blob/15f999fbdb2322f9ac6c2318c753bc502ee1d2ed/config/serverless.oblt.yml#L44)).

----

This change adds a new validation to fix issues in packages, and doesn't require changes in the stack, so it can go in the next patch version, `3.0.5-next` in this case. If we wanted to release it on next minor, it should be in `3.1.0-next` that hasn't been released yet.

spec/integration/elasticsearch/pipeline.spec.yml
```diff
@@ -75,3 +107,11 @@ spec:
     on_failure:
       description: List of processors to apply in case of failure.
       $ref: "#/definitions/processors"
+
+  versions:
+  - before: 3.2.0-next
```
jsoriano, line 112:
```suggestion
  - before: 3.0.5-next
```

jsoriano (commented):
LGTM, but added comment about the version to use, I would include it in `3.0.5`. It could be included in 3.1.0 too, but I wouldn't start a new minor.

jillguyonnet:
@jsoriano Thank you for the context and details around version change. I've changed it to 3.0.5 as requested.

Regarding the discussion around error wrapping, should we open a followup issue? Depending on the decision, it might incur other changes in the validator.

jsoriano (approved):
Nice! :+1:

jillguyonnet:
Awesome, thanks a lot @jsoriano and @mrodm for your help and comments!

spec/changelog.yml
```diff
@@ -2,22 +2,17 @@
...
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/573
 - version: 3.0.5-next
   changes:
-  - description: Prepare for next version
+  - description: Add validation for ingest pipeline rename processor
+    type: breaking-change
     type: enhancement
-    link: https://github.com/elastic/package-spec/pull/697
+    link: https://github.com/elastic/package-spec/pull/690
```
mrodm, line 15: This changelog entry should be moved to the bottom of this version

mrodm (approved):
:+1: Thanks @jillguyonnet !

Just a minor comment to move the changelog entry

jillguyonnet:
Thank you @mrodm, done! Looks like the new commit dismissed your review though (also @jsoriano).

mrodm (approved):
:+1:

PR #700, opened 2024-01-24
Refactor tests for field definitions

### LINKED ISSUE #676: Add minimal test to test validation of fields on specific spec versions

jsoriano:
## What does this PR do?

Adds a new test for field definitions, whose cases receive as inputs a template package, a set of fields definitions and a list of expected errors. The test builds for each case a new package from the template package and inserts the field definitions, then it runs the validations and checks for the expected errors. A test case looks like this now:
```go
                {
                        title:           "required object type",
                        packageTemplate: "integration_v3_0",
                        fields: []map[string]any{
                                {
                                        "name": "object_without_object_type.*",
                                        "type": "object",
                                },
                        },
                        expectedErrors: []string{
                                `field 0: object_type is required`,
                        },
                },
```

Test packages whose only purpose was to test field definitions have been removed and their test cases migrated to the new test.

I have decided to validate complete packages so we also validate against semantic rules.

## Why is it important?

This approach makes easier to add new test cases for fields. With previous implementation we need to add fields to a package and then add the list of expected errors to a test, with the new approach the field definitions to be tested and the expected errors are together in the same test case.
Base packages can be more easily reused, so we don't need to add new packages for each new feature we want to test.

## Checklist

- [ ] ~~I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).~~

## Related issues

- Closes #676


code/go/pkg/validator/validator_fields_test.go
```diff
@@ -0,0 +1,521 @@
...
+					"enabled":     false,
+				},
+			},
+			expectedErrors: []string{
+				`field 0.enabled: 0.enabled does not match: true`,
+			},
+		},
+		{
+			title:           "bad disabled object was allowed till 3.0.2",
+			packageTemplate: "integration_v3_0_2",
```
mrodm, line 166: Could it be added a `version` field to set the `format_version` in the manifest.yml when creating the temporal package for this unit test?

I see for instance that there are some tests using `integration_v3_0`  and another `integration_v3_0_2` templates. Looking into these templates, they look the same package just changing the spec version.

That would allow us to create a template just per major (o minor?), or just packages containing some specific settings like the ones for time series.
WDYT?

code/go/pkg/validator/validator_fields_test.go
```diff
@@ -0,0 +1,521 @@
...
+	return createPackageWithRawFields(t, pkgName, pkgTemplate, encodedFields)
+}
+
+func createPackageWithRawFields(t *testing.T, pkgName string, pkgTemplate string, rawFields []byte) string {
+	basePackageDir := filepath.Join("testdata", "templates", pkgTemplate)
+	require.DirExists(t, basePackageDir, "checking if template package exists")
+
+	pkgDir := filepath.Join(t.TempDir(), pkgName)
+	err := cp.Copy(basePackageDir, pkgDir)
+	require.NoError(t, err, "copying template package")
```
mrodm, line 512: One option to set spec version dynamically could be setting a specific string for the `format_version` field in the manifest file for each template. For instance:
```
format_version: <to_be_set>
```

This could be replaced by tests when creating the package with something like:

```suggestion
    basePackageDir := filepath.Join("testdata", "templates", pkgTemplate)
    require.DirExists(t, basePackageDir, "checking if template package exists")

    packageManifest := filepath.Join(basePackageDir, "manifest.yml")
    manifest, err := os.ReadFile(packageManifest)
    require.NoError(t, err)

    output := bytes.ReplaceAll(manifest, []byte("format_version: <to_be_set>"), []byte(fmt.Sprintf("format_version: %s", pkgVersion)))

    err = os.WriteFile(packageManifest, output, 0644)
    require.NoError(t, err)

    pkgDir := filepath.Join(t.TempDir(), pkgName)
    err = cp.Copy(basePackageDir, pkgDir)
    require.NoError(t, err, "copying template package")
```

code/go/pkg/validator/validator_fields_test.go
```diff
@@ -0,0 +1,521 @@
...
+
+	if len(rawFields) > 0 {
+		fieldsPath := filepath.Join(pkgDir, "data_stream", "test", "fields", "fields.yml")
+		err := os.WriteFile(fieldsPath, rawFields, 0644)
+		require.NoError(t, err, "creating fields file")
+	}
```
mrodm, line 518: This looks really nice! 
Following this approach, it probably could also be added specific unit tests for some other packages tested in TestValidateFile from `validator_test.go`. For instance those test packages that are testing fields on `manifest.yml` (missing_version, bad_github_owner, bad_owner_type), data stream manifests, ingest pipelines...

Like having a base template package, and the test could add any file within the package with any content. Once that package is created call `ValidateFromPath()` as it is done here.

Do you think it would be possible to do that ? I don't mean to do it in this PR

code/go/pkg/validator/validator_fields_test.go
jsoriano: I thought about this, but there are versions that require more changes apart from the format version in the manifest. It is true that this would help though in this case. I will give it a try. We will still need to keep the `packageTemplate`.

code/go/pkg/validator/validator_fields_test.go
jsoriano: I would leave a default version and then optionally allow to override it. I will give it a try.

code/go/pkg/validator/validator_fields_test.go
jsoriano: Agree, we could follow this approach in more cases. We can iterate later. Though many of these changes don't have so many cases and don't change so much as field, so maybe the investment doesn't have so many benefits.

jsoriano:
@mrodm added the option to override the spec version. It has been a bit rabbit hole because we have to modify also the changelog, and take into account that using non-GA versions of the spec also requires the package to have a non-GA version.

I also added an option to apply patches to files. It only supports appending, but is all we need by now.

mrodm:
> @mrodm added the option to override the spec version. It has been a bit rabbit hole because we have to modify also the changelog, and take into account that using non-GA versions of the spec also requires the package to have a non-GA version.
> 
> I also added an option to apply patches to files. It only supports appending, but is all we need by now.

Oh! I didn't forsee that it would require that many changes, thanks @jsoriano!!

mrodm (approved):
Great!! That would help us to not create that many packages for testing , thanks!!

Issue #698, opened 2024-01-23
[Change Proposal] Add support for counted_keyword field type


JonasKunz:
Recently the `counted_keyword` field type has been added to elasticsearch ((https://github.com/elastic/elasticsearch/pull/101826).

We would like to add a new field of this type to the APM package (see https://github.com/elastic/apm-server/issues/12054).

Unfortunately this is not yet possible, because the `elastic-package` utility complains about `counted_keyword` being an invalid field type:

```
jonas@Jonass-MBP apm % ../../elastic-package build
2024/01/23 13:48:29  WARN CommitHash is undefined, in both /Users/jonas/.elastic-package/version and the compiled binary, config may be out of date.
Build the package
README.md file rendered: /Users/jonas/git/integrations/packages/apm/docs/README.md
2024/01/23 13:48:30  INFO License text found in "/Users/jonas/git/integrations/LICENSE.txt" will be included in package
Error: building package failed: invalid content found in built zip package: found 6 validation errors:
   1. item [.DS_Store] is not allowed in folder [/Users/jonas/git/integrations/build/packages/apm-8.13.0-preview-1.zip/data_stream]
   2. item [.DS_Store] is not allowed in folder [/Users/jonas/git/integrations/build/packages/apm-8.13.0-preview-1.zip/data_stream/traces]
   3. item [.DS_Store] is not allowed in folder [/Users/jonas/git/integrations/build/packages/apm-8.13.0-preview-1.zip/data_stream/traces/elasticsearch]
   4. file "/Users/jonas/git/integrations/build/packages/apm-8.13.0-preview-1.zip/data_stream/traces/fields/fields.yml" is invalid: field 101.type: 101.type must be one of the following: "aggregate_metric_double", "alias", "histogram", "constant_keyword", "text", "match_only_text", "keyword", "long", "integer", "short", "byte", "double", "float", "half_float", "scaled_float", "date", "date_nanos", "boolean", "binary", "integer_range", "float_range", "long_range", "double_range", "date_range", "ip_range", "group", "geo_point", "object", "ip", "nested", "flattened", "wildcard", "version", "unsigned_long"
```

So to me it seems like we have to add the `counted_keyword` to the field type specification, maybe [this PR](https://github.com/elastic/package-spec/pull/500/files) can serve as a template.

```[tasklist]
### Tasks
- [x] Add field type to the package spec in a new minor version
- [ ] Add validation for subfields of this type in elastic-package
[...]

PR #707, opened 2024-01-31
Added counted_keyword type

JonasKunz:
## What does this PR do?

Adds `counted_keyword` as supported field type in data streams.

## Checklist

- [x] ~I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~ I updated an existing test package instead

## Related issues

Part of #698

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -7,6 +7,9 @@
   - description: Add support for subobjects in fields
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/573
+  - description: Add support for counted_keyword field type in data streams
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/707
 - version: 3.0.5-next
   changes:
   - description: Disallow to use 'integration' as data stream name
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -17,7 +17,6 @@
     link: https://github.com/elastic/package-spec/pull/704
   - description: Add validation for ingest pipeline rename processor
     type: breaking-change
-    type: enhancement
     link: https://github.com/elastic/package-spec/pull/690
 - version: 3.0.4
   changes:
--- a/spec/integration/data_stream/fields/fields.spec.yml
+++ b/spec/integration/data_stream/fields/fields.spec.yml
@@ -90,6 +90,7 @@ spec:
         - wildcard
         - version
         - unsigned_long
+        - counted_keyword
 
       description:
         description: Short description of field
--- a/spec/integration/data_stream/fields/fields.spec.yml
+++ b/spec/integration/data_stream/fields/fields.spec.yml
@@ -592,6 +592,8 @@ versions:
         path: "/items/allOf/8" # removing subobjects when type is object
       - op: remove
         path: "/items/properties/subobjects"
+      - op: remove
+        path: "/items/properties/type/enum/35" #remove counted_keyword type
   - before: 3.0.3
     patch:
       - op: remove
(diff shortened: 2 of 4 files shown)


spec/changelog.yml
```diff
@@ -7,6 +7,9 @@
   - description: Add support for subobjects in fields
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/573
+  - description: Add support for counted_keyword field type in data streams
+    type: enhancement
+    link: https://github.com/elastic/package-spec/pull/707
```
JonasKunz, line 12: I just noticed that the `Add validation for ingest pipeline rename processor` entry in `3.0.5-next` has a double `type` key: I guess this will likely break the release, shall I fix it in this PR?

I assume the `type` for it should be `breaking-change` instead of `enhancement`?

jsoriano: Yep, this looks incorrect, feel free to make the change here, if not we will do it when we prepare the changelog on next version.

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -90,6 +90,7 @@ spec:
         - wildcard
         - version
         - unsigned_long
+        - counted_keyword
```
jsoriano, line 93: We would need to add a patch to the `versions` at the end of the document, so this is not available in previous versions.

test/packages/good_v3/data_stream/foo/fields/some_fields.yml
```diff
@@ -89,3 +89,8 @@
 - name: empty_expected_values
   type: keyword
   expected_values: []
+- name: some_counted_keyword
+  type: counted_keyword
```
jsoriano, line 93: I have noticed that this type is incompatible with synthetic source. Regarding this change this is fine, but just in case you want to use it in some APM data streams that use it.
```
field [event.counted] of type [counted_keyword] doesn't support synthetic source
```

jsoriano, line 93: It looks like we won't need changes in Fleet, I have tried this definition, and Fleet generates the mapping below, what seems enough. One thing less to do :slightly_smiling_face: 
```
- name: event.counted
  type: counted_keyword
```
```
              "counted": {
                "type": "counted_keyword"
              },
```

jsoriano (changes requested):
We need to add a patch in the `versions` block to remove this type from previous versions. For the rest the change looks good. Thanks!

test/packages/good_v3/data_stream/foo/fields/some_fields.yml
JonasKunz: We we'll use this field type on the data stream storing traces, which does not use synthetic source. But thanks for the heads-up!

jsoriano (approved):

test/packages/good_v3/data_stream/foo/fields/some_fields.yml
JonasKunz: So you mean that the second part you were referring to in [this comment](https://github.com/elastic/package-spec/issues/698#issuecomment-1906367674) is not necessary?

> Also, the change will need to be included in a new package spec minor, probably 3.1, and this version should be used in the kibana configuration for serverless deployments that support it, this can be updated in these files: https://github.com/elastic/kibana/blob/ee11d8bbce96acfc4f1eeff5588087c360b74ab9/config/serverless.oblt.yml#L44

jsoriano: Well, I would still keep it in 3.1 because this is a new feature in the stack, even if it doesn't seem to need explicit support in Kibana.

PR #712, opened 2024-02-07
Discard variables ending with _file or _url as secret candidates

jsoriano:
## What does this PR do?

Disable variables ending with `_file` or `_url` as secret candidates.

## Why is it important?

Reduce the number of false positives. Variables with potential secrets with these names use to refer to files that contain secrets, and not to secrets themselves.

## Checklist

## Related issues

- Proposed by @tetianakravchenko in https://github.com/elastic/integrations/issues/8610#issuecomment-1931620943

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -7,7 +7,7 @@
   - description: Prepare for next version
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/711
-  - description: Discard variables ending with _file as secret candidates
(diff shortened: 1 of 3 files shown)


tetianakravchenko (dismissed):

kpollich (dismissed):

jsoriano:
Excluding also variables ending with `_url`.

mrodm (approved):

Issue #684, opened 2023-12-13
[Change Proposal] Define requirements for agentless support


jsoriano:
There are integrations that could be deployed on fully managed agents. These integrations are going to be called "agentless". We need to define the mechanisms required for packages that include such integrations.

The main proposal would be to have a `supports_agentless` flag that packages that support serverless should set to `true`.

There are some open questions:
* In what manifest to set the flag. Package, policy (for multi-tile) or data stream level?
* Requirements to support agentless:
  * Data should be collected from publicly reachable APIs.
* Are there integrations that _require_ agentless?
* Are there variables that we need to mark as non-configurable for agentless? Variables available only on agentless?
* What is the impact in Kibana UX?
* Testing: Is there any way to validate that a package or data stream actually supports agentless?

Document to facilitate the discussion on what needs to be filtered and where in the Fleet/Integrations UI: https://docs.google.com/document/d/1ccAswS94hr6qNnzQ_HKsmygQch2wgSQYMkMoQHo21Ds/edit?usp=sharing

cc: @kpollich 

## Implementation

- [x] Add a new `deployment_modes` property to the policy template schema that contains an object with `default` and `agentless` properties. Each property in this object should have a single nested `enabled` boolean property for now, but should be extensible to support additional configuration in the future. 
- [x] Add a new hide_in_deployment_modes property to var declarations at all manifest levels that allows integration maintainers to opt certain variables in or out of being displayed in agentless mode. This would be an enum array, supporting `default` and `agentless` strings.

Example policy template declaration:

```yaml
format_version: 3.0.0
name: aws
title: AWS
version: 2.13.1
...
policy_templates:
  - name: billing
    title: AWS Billing
    description: Collect billing metrics with Elastic Agent
    deployment_modes: # <---
      default:
        enabled: false
      agentless:
        enabled: true
    data_streams:
      - billing
    ...
```

Example variable declaration:

```yaml
streams:
  - input: logfile
    vars:
      - name: paths
        type: text
        title: Paths
        multi: true
        required: true
        show_user: true
        default:
          - /var/log/my-package/*.log
      - name: agentless_only
        type: text
        title: Agentless only variable
        multi: false
        required: false
        show_user: true
        hide_in_deployment_modes: # <---
          - default
     - name: hidden_in_agentless 
        type: text
        title: Hidden in agentless variable
        multi: false
        required: false
        show_user: true
        hide_in_deployment_modes: # <---
          - agentless
```

[...]

PR #738, opened 2024-04-15
Add support for definitions related to deployment modes

jsoriano:
## What does this PR do?

Add support for definitions related to deployment modes. It adds two main features:
* Policy templates can indicate on what deployment modes they can be used.
* Variables can be hidden per deployment mode.

It will be released in two versions:
* 3.1.4 will add support for `agentless` mode.
* 3.2.0 will add settings for `default` mode.

`default` mode is introduced in the next minor because it will need changes in the stack to be properly supported, while the changes for agentless will merely be ignored in stacks that don't support it.

## Why is it important?

With the introduction of new deployment modes we need ways to define policy templates or use variables only on certain deployment modes.

## Checklist

## Related issues

- Part of https://github.com/elastic/package-spec/issues/684


kpollich (approved):
Code + test package LGTM, and the release/version strategy makes sense to me. Thanks for tackling this! 🚀

mrodm (approved):
:+1:

PR #739, opened 2024-04-16
Add agent fields into config file for system tests

mrodm:
## What does this PR do?

Add agent fields into the test system configuration files that would be used to customize Elastic Agents.

## Why is it important?

These new settings would allow the developers to specify the required settings (user, pid mode, linux capabilities...) to run the tests of the given Data Stream of that package.

## Checklist

    - Added a new Data Stream to `good_v3` test package.

## Related issues

- Depends on https://github.com/elastic/elastic-package/issues/787


spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -11,4 +11,99 @@ spec:
     wait_for_data_timeout:
       description: Timeout for waiting for metrics data during a system test.
       type: string
-      example: 10m
\ No newline at end of file
+      example: 10m
+    agent:
```
mrodm, line 15: I did not set any JSON patch since I thought it was better to allow all packages to define these keys. WDYT ?

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -11,4 +11,99 @@ spec:
...
+              - "127.0.0.1:7443:7443"
+              - "8000:8000/udp"
+        linux_capabilities:
+          description: "Linux Capabilities that must been enabled in the system to run the Elastic Agent process"
+          type: array
+          items:
+            type: string
+            description: Capability name
+            # https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
+            enum:
```
mrodm, line 55: Do you think it should be added `ALL` here to this list too?

spec/integration/data_stream/_dev/test/system/config.spec.yml
jsoriano: Yes, no need to add a json patch here.

spec/integration/data_stream/_dev/test/system/config.spec.yml
jsoriano: We can add it later if needed, I think it would be better if developers need to check the capabilities the package really needs.

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -11,4 +11,99 @@ spec:
...
+        custom_script:
+          description: "Optional: Custom script to be run to update environment where Elastic Agent runs (e.g. installing new libraries/dependencies)"
+          type: string
+          default: ""
+        pre_start_script:
+          description: "Optional: Custom script to be run to set required changes for the Elastic Agent process (e.g. export environmen variables)"
+          type: string
+          default: ""
+      required:
+        - runtime
```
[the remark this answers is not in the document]
jsoriano: Yes, we can make `docker` the default.

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -23,3 +23,125 @@ spec:
       items:
         type: string
         example: "error.message"
+    agent:
+      description: "Configuration overrides for the Elastic Agent"
+      type: object
+      additionalProperties: false
```
mrodm, line 29: Would it be better to let define new properties here to allow `elastic-package` add new settings easily? @jsoriano

jsoriano: Yes, I think we can be less restrictive here than in other places.

jsoriano (approved):
LGTM, and looks good to me also to allow additional properties in the `agent` object.

jsoriano (approved):

PR #743, opened 2024-05-03
Add accept_ignored_values to field spec

flash1293:
## What does this PR do?

This PR adds a new property `skip_ignored_fields` to the system test spec which is planned to be used in the `elastic-package` system tests.

## Why is it important?

As discussed in https://github.com/elastic/elastic-package/pull/1738 , it's sometimes important to mute errors during testing because they are not indicative of a real problem but related to the test setup.

This addition allows to override test failures similar to comments muting linting errors in source code.

## Checklist


test/packages/good/data_stream/foo/fields/some_fields.yml
```diff
@@ -27,6 +27,11 @@
   description: A field with a pattern defined
   pattern: '^[a-zA-Z]$'
   example: abcd
+- name: foobar
+  type: keyword
+  ignore_above: 5
+  description: A field that can be ignored without failing tests
+  accept_ignored_values: true
```
ruflin, line 34: Brainstorming around a better name:

* Something like skip or ignore, not validate
* Something around errors?
* Unfortunately already called ignore 

`skip_ignore_above`, `skip_ignore_above_error`? -> I made it specific as I expect in the future we can tell apart the different reasons. Maybe for now `skip_ignore_error(s)`

Did you used `ignored` on purpose vs `ignore`?

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -318,6 +318,16 @@ spec:
         type: integer
         default: 1024
 
+      accept_ignored_values:
+        description: >
+          If set to true, elastic-package system tests will not fail if values
+          for this field can't be indexed for any incoming documents. This
+          should only be used if the failure is related to the test environment
+          and wouldn't happen in production, mitigate the issue via mapping or
+          pipeline changes otherwise. Defaults to false.
```
jsoriano, line 327: Are there actual cases where we want to ignore these failures? Or this is a temporary measure till this issue is fixed in packages?

If there are actual cases where we want to ignore these failures, and as described here, this is only expected in tests, I think we should add this to the test config files. We have other test settings like `skip`, `numeric_keyword_fields` or `dynamic_fields` that are used to customize how validations behave on certain cases. I think this setting would fit there.

If this is a temporary measure to help merging https://github.com/elastic/elastic-package/pull/1738 before the issues are solved, we can think on other approaches. Something we have done in some cases is to apply new validations only on packages starting on a specific spec version. On these cases the issue is eventually fixed as packages upgrade their spec versions.

test/packages/bad_accept_ignore_above/changelog.yml
```diff
@@ -0,0 +1,26 @@
+- version: 1.0.0
```
jsoriano, line 1: Test packages need to be referenced in `code/go/pkg/validator/validator_test.go` to be actually used in testing.

jsoriano (changes requested):
If needed, I think we should make this a test config setting rather than part of the field definition.

spec/integration/data_stream/fields/fields.spec.yml
flash1293: > Are there actual cases where we want to ignore these failures? Or this is a temporary measure till this issue is fixed in packages?

There is at least one case in spring boot.

The test config files are indeed a better place, didn't think of those - I will adjust.

flash1293:
@ShourieG This was an old version, I just pushed a change, could you have a look? I added it to the system test definition now.

code/go/pkg/validator/validator_test.go
```diff
@@ -49,7 +49,7 @@ func TestValidateFile(t *testing.T) {
 			},
 		},
 		"bad_deploy_variants": {
-			"_dev/deploy/variants.yml",
+			"_dev/deploy/variants.yVml",
```
ShourieG: @flash1293 is this a typo here ? Noticed this while going through the change.

flash1293:
>  is there any other way we can include this, example adding it in the validations.yml file

I think it should be a per-field setting, not a wholesale config to ignore all ignored fields in all tests as this would shadow newly introduced issues. Is there a way to do this without releasing a new package spec version?

flash1293:
@ShourieG I might misunderstand, but in order to extend the `validation.yml` to also take a list of fields to not check `_ignored` for, don't we need to release a version of the package spec as well? By my understanding currently `validation.yml` can only completely exclude certain checks, which is not what we want here

flash1293:
Hmm, maybe we could have something like `SVR00006:<field name>` would be an option (but it still seems like a workaround):

```yaml
errors:
  exclude_checks:
   - "SVR00006:error.message" # do not fail on error.message getting ignored
```

What do you think @jsoriano ? To me it seems like the system test config is the right place for this, but I don't have a strong opinion really.

jsoriano:
> What do you think @jsoriano ? To me it seems like the system test config is the right place for this, but I don't have a strong opinion really.

Yeah, I prefer to have it in the test config.

> Having this tied to the package-spec forces us to increase the format_version

The spec for test config files is barely validated, we accept almost anything there (due to this [`additionalProperties: true`](https://github.com/elastic/package-spec/pull/743/files#diff-5b01b03352fa50eee40a9e3b46fc32433be1958ede274d9c00dc7c203c07a1c1R7)), so using this feature won't force to increase the `format_version`, at least as it is now being considered.

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -11,4 +11,15 @@ spec:
...
-      example: 10m
\ No newline at end of file
+      example: 10m
+    skip_ignored_fields:
+      description: >
+        If listed here, elastic-package system tests will not fail if values
+        for the specified field names can't be indexed for any incoming documents. This
+        should only be used if the failure is related to the test environment
+        and wouldn't happen in production. Mitigate the issue via mapping or
+        pipeline changes otherwise.
```
jsoriano, line 21: :+1:

test/packages/good/data_stream/foo/_dev/test/system/test-default-config.yml
```diff
@@ -1,2 +1,4 @@
 wait_for_data_timeout: 10m
+skip_ignored_fields:
+  - error.message
```
jsoriano, line 3: Nit. Could we also add a good example in the good_v3 package?

code/go/pkg/validator/validator_test.go
```diff
@@ -132,6 +132,12 @@ func TestValidateFile(t *testing.T) {
 				"field streams.0.vars.3: Must not be present",
 			},
 		},
+		"bad_skip_ignored_fields": {
```
jsoriano, line 135: This package seems to have more issues, did you create it by copying an old package? Consider creating it from scratch with `elastic-package create package` and `elastic-package create data-stream`.

jsoriano (commented):
This approach LGTM, let's solve the issues with CI. Thanks!

jsoriano:
As this change is not required for https://github.com/elastic/elastic-package/pull/1738, could you include the implementation for this feature there before merging this?
Thanks!

jsoriano (approved):
Please consider adding the implementation for this in https://github.com/elastic/elastic-package/pull/1738 before merging :pray: 

Thanks!

flash1293:
https://github.com/elastic/elastic-package/pull/1738 is merged, could you merge @jsoriano ? (I don't have the permission to do so)

jsoriano:
Merged, thanks!

PR #748, opened 2024-05-14
Add support for store mapping property

jsoriano:
## What does this PR do?

Add support for the [`store` mapping option](https://www.elastic.co/guide/en/elasticsearch/reference/8.13/mapping-store.html).

## Why is it important?

When synthetic mode is used, there are some cases where we may need to ensure that some fields are stored. See for example https://github.com/elastic/integrations/pull/9826.

## Checklist

## Related issues

- Needed for uses cases like https://github.com/elastic/integrations/pull/9826 in recent versions of the spec.


jsoriano:
This change is not enough, this needs to be supported by Fleet, I will move the change to 3.2.0.

compliance/features/synthetic-source.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Synthetic source
+  Support synthetic source
+
+  @3.2.0
+  Scenario: Installer leverages source true
+   Given the "logs_synthetic_mode" package is installed
+     And a policy is created with "logs_synthetic_mode" package
+    Then index template "logs-logs_synthetic_mode.synthetic" has a field "decision_id" with "store:true"
```
mrodm, line 8: I thought as it required changes in Fleet, this test would fail

mrodm: Ok, I think as it requres spec 3.2.0, this complicance test is not tested:
https://github.com/elastic/package-spec/blob/a32f97206dccb24ff6343028d9ef54898b585f75/.buildkite/pipeline.trigger.compliance.tests.sh#L29-L31

Should we take the opportunity to set 8.13.4 as Elastic stack version for the spec version 3.1.0? 

Even if we add there `compliance_test 8.14.0-SNAPSHOT 3.2.0`, this test would fail because of the lack of support, am I right ?

jsoriano: > Should we take the opportunity to set 8.13.4 as Elastic stack version for the spec version 3.1.0?

Or I can change to 8.14.0-SNAPSHOT, yes.

> Even if we add there compliance_test 8.14.0-SNAPSHOT 3.2.0, this test would fail because of the lack of support, am I right ?

Correct, we cannot enable 3.2.0 yet.

mrodm (approved):
:+1:

PR #759, opened 2024-06-05
Add main configuration file for tests

mrodm:
## What does this PR do?

Adds a new configuration file to define global settings for tests: `_dev/test/config.yml`

## Why is it important?

This configuration file is going to be used to define if tests of specific runners could be run in parallel or they should be run sequentially.

Example of the contents of this configuration file:
```
system:
  parallel: true
policy:
  parallel: false
  skip:
    reason: flaky tests
    link: https://github.com/elastic/package-spec/issues/1
```

## Checklist

## Related issues

- Part of https://github.com/elastic/elastic-package/issues/787

### DIFF
--- a/test/packages/good_input/_dev/test/config.yml
+++ b/test/packages/good_input/_dev/test/config.yml
@@ -2,3 +2,6 @@ system:
   parallel: true
 policy:
   parallel: false
+  skip:
+    reason: ignoring all system tests
+    link: https://github.com/elastic/package-spec/issues/1
--- a/test/packages/good_input/_dev/test/config.yml
+++ b/test/packages/good_input/_dev/test/config.yml
@@ -1,7 +1,7 @@
 system:
-  sequential: true
+  parallel: true
 policy:
-  sequential: false
+  parallel: false
   skip:
-    reason: ignoring all policy tests
+    reason: ignoring all system tests
     link: https://github.com/elastic/package-spec/issues/1
--- a/test/packages/good_input/_dev/test/config.yml
+++ b/test/packages/good_input/_dev/test/config.yml
@@ -1,0 +1,4 @@
+system:
+  parallel: true
+policy:
+  parallel: false
--- a/test/packages/good_input/_dev/test/config.yml
+++ b/test/packages/good_input/_dev/test/config.yml
@@ -1,7 +1,7 @@
 system:
-  parallel: true
+  sequential: true
 policy:
-  parallel: false
+  sequential: false
   skip:
-    reason: ignoring all system tests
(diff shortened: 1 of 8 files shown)


test/packages/good_v3/_dev/test/config.yml
```diff
@@ -0,0 +1,10 @@
+system:
+  parallel: true
+pipeline:
+  parallel: false
+asset:
+  parallel: false
+policy:
+  parallel: false
+static:
+  parallel: false
```
jsoriano, line 10: Nice! Maybe we can also use this common config to skip all the tests of a kind.

test/packages/good_input/_dev/test/config.yml
```diff
@@ -0,0 +1,4 @@
+system:
+  parallel: true
+policy:
+  parallel: false
```
jsoriano, line 4: Have you considered having one file per test directory? I guess it depends on what is going to read this config. If it is going to be read by the commands it is fine here, if it is going to be read by the runners, I think it would be better in each test directory. But not a strong opinion.

test/packages/good_v3/_dev/test/config.yml
mrodm: Adding a new field `skip` in every kind of tests ? Something like this? 
```yaml
static:
  parallel: false
  skip: true
```

IIUC that would avoid to add the skip field in every test config file?

test/packages/good_input/_dev/test/config.yml
mrodm: These files should be read/parsed before triggering all tests.
Currently, this means to be read in the command definitions. There is where the tests are triggered.

There is an option to move all the logic to get all tests to each runner. If so, then these files could be read in each runner before triggering any test.

test/packages/good_v3/_dev/test/config.yml
jsoriano: Yes, this would be useful for example to disable all system tests if there is some issue with the service deployment.

mrodm: Added `skip`. :+1: 

I also renamed `parallel` to be `sequential`, not sure if that would be better or not.

Thinking in `elastic-package` , if `sequential` is not defined it means is `false` by default, and therefore tests could be run in parallel.

WDYT ?

jsoriano: I think I prefer `parallel` to `sequential`, but not a strong opinion, as you prefer.

jsoriano (approved):

Issue #435, opened 2022-10-11
Including SLO assets in Integrations.


agithomas:

jsoriano:
Transferred to package-spec to discuss about how this can be modelled in packages.

ruflin:
Here some more details around SLO: https://github.com/elastic/kibana/issues/137323

@emma-raffenne @fkanout @kdelemme Would be great if someone of the team around SLO could chime in.

kdelemme:
Adding @simianhacker 

I'm not sure to fully understand what is being suggested / built here. But happy to schedule a call to discuss the current SLO work being done by the @elastic/actionable-observability team, and understand your need better.

kdelemme:
Spoke with @ruflin earlier today, and integration team can use the SLO api to provide out of the box SLO for some integrations. 
Ideally a default SLO could be configured (and tweaked) during the integration installation.

When creating an SLO, the system takes care of creating the necessary resources and starts aggregating the data into a specific index. We don't support _enabling_ or _disabling_ an SLO. Therefore, if an SLO is created, the user has to delete it to stop data aggregation, or a manual call to the transform API could be used to stop it.

The SLO API is not GA yet :)

ruflin:
As @kdelemme described, it seems the only missing bit is a enabled / disable feature, I think that would not only be useful for the Fleet scenarios but also for users that want to pause a SLO without having to learn about what transforms are etc.

The SLO API input has a neat structure. @kdelemme could you post 1-2 json examples? The reason is that I assume this is what would become part of the package spec and like this we can share it here with the team.

kdelemme:
Regarding enabling and disabling an SLO, @vinaychandrasekhar what do you think?

@ruflin Here's some examples with different SLI and configuration options:

<details>
<summary>SLO Availability based on APM transaction error, using 30d rolling and occurrences method</summary>

```
curl --request POST \
  --url http://localhost:5601/qca/api/observability/slos \
  --header 'Authorization: Basic ZWxhc3RpYzpjaGFuZ2VtZQ==' \
  --header 'Content-Type: application/json' \
  --header 'kbn-xsrf: oui' \
  --data '{
	"name": "My SLO Availability - rolling 30d",
	"description": "Some description",
	"indicator": {
		"type": "slo.apm.transaction_error_rate",
		"params": {
			"environment": "development",
			"service": "o11y-app",
			"transaction_type": "request",
			"transaction_name": "GET /flaky",
[...]

PR #767, opened 2024-06-24
Add SLO assets to the package spec

mgiota:
Fixes https://github.com/elastic/observability-dev/issues/3602


spec/integration/kibana/spec.yml
```diff
@@ -125,7 +125,17 @@ spec:
     name: "tags.yml"
     required: false
     $ref: "./tags.spec.yml"
-
+  - description: Folder containing Kibana SLO assets
+      type: folder
+      name: slo
```
mgiota, line 130: @jsoriano I have a draft PR here that adds the SLO assets to the kibana spec. I plan to add some unit tests in the PR, but here's what CI is complaining about at the moment:

`could not load spec for "integration/kibana/spec.yml": could not parse folder specification file: yaml`

I have a couple of questions:
- Could you further assist why I am getting this error? I gave the name `slo` to the new folder. Shall I specify somewhere else what is considered a valid folder name? 
- Where shall I write my unit tests?
- I was reading the [Contributing guide](https://github.com/elastic/package-spec/blob/main/CONTRIBUTING.md). Shall I create a [new Change Proposal issue](https://github.com/elastic/package-spec/issues/new/choose)?

mgiota: > CI is complaining

 It turns out it was an indentation issue, which was fixed in this [commit](https://github.com/elastic/package-spec/pull/767/commits/582c5c6970cfe5a9cdf5527816e477fdfccff027)

> Where shall I write my unit tests?

I added following slo examples:
- `test/packages/good/kibana/slo/good-slo-abc-1.json`
- `test/packages/good_v2/kibana/slo/good_v2-slo-abc-1.json`
- `test/packages/good_v3/kibana/slo/good_v3-slo-abc-1.json`

> Shall I create a [new Change Proposal issue](https://github.com/elastic/package-spec/issues/new/choose)?

I didn't create any new issue, cause I found [this one](https://github.com/elastic/package-spec/issues/435).

mgiota:
@jsoriano I open up this PR for review. Let me know what other changes I might need to do that I am not aware of. Thanks!

mgiota:
test integrations

mgiota:
I was following the [Contributing guide](https://github.com/elastic/package-spec/blob/main/CONTRIBUTING.md) and I commented `test integrations` in this PR, but I am not sure why the auto generated [PR](https://github.com/elastic/integrations/pull/10231) is failing. Could you help me out here?

spec/integration/kibana/spec.yml
```diff
@@ -125,7 +125,17 @@ spec:
...
+    name: slo
+    required: false
+    contents:
+    - description: An SLO asset file
+      type: file
+      contentMediaType: "application/json"
+      pattern: '^{PACKAGE_NAME}-.+\.json$'
+      forbiddenPatterns:
+        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden 
 versions:
```
jsoriano, line 139: We could add a patch here to ensure that this feature is only available in versions of the spec implemented by Kibana. This would likely mean to add something like this:
```suggestion
versions:
  - before: 3.3.0
    patch:
      - op: remove
        path: "/contents/13" # remove SLO definitions
```

Or 3.2.1 if we decide to add it to the next patch version.

jsoriano (commented):
Change LGTM, but we have to decide on what version we would release it, and add a changelog entry.

To decide what version to use, the main question would be regarding the level of support in Kibana. If a package using this feature requires changes in Kibana to work properly, then we should add the change into `3.3.0-next`, if not, we can add it in `3.2.1-next`.

We release almost everything on patch versions, we use minors for changes that require also changes in Kibana, and majors for breaking or significant changes in the spec itself.

jsoriano:
> I am not sure why the auto generated [PR](https://github.com/elastic/integrations/pull/10231) is failing. Could you help me out here?

The failures there are not related to this change, no worries :slightly_smiling_face: aws package uses to be flaky, and azure_blob_storage has a known issue.

spec/integration/kibana/spec.yml
muthu-mps: @mgiota - Do we need to load the SLO's inside the Kibana directory? Are there any guideline on defining SLO's.

mgiota: @muthu-mps I added it here as per @jsoriano 's recommendation in the past. Here's the [public SLO API](https://docs.elastic.co/api-reference/observability/createsloop) and the [SLO Elastic documentation ](https://www.elastic.co/guide/en/observability/current/slo-create.html)

In this PR I added some examples of SLO creation.

mgiota: In this [commit](https://github.com/elastic/package-spec/pull/767/commits/d78edf99120d88f117ce067e7b560e131fd7f053) I updated the README file to include the SLO assets. As you can see there are many other Kibana assets in the list as for example rules, ml modules, dashboards and I guess it makes sense to add `slo` here as well.

mgiota:
> To decide what version to use, the main question would be regarding the level of support in Kibana. If a package using this feature requires changes in Kibana to work properly

As far as I understand there will be some changes in the Fleet module in kibana to make the whole integration work. So I guess we should add the change into `3.3.0-next`. 

> add a changelog entry
I found a couple of changelog files. Where exactly should we add the new SLO asset support? Or is the change you recommnded in the spec file enough?

mgiota:
@jsoriano Now I am getting some validation errors:

```
found 1 validation error:
1. item [slo] is not allowed in folder [../../../../test/packages/good_v2/kibana]        	            	   
```

Do I need to specify the new `slo` item anywhere else? The tests started failing after I committed the change with the patch version. So I guess I am missing something else as well.

jsoriano:
> Do I need to specify the new `slo` item anywhere else? The tests started failing after I committed the change with the patch version. So I guess I am missing something else as well.

The problem is that the `good` package uses an old version of the spec, and packages with SLOs will need to have at least a `format_version` equal or greater than `3.3.0`.

I would recommend to move the test files you added to the `good_v3` package and increase its `format_version` to 3.3.0.

You can run these tests locally for faster feedback, with `go test ./...` from the root directory of the repository.

muthu-mps:
@mgiota , @jsoriano - Just FYI, The directory structure we follow for the integrations to include SLO's and the general guidelines around using the ID's as name of the json file is still in discussion. We can continue those discussions in the issue. This may evolve going forward.

muthu-mps (approved):

mgiota:
@jsoriano Looks like I am not authorized to merge this PR. Could you either merge the PR or grant me access to merge it?


Screenshot 2024-06-25 at 14 57 19

ruflin:
@mgiota Thank you for making this happen!!!

PR #769, opened 2024-06-26
Make missing dataset field error more informative

efd6:
## What does this PR do?

Explains in greater detail how a user can resolve the error presented.

## Why is it important?

The existing error does not have enough information for a user to easily understand why the error is being printed.

## Checklist

## Related issues

-


jsoriano (approved):
Thanks!

efd6:
@jsoriano Would you merge this? I cannot.

PR #788, opened 2024-08-28
Add agent.base_image setting in system configuration file

mrodm:
## What does this PR do?

Add a new setting into the system test configuration file to indicate the Elastic Agent base image to be used for testing.

## Why is it important?

This new setting allows developers to set the required Elastic Agent base image for their packages.

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-package/issues/2037

### DIFF
--- a/spec/integration/data_stream/_dev/test/system/config.spec.yml
+++ b/spec/integration/data_stream/_dev/test/system/config.spec.yml
@@ -42,7 +42,6 @@ spec:
           enum:
             - default
             - complete
-            - minimal
           default: "default"
         user:
           description: "User that runs the Elastic Agent process"
--- a/spec/integration/data_stream/_dev/test/system/config.spec.yml
+++ b/spec/integration/data_stream/_dev/test/system/config.spec.yml
@@ -36,8 +36,10 @@ spec:
           default: docker
         type:
           description: >
-            "Elastic Agent image to be used for testing. Setting default will be used the
-            same Elastic Agent image as the stack."
+            Elastic Agent image to be used for testing. Setting `default` will be used the
+            same Elastic Agent image as the stack. Setting `complete` will use the "complete"
+            image, that includes all supported collectors, a web browser and the required runtime
+            for synthetic testing.
           type: string
           enum:
             - default
--- a/spec/integration/data_stream/_dev/test/system/config.spec.yml
+++ b/spec/integration/data_stream/_dev/test/system/config.spec.yml
@@ -34,7 +34,7 @@ spec:
           enum:
             - docker
           default: docker
-        type:
+        base_image:
           description: >
             Elastic Agent image to be used for testing. Setting `default` will be used the
             same Elastic Agent image as the stack. Setting `complete` will use the "complete"
--- a/spec/integration/data_stream/_dev/test/system/config.spec.yml
+++ b/spec/integration/data_stream/_dev/test/system/config.spec.yml
@@ -34,6 +34,16 @@ spec:
           enum:
             - docker
           default: docker
+        type:
+          description: >
+            "Elastic Agent image to be used for testing. Setting default will be used the
+            same Elastic Agent image as the stack."
+          type: string
+          enum:
+            - default
+            - complete
+            - minimal
+          default: "default"
         user:
           description: "User that runs the Elastic Agent process"
           type: string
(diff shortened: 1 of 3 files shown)


spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -34,6 +34,16 @@ spec:
...
           default: docker
+        type:
+          description: >
+            "Elastic Agent image to be used for testing. Setting default will be used the
+            same Elastic Agent image as the stack."
+          type: string
+          enum:
+            - default
+            - complete
+            - minimal
```
mrodm, line 45: Complete would mean to use `elastic-agent-complete` image for docker.
Minimal would mean to use `elastic-agent` image for docker.
Default the same image as the Elastic Agent from the stack.

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -34,6 +34,15 @@ spec:
...
           default: docker
+        type:
+          description: >
+            "Elastic Agent image to be used for testing. Setting default will be used the
+            same Elastic Agent image as the stack."
+          type: string
+          enum:
+            - default
+            - complete
+          default: "default"
```
mrodm, line 45: Complete would mean to use elastic-agent-complete image for docker.
Default the same image as the Elastic Agent from the stack.

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -34,6 +34,15 @@ spec:
           enum:
             - docker
           default: docker
+        type:
```
jsoriano, line 37: Nit. Maybe we can be more explicit on the name of this parameter. But I don't have a better idea for this, maybe `base_image`, or `base_image_type`?
```suggestion
        base_image:
```

We have to keep in mind that this should be runtime-agnostic, but I think "image" is agnostic enough.

spec/integration/data_stream/_dev/test/system/config.spec.yml
```diff
@@ -34,6 +34,15 @@ spec:
...
+          description: >
+            "Elastic Agent image to be used for testing. Setting default will be used the
+            same Elastic Agent image as the stack."
```
jsoriano, line 40: Quotes not needed here.

Andd add here the description for the `complete` image too?
```suggestion
            Elastic Agent image to be used for testing. Setting `default` will be used the
            same Elastic Agent image as the stack. Setting `complete` will use the "complete"
            image, that includes all supported collectors, a web browser and the required runtime
            for synthetic testing.
```

spec/integration/data_stream/_dev/test/system/config.spec.yml
mrodm: I was thinking in that and I had the same doubts.
I was wondering what would mean if at some point it is supported as runtime VMs.
Would it be used docker image inside that VM too ?

jsoriano: > I was wondering what would mean if at some point it is supported as runtime VMs.
> Would it be used docker image inside that VM too ?

I don't think so. Ideally we would install the agent directly in the VMs, or use an image that has the agent installed as a package or from tar/zip.
For the case of "complete" maybe we need to provide a different base image that includes the dependencies.
In that sense `base_image` would be fine and agnostic enough. It would not refer specifically to a docker image, but to a image in the general sense.

Another approach we could follow is to define this as dependencies or dependency sets:
```
depends:
  - synthetic_testing
```

This indicates the runtime that the test depends on the agent running on a system with capability to run synthetic tests. This is resolved in the docker runtime by using the complete image, but could be resolved in other systems in different ways.

mrodm: > In that sense base_image would be fine and agnostic enough. It would not refer specifically to a docker image, but to a image in the general sense.

Using `base_image` looks good for me.

It would be like defining which kind of elastic-agent artifact/tar/zip it is required for that system test in a given package.

> Another approach we could follow is to define this as dependencies or dependency sets

This approach would allow developers to be more detailed to specify the requirements of each test. However, not sure if that could be too fine grained and therefore complicate the logic to select the required artifact/zip/docker.

jsoriano: > > Another approach we could follow is to define this as dependencies or dependency sets
> 
> This approach would allow developers to be more detailed to specify the requirements of each test. However, not sure if that could be too fine grained and therefore complicate the logic to select the required artifact/zip/docker.

Agree, so far we have only had two kinds of images and I don't think we are going to have many more cases.

jsoriano (approved):

Issue #784, opened 2024-08-09
Misleading implementation of mappings of nested object subfields


jsoriano:
We have found that fields with `type: nested` and subfields don't get the mappings for the subfields in the final index template, this can lead to unexpected mappings.

That is that for example for [a mapping with several subfields](https://github.com/elastic/integrations/blob/81da2868b3796fa0bb4664a9c6aaed6939ccb574/packages/tanium/data_stream/threat_response/fields/fields.yml#L111), this empty nested mapping is generated:
```
"whats": {
    "type": "nested"
}
```
In combination with dynamic mappings introduced by `ecs@mappings` this can lead subfields to have completely unexpected mappings.

While investigating we have found that the implementation in Fleet has two different codepaths for [`type: nested`](https://github.com/elastic/kibana/blob/6d46edc275ba670a885acfc40e145a99b7abbf78/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L567) and for [`type: group-nested`](https://github.com/elastic/kibana/blob/6d46edc275ba670a885acfc40e145a99b7abbf78/x-pack/plugins/fleet/server/services/epm/elasticsearch/template/template.ts#L516). For `nested` it ignores any subfieds, for `group-nested` it generates the expected mappings. We have also found that package spec only allows subfields with the `group` and `nested` fields, so any workaround based on the use of `group-nested` would not work at the moment.

<s>So we need to align the behavior of Fleet, package-spec, and the current assumptions in existing integrations. Probably doing the following:
* Assume that `nested` and `group-nested` are synonyms, even if they were not originally, we are in a situation now where `nested` is used as `group-nested`, and `group-nested` is not used.
* Replace implementation of `nested` with the implementation of `group-nested` in Fleet, so they effectively behave the same.
* Allow in package spec the use of `group-nested` in all cases where `nested` is allowed, this will allow to fix issues in packages for current versions of packages.
  * https://github.com/elastic/package-spec/pull/789
* There are two additional parameters for nested objects called `include_in_parent` and `include_in_root`, we have only found uses of them in tests, and not on any actual package. We will need to confirm its expected behaviour and decide if we need to keep them.
  * Decided to keep them if possible, just in case they are needed, but no integration is using them at the moment. 
* Consider replacing `nested` with `group-nested` in affected packages.</s>

[...]

PR #789, opened 2024-08-29
Allow to use group-nested wherever nested can be used

jsoriano:
nested and group-nested have different implementations in Fleet, for nested it doesn't generate any mapping for the subfields, but for group-nested it does. Integrations are using nested expecting that the subfields are going to have mappings.

Add group-nested to the spec, for all versions, as a synonymous of nested, so integrations can use it when supporting older versions of the stack. nested type should be fixed in Fleet to work also the same as group-nested.

Part of https://github.com/elastic/package-spec/issues/784

### DIFF
--- a/code/go/pkg/validator/validator_fields_test.go
+++ b/code/go/pkg/validator/validator_fields_test.go
@@ -95,7 +95,7 @@ func TestValidateFields(t *testing.T) {
 				},
 			},
 			expectedErrors: []string{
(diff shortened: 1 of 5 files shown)


jsoriano:
Moving to draft. After checking implementation in Kibana it looks like `group-nested` is only used internally.

jsoriano:
Closing, it is probably enough with fixing the implementation in Fleet and adapting the affected packages for older versions.

PR #796, opened 2024-09-09
Update description agent.base_image field

mrodm:
## What does this PR do?

Update description `agent.base_image` field for system tests configuration files.

## Checklist

- ~[ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~
- ~[ ] I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).~

## Related issues

- Follows #788

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -12,6 +12,11 @@
   - description: Add support for `slo` assets.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/767
+- version: 3.2.3-next
+  changes:
+  - description: Update description of agent.base_image in system tests configuration files.
+    type: bugfix
+    link: https://github.com/elastic/package-spec/pull/796
 - version: 3.2.2
   changes:
   - description: Add spec for img directory.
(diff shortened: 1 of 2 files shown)


spec/changelog.yml
```diff
@@ -12,6 +12,11 @@
   - description: Add support for `slo` assets.
     type: enhancement
     link: https://github.com/elastic/package-spec/pull/767
+- version: 3.2.3-next
+  changes:
+  - description: Update description of agent.base_image in system tests configuration files.
+    type: bugfix
+    link: https://github.com/elastic/package-spec/pull/796
```
jsoriano, line 19: Maybe we don't need to add changelog for this.

jsoriano (approved):

spec/changelog.yml
mrodm: Ok, as it is just a description change, no need to do this.
Thanks!
I'll remove this changelog entry

jsoriano (approved):

Issue #795, opened 2024-09-06
[Change Proposal] When the deployment_mode agentless is enabled we want the package to have additional fields populated.


seanrathier:
#### Motivation
We would like to have additional fields required to be populated in an integration package when the `deployment_mode` `agentless` is enabled. These fields are for identifying the organization, division, and team responsible for the integration. This information will be used to tag the agentless agent deployed in MKI so that we know who to contact when support is needed and for the teams to monitor their agentless agents.

#### Some open questions
- Should these new fields be siblings of the `deployment_mode.agentless`?
- Can we require fields to be populated by the integration developer if parent or sibling fields are enabled?

For example....
```
format_version: 3.0.0
name: aws
title: AWS
version: 2.13.1
...
policy_templates:
  - name: billing
    title: AWS Billing
    description: Collect billing metrics with Elastic Agent
    deployment_modes: # <---
      default:
        enabled: false
      agentless:
        enabled: true
        organization: 'Security'
        division: 'Cloud Security'
        team:  'Cloud Security Posture Management'
    data_streams:
      - billing
```

#### Related Issues
- https://github.com/elastic/package-spec/issues/684
- https://github.com/elastic/security-team/issues/10448
- https://github.com/elastic/agentless-api/issues/254

jsoriano:
> This information will be used to tag the agentless agent deployed in MKI

What is the plan for these tags to reach MKI from the manifest? Will this be done by Fleet?

> * Should these new fields be siblings of the `deployment_mode.agentless`?

This sounds like a reasonable option, yes.

> * Can we require fields to be populated by the integration developer if parent or sibling fields are enabled?

We could add these fields as required, but only for new versions of the spec, to avoid breaking existing packages.

jsoriano:
Ok, if Kibana/Fleet is the one reading the manifest the plan sounds good to me.

seanrathier:
CC: @eyalkraft @oren-zohar @acorretti  @orestisfl

seanrathier:
Full disclosure this is my first time doing this,  I've started a PR here
- https://github.com/elastic/package-spec/pull/801

PR #801, opened 2024-09-12
Add support for integration deployment mode agentless to include identity fields

seanrathier:
## What does this PR do?
We would like to have additional fields required to be populated in an integration package when the deployment_mode agentless is enabled. These fields are to identify the organization, division, and team responsible for the integration. 

## Why is it important?

This information will be used to tag the agentless agent deployed in MKI so that we know who to contact when support is needed and for the teams to monitor their agentless agents.

## Checklist

## Related issues
- https://github.com/elastic/package-spec/issues/795


spec/integration/manifest.spec.yml
```diff
@@ -174,6 +174,38 @@ spec:
...
+            allOf:
+              - if:
+                  properties:
+                    enabled:
+                      const: true
+                then:
+                  required:
+                    - organization
+                    - division
+                    - team
```
jsoriano, line 207: Setting these new fields as required may break existing packages using `agentless`, do we have any?

spec/changelog.yml
```diff
@@ -4,461 +4,464 @@
...
-  - description: Validate that data streams with time series enabled have configured dimensions
-    type: enhancement
-    link: https://github.com/elastic/package-spec/issues/655
-  - description: Fix spec for rally benchmark scenario
-    type: bugfix
-    link: https://github.com/elastic/package-spec/pull/653
+    - description: Using non-GA versions of the spec in GA packages produces a filterable validation error instead of a warning
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/627
+    - description: "Allow to set index: false in dynamic templates defined in data stream manifests"
```
jsoriano, line 140: There are many changes in formatting and quotes in this PR. I would prefer to avoid them if not needed.

spec/changelog.yml
```diff
@@ -4,461 +4,464 @@
...
   changes:
-  - description: Add support for `slo` assets.
-    type: enhancement
-    link: https://github.com/elastic/package-spec/pull/767
+    - description: Add support for `slo` assets.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/767
+    - description: Add support for identity of agentless deployment mode.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/801
```
jsoriano, line 17: Please put this change in a new `3.2.3-next` version, we add changes to new minors only if they require changes in Kibana for the package to properly work.

Also, maybe we should add a breaking change notice mentioning that we are adding required fields.

spec/integration/manifest.spec.yml
seanrathier: Not at them moment.  AFAIK there are some integrations like Okta and Cloud Connector that want to use this setting.

@eyalkraft can you confirm?

spec/changelog.yml
seanrathier: My linting picked it up changed it automatically, I noticed some other settings used double quotes. I will revert those formatting changes.

spec/changelog.yml
seanrathier: Is this the only place where we should specify the breaking change?

https://github.com/elastic/package-spec/pull/801/files#diff-2782fabcaeaa32cbab3eb1fabfbae4aa717d626056e383732eb8e014797fadbfR18

spec/integration/manifest.spec.yml
jsoriano: > How could we still make these fields required? Can the integrations be updated first? Or could we provide a default value?

We cannot update the integrations first because the field is not allowed yet. We have two options:
* We add the fields as now, but we make them mandatory only for packages using spec >= 3.2.3. Then integrations can add the fields, but they are only required on newer versions of the spec. This would be the less "breaking" approach, but we can have packages with older versions of the spec and without these fields, what would need to be handled by agentless code.
* We make the breaking change, so these fields are required. Then when updating the package-spec in the integrations repository, we need to add the fields where they are missing. This option is more "breaking", but probably better in the long term, but we need to know what fields to use when updating the package-spec.

Regarding default fields, we can add a default to the spec as documentation, but its implementation depends on whatever reads these settings, defaults are not enforced by the spec.

spec/changelog.yml
```diff
@@ -4,461 +4,466 @@
...
   changes:
-  - description: Add support for `slo` assets.
-    type: enhancement
-    link: https://github.com/elastic/package-spec/pull/767
+    - description: Add support for `slo` assets.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/767
+- version: 3.2.3-next
+  changes:
+    - description: Add support for identity of agentless deployment mode.
```
jsoriano, line 17: Nit. I would add that this identity is required, so it is more clear why this can be breaking.
```suggestion
    - description: Add support for identity of agentless deployment mode, and make it mandatory when this mode is enabled.
```

spec/changelog.yml
jsoriano: Yes, this is enough.

jsoriano (approved):
LGTM if we are fine with the path of making this a breaking change (see https://github.com/elastic/package-spec/pull/801#discussion_r1763051904).

spec/integration/manifest.spec.yml
seanrathier: @jsoriano, I had an offline conversation with @eyalkraft, we decided to change all 3 integrations that have or will have the `depoyment_mode` attribute.

test/packages/bad_deployment_mode_without_identities/manifest.yml
```diff
@@ -0,0 +1,49 @@
+format_version: 3.1.4
+name: bad_deployment_mode
```
jsoriano, line 2: For the package to be used in tests it must be added to the list of packages here: https://github.com/elastic/package-spec/blob/5f2c9ee5976c4833d3fe9791e5a616b369151fae/code/go/pkg/validator/validator_test.go#L28

test/packages/bad_deployment_mode_without_identities/manifest.yml
```diff
@@ -0,0 +1,49 @@
+format_version: 3.1.4
```
jsoriano, line 1:
```suggestion
format_version: 3.2.2
```
This should fix this error:
```
Additional property default is not allowed
```

code/go/pkg/validator/validator_test.go
```diff
@@ -203,7 +203,14 @@ func TestValidateFile(t *testing.T) {
 			"manifest.yml",
 			[]string{
 				`field policy_templates.0.deployment_modes: Additional property default is not allowed`,
-				`field policy_templates.0.inputs.0.vars.0.hide_in_deployment_modes.0: policy_templates.0.inputs.0.vars.0.hide_in_deployment_modes.0 must be one of the following: "agentless"`,
```
jsoriano, line 206: No need to remove this line, right?

seanrathier: Aaah 💩 ,   I originally changed the wrong test.

jsoriano (approved):

Issue #693, opened 2024-01-18
[Change Proposal] Support Knowledge Base Packages


spong:
### Summary

This is a proposal much like #346 or #351 for enabling the ability to bundle static data stream data within a package, such that when the package is installed, the data stream is created, and the bundled data is ingested into the data stream.

The specific use case here is for shipping 'Knowledge Base' content for use by the Elastic Assistants. For example, both [Security](https://github.com/elastic/kibana/pull/169593) and [Observability](https://github.com/elastic/kibana/pull/174583) Assistants are currently bundling our ES|QL docs with the Kibana distribution for each release. We then take this data, optionally chunk it, and then embed/ingest it using ELSER into a 'knowledge base' data stream so the assistants can query it for their ES|QL query generation features. Each release we'll need to update this content, and ship it as part of the Kibana distribution, with no ability to ship intermediate content updates outside of the Kibana release cycle.

Additionally, as mentioned in https://github.com/elastic/package-spec/issues/346#issuecomment-1890186256, this essentially provides us the ability to ship 'Custom GPTs' that can integrate with our assistants, and so opens up a world of possibilities for users to configure and expand the capabilities of the Security and Observability Assistants.

#### Configuration 

The core requirement here is for the ability to include the following when creating a package:

* Any number of data streams to create, though realistically one is probably sufficient
* An arbitrary number of documents, perhaps in json format, or zipped as detailed in #346 
  *  This generally won't be large amounts of data as detailed in #346 (our ES|QL docs are 196 documents and ~125KB), however I would expect some users would push this to enable RAG over larger data sets
* Some configuration for the destination data stream of the bundled documents. If we include a raw dump of the documents from ES, perhaps we can use just the `_index` fields to route them accordingly?

#### Behavior

Upon installation, the package should install the included data streams, then ingest the bundled documents into their destination data stream. This initial data should stick around for as long as the package is installed. If the package is removed, the data stream + initial data should be removed as well. When the package is updated, it would be fine to wipe the data stream/initial data and treat it as a fresh install. Whatever is easiest/most resilient would be fine for the first iteration here. No need to worry about appending new data on upgrade, or dealing with mapping changes, just delete the data streams and re-install/re-ingest the initial data.

---

The above would be sufficient enough for us to start bundling knowledge base documents in packages, at which point we could install as needed in support of specific assistant features.

spong:
[...]

PR #807, opened 2024-09-25
Add support for Kibana knowledge base entry assets

pgayvallet:
## What does this PR do?

Add support for Kibana knowledge base entries, as discussed in https://github.com/elastic/package-spec/issues/693

Knowledge base entries are a new Kibana type of content, living under `kibana/knowledge_base_entry/*`.
A KB entry is composed of 3 parts:
- a `manifest.yml` file
- a `fields` folder, containing one or more fields file
- a `content` folder, containing one or more content file, containing the ES documents of the entry in an NDJSON format.

Please refer to the changes and the self-review for more details.

## Why is it important?

As it is required for https://github.com/elastic/kibana/issues/192031 via https://github.com/elastic/kibana/issues/193849

## Checklist

## Related issues

- Closes https://github.com/elastic/package-spec/issues/693
- Related to https://github.com/elastic/kibana/issues/193849


spec/content/kibana/spec.yml
```diff
@@ -29,3 +29,8 @@ spec:
       pattern: '^{PACKAGE_NAME}-.+\.json$'
       forbiddenPatterns:
         - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
+  - description: Folder containing Kibana knowledge base entries
+    type: folder
+    name: kb_entry
+    required: false
+    $ref: "../../integration/kibana/kb_entry/spec.yml"
```
pgayvallet, line 36: So, I decided to make knowledge base entries (first-class) citizens of the `kibana` spec instead of having them live in their own, dedicated root folder. 

The reasoning is that:
1. KB entries will be "managed" by Kibana 
1. KB entries will only be used via Kibana (at least for now)

So I felt like it made more sense to have it done that way.

spec/integration/kibana/kb_entry/spec.yml
```diff
@@ -0,0 +1,27 @@
...
+          sizeLimit: 2MB
+          name: "manifest.yml"
+          required: true
+          $ref: "./manifest.spec.yml"
+        - description: A knowledge base entry's index mapping
+          type: file
+          contentMediaType: "application/json"
+          sizeLimit: 5MB
+          name: "index-mapping.json"
+          required: true
```
pgayvallet, line 23: I decided to KISS and keep the index's mapping as a plain `json` file, first because I don't think adding validation on the index's mapping format or having it as a yaml file would bring much value (AFAIK we're not doing it for index templates, mappings are just defined as `object` type with allowed extra props), and I also did not use the `fields.yml` feature because we don't need, or even want, any kind of discoverability on this type of content. 

Please tell me if you think this doesn't make sense.

spec/integration/kibana/kb_entry/spec.yml
```diff
@@ -0,0 +1,27 @@
...
+        - description: A knowledge base entry's content file
+          type: file
+          contentMediaType: "application/json"
+          pattern: '^content(\.[0-9]+)?\.json$'
```
pgayvallet, line 27: As suggested by @jsoriano, we're allowing an arbitrary number of file following the fixed `content.json` and 
 dynamic `content.{num}.json` naming format. Each file have a single `documents` property containing the source of the documents we will be indexing into the KB's index.

I did not add any validation on the file's content format, both because I wasn't sure we can add validation on JSON documents, and also because we wouldn't have much to validation (only thing we could do it make sure the only top level property present is `documents` and that it's an array).

But if there is a way to perform validation on JSON docs with the spec validator, I can absolutely add it.

test/packages/knowledge_base/kibana/kb_entry/foo/manifest.yml
```diff
@@ -0,0 +1,14 @@
...
+  name: .kibana_kb_foo
+  system: true
+retrieval:
+  syntactic_fields:
+    - content_title
+  semantic_fields:
+    - content_body
+    - ai_subtitle
+    - ai_summary
+    - ai_questions_answered
```
pgayvallet, line 14: Example file of the format specified in `spec/integration/kibana/kb_entry/manifest.spec.yml`.

Most of this manifest is meta that will be used by the KB implementation: 

- `name` and `description` are meta that will be surfaced in the KB management page. `name` will be used at the identifier.
- `retrieval.syntactic_fields` and `retrieval.semantic_fields` are the list of fields that will be used for retrieving documents from this KB entry. Later we could find way to be smarter and more automated about field detection, but I wanted to start with something both easy and concrete, so the KB entry contains those meta. 

- `index.name` is self explanatory - the name of the index that will be created with the specified mappings from the `index-mapping.json` file
- `index.system` is just some future-proofing by exposing the info about the index being Kibana-managed or not (probably don't need more details on this specific PR)

pgayvallet (commented):
self-review explaining my decisions and the reasoning

test/packages/bad_knowledge_base/kibana/kb_entry/foo/content.0.json
```diff
@@ -0,0 +1,106 @@
+{
+  "documents": [
```
jsoriano, line 2: I wonder if this file could be just a ndjson file.

Is the `documents` entry needed? Do we expect other entries in this file?

How are these files generated? Do we expect manual editions on these files?

test/packages/knowledge_base/kibana/kb_entry/foo/index-mapping.json
```diff
@@ -0,0 +1,42 @@
+{
+  "dynamic": "strict",
```
jsoriano, line 2: Is this setting going to be the same in all knowledge bases?

test/packages/knowledge_base/kibana/kb_entry/foo/index-mapping.json
```diff
@@ -0,0 +1,42 @@
...
+  "properties": {
```
jsoriano, line 3: Are these properties going to be the same in all knowledge bases?

test/packages/knowledge_base/kibana/kb_entry/foo/index-mapping.json
```diff
@@ -0,0 +1,42 @@
+{
```
jsoriano, line 1: If all knowledge bases are expected to have the same mappings, I think the mapping could be provided by Fleet. We should include in packages only things that are different between them.

test/packages/knowledge_base/kibana/kb_entry/foo/manifest.yml
jsoriano: > * retrieval.syntactic_fields and retrieval.semantic_fields are the list of fields that will be used for retrieving documents from this KB entry. Later we could find way to be smarter and more automated about field detection, but I wanted to start with something both easy and concrete, so the KB entry contains those meta.

How are these going to be different between different knowledge bases?

> * `index.name` is self explanatory - the name of the index that will be created with the specified mappings from the `index-mapping.json` file

The index name, or anything that could conflict with other resources, should be managed by Fleet. It could be built from the package and entry names.

jsoriano (commented):
Thanks for the PR, it is helpful to see the purpose of the new feature.

Overall it looks good to me, I have added some small suggestions.

My main question at the moment would be about how/if different knowledge bases differ on their mappings. If we expect all the knowledge bases to have the same mappings, I think that the mappings should be provided by Fleet, and not by the packages. Having them in the packages can be error prone. If this mapping is only expected for knowledge bases containing documentation, I think that we could still add some kind of "type" field, that tells Fleet what mappings to install. And we could also have a mix, where Fleet provides a base set of common mappings, and additional custom mappings can still be provided by packages.

spec/content/kibana/spec.yml
pgayvallet: > Do you foresee other uses of this data?

We **might** want at some point to expose this data more directly to the end user, and make is less of an implementation detail. But we don't have any use case for this right now, so I think we should just ignore it for now and eventually revisit later.

test/packages/knowledge_base/kibana/kb_entry/foo/manifest.yml
pgayvallet: > How are these going to be different between different knowledge bases?

By having those defined in the manifest, and then registered during installation. The search/retrieval code will then forge its request to target the specified fields.

> The index name, or anything that could conflict with other resources, should be managed by Fleet 

I guess that makes sense

spec/content/kibana/spec.yml
```diff
@@ -29,3 +29,8 @@ spec:
       pattern: '^{PACKAGE_NAME}-.+\.json$'
       forbiddenPatterns:
         - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
+  - description: Folder containing Kibana knowledge base entries
+    type: folder
+    name: kb_entry
```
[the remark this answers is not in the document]
pgayvallet: That's totally fair. `knowledge_base_entry` sounds good, I will adapt accordingly

test/packages/bad_knowledge_base/kibana/kb_entry/foo/content.0.json
pgayvallet: I'm being (probably overly) cautious with NDJSON, as we've learned the hard way with Kibana import/export that making the format evolve can be a pain (very difficult to add global metadata in addition to the entries). But for Kibana import/export, we're only generating a single file, we don't have a full folder structure like we do there, so the limitations of the format were more impactful. 

So no, I don't foresee us needing to add anything to those files, so if you prefer using ndjson, I'm fine with it

test/packages/knowledge_base/kibana/kb_entry/foo/index-mapping.json
pgayvallet: Related to https://github.com/elastic/package-spec/pull/807#discussion_r1775216062 - they won't, we want to stay open to the format. As long as we know which fields are searchable (semantic/syntactic) our retrieval logic will manage.

test/packages/knowledge_base/kibana/kb_entry/foo/index-mapping.json
pgayvallet: Not *necessarily*, I mean given the mappings are defined per KB entry, I assume we would be fine letting this be the source of truth for the `dynamic`allocation too (given it's technically a part of the mapping, not a setting).

But having this "hardcoded" and non-configurable by the package would be fine too ihmo

spec/integration/kibana/kb_entry/spec.yml
pgayvallet: So an other part of the reasoning was that each knowledge base entry can, technically, have different mappings. I'm not sure how that would play with that `fields.yml` feature, given that from my understanding, this is only a top level section in the package?

Or should we use the same format but with a different pattern, like having one fields.yml file per knowledge base entry folder (that would then replace the `index-mappings.json` file)?

spec/integration/kibana/kb_entry/spec.yml
pgayvallet: > In other cases where we allow multiple files we allow to use letters too, so it can give meaningful names when this is used for organizational purposes, and we use to separate with hyphens.

Sounds good, I will follow the pattern for consistency. I'll just wait for the other question about the format (json vs ndjson) to be addressed to perform both changes at the same time if required.

spec/integration/kibana/kb_entry/spec.yml
jsoriano: We reuse the fields definitions on different places. The more direct way to add it here would be to have a `fields` directory here, something like this:
```
        - description: Folder containing field definitions to be used as the mappings for the index template
          type: folder
          name: fields
          required: false
          $ref: "../../data_stream/fields/spec.yml"
```

test/packages/bad_knowledge_base/kibana/kb_entry/foo/content.0.json
jsoriano: Not a strong opinion, but I think I'd prefer using ndjson here, yes. If there is some case were we need to add metadata in the future we can use the manifest file, or some other file.

test/packages/knowledge_base/kibana/kb_entry/foo/index-mapping.json
jsoriano: > But having this "hardcoded" and non-configurable by the package would be fine too ihmo, if we want to switch to the `fields.yml` approach

Yes, if we use the `fields.yml` approach this could be a setting managed by Fleet, hardcoded by now to `strict`. If needed for some future case we could add a setting to the knowledge base manifest to control the value.

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -590,12 +597,30 @@ spec:
...
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.4.0
+    patch:
+      - op: remove
+        path: "/items/properties/type/enum/35" #remove semantic_text type
+      - op: remove
+        path: "/items/allOf/9" # removing inference_id when type is semantic_text
+      - op: remove
+        path: "/items/properties/inference_id" # removing inference_id field
```
jsoriano, line 623: :+1:

spec/integration/kibana/spec.yml
```diff
@@ -135,8 +135,17 @@ spec:
       contentMediaType: "application/json"
       pattern: '^{PACKAGE_NAME}-.+\.json$'
       forbiddenPatterns:
-        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden 
+        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
+  - description: Folder containing Kibana knowledge base entries
+    type: folder
+    name: knowledge_base_entry
+    required: false
+    $ref: "./knowledge_base_entry/spec.yml"
```
jsoriano, line 143: This should be added only to content packages by now.

spec/integration/kibana/knowledge_base_entry/manifest.spec.yml
```diff
@@ -0,0 +1,43 @@
+##
+## Describes the specification for a Kibana knowledge base entry's manifest.yml file
+##
+spec:
+  type: object
+  additionalProperties: false
+  properties:
+    name:
+      description: The name of the knowledge base entry
+      type: string
```
jsoriano, line 10: Is `name` needed? Fleet could use the package and knowledge base directory names to build a unique one.

spec/integration/kibana/knowledge_base_entry/manifest.spec.yml
```diff
@@ -0,0 +1,43 @@
...
+    retrieval:
+      type: object
+      additionalProperties: false
+      properties:
+        syntactic_fields:
+          description: List of fields that should be used for syntactic search during retrieval.
+          type: array
+          items:
+            type: string
+        semantic_fields:
```
jsoriano, line 32: Where are these syntatic and semantic fields used? I wonder if we should add this to the descriptions.

spec/integration/kibana/knowledge_base_entry/spec.yml
```diff
@@ -0,0 +1,33 @@
+spec:
```
jsoriano, line 1: Please move this file under `spec/content` by now.

test/packages/knowledge_base/manifest.yml
```diff
@@ -0,0 +1,27 @@
...
+  This package is a dummy example for packages containing a knowledge base entry.
+version: 0.1.0
+type: content
+source:
+  license: "Elastic-2.0"
+conditions:
+  kibana:
+    version: '^8.15.0'
+  elastic:
+    subscription: 'enterprise'
```
jsoriano, line 14: Do knowledge bases require enterprise subscription? (out of curiosity, no need to do anything about this here)

test/packages/knowledge_base/manifest.yml
```diff
@@ -0,0 +1,27 @@
...
+title: Knowledge base test content package
+description: >
+  This package is a dummy example for packages containing a knowledge base entry.
+version: 0.1.0
+type: content
+source:
+  license: "Elastic-2.0"
+conditions:
+  kibana:
+    version: '^8.15.0'
```
jsoriano, line 12: Are knowledge bases going to be available in all kinds of deployments? (in all serverless project types, on-prem and cloud)
Wondering in case we need to add a [capability](https://github.com/elastic/package-spec/blob/d9888bd70ae1d344a98ffc1dae08eccc1f90ad63/spec/integration/manifest.spec.yml#L117), we might need to add one if this feature is not going to be available in all kinds of deployments.

jsoriano (changes requested):
Looks good, the main change I would like to see in current code is limiting the changes to content packages and don't do changes to integration packages.

spec/integration/kibana/knowledge_base_entry/manifest.spec.yml
pgayvallet: I was planning on autogenerating an "id" based on the package name and KB directory, but the manifest's `name` is more of a meta data field for a user-readable "name", which will be displayed in the KB registry's UI.

spec/integration/kibana/knowledge_base_entry/manifest.spec.yml
pgayvallet: These are metadata that will be registered to the KB registry during the package installation, so that the KB knows which field to use to generate queries for retrieval.

test/packages/knowledge_base/manifest.yml
pgayvallet: Yeah, those features are gated behind enterprise subscription for now (mostly because it's part of the AI assistant feature, which is also gated behind enterprise)

test/packages/knowledge_base/manifest.yml
pgayvallet: Yeah, they will be available on all kinds of deployments

spec/integration/kibana/knowledge_base_entry/manifest.spec.yml
jsoriano: Ah ok, makes sense. Maybe then we could call it `label` or `title`? But as you prefer, name in that context sounds also good.

pgayvallet: I think you're right, `title` is probably more explicit and avoid mixing the concepts. Will update

test/packages/knowledge_base/kibana/knowledge_base_entry/foo/manifest.yml
```diff
@@ -0,0 +1,11 @@
+title: foo
```
jsoriano, line 1: Thinking about the issue with the assumptions in the kibana directory, and if we have to move the knowledge bases out of it, maybe we can revisit the idea of making this part of a more generic feature for data.

It could be done by:
* Moving knowledge base from `./kibana/knowledge_base_entry/foo` to `./index_data/foo`.
* Adding a `type` field in the manifest indicating what this data is about.
```suggestion
title: foo
type: knowledge_base
```

Thinking on other features that might provide data in the future, we would probably follow the same structure in all cases: the actual content, the field mappings, and some manifest with metadata or additional options. The `type` would instruct Fleet about different steps done during installation for each kind of data.

jsoriano (commented):
Looks good, my only concern now would be if we should get this out of the kibana repository.

test/packages/knowledge_base/kibana/knowledge_base_entry/foo/manifest.yml
pgayvallet: 👍, reworked in in https://github.com/elastic/package-spec/pull/807/commits/cdb409846c86dc36eecb29a3c89591e50f632c40

spec/content/index_data/manifest.spec.yml
```diff
@@ -0,0 +1,69 @@
...
+    retrieval: true
+  allOf:
+    - if:
+        properties:
+          type:
+            const: knowledge_base_entry
+        required:
+          - type
+      then:
+        properties:
```
pgayvallet, line 30: This is the part I'm the less sure about. Are we fine with all subtype's fields "leaking" at top level, or would we rather have dedicated groups per subtype, e.g.

```
knowledge_base_entry:
   type: object
   properties: 
       [...]
```

spec/content/spec.yml
```diff
@@ -46,6 +46,11 @@ spec:
     name: kibana
     required: false
     $ref: "./kibana/spec.yml"
+  - description: Folder containing Index data assets used by the package
+    type: folder
+    name: index_data
+    required: false
+    $ref: "./index_data/spec.yml"
```
jsoriano, line 53: @mrodm we are adding a new "index_data" directory at the root level, for collections of documents that should be installed on package installation. By now the only kind of collection we support would be "knowledge bases" for AI assistants. Other collections we could add in the future could be sample data for example.

Do you foresee any problem with this approach or this new directory at the root level?

jsoriano (approved):
Thanks!

LGTM, pinging Mario in case he sees any last minute problem with this approach.

mrodm (approved):
LGTM!

As a note related to this, the current issue in Kibana https://github.com/elastic/kibana/issues/192484 will not considerate this new folder or subfolders (i.e. fields) when addign support for content packages @jsoriano @pgayvallet

spec/content/index_data/manifest.spec.yml
```diff
@@ -0,0 +1,69 @@
+##
```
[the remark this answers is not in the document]
jsoriano: Thanks for checking!

spec/content/spec.yml
jsoriano: Yeah, we would only need a change in the registry if this were going to be used for discovery of packages. Let's see how the support in Kibana progresses.

jsoriano:
> As a note related to this, the current issue in Kibana [elastic/kibana#192484](https://github.com/elastic/kibana/issues/192484) will not considerate this new folder or subfolders (i.e. fields) when addign support for content packages @jsoriano @pgayvallet

No problem, Pierre already created an issue and started a PR for this:
* https://github.com/elastic/kibana/issues/193849
* https://github.com/elastic/kibana/pull/194209

PR #808, opened 2024-10-01
New changelog versions validation check

mrodm:
## What does this PR do?

Add a new validation check to ensure that the first (newer) changelog entry defined in the changelog file is not lower than the other versions.

Example of PR releasing an old version: https://github.com/elastic/integrations/pull/10800

## Why is it important?

Ensure that it is not released a older version. Releasing an older version should be performed through another mechanism.

## Checklist

## Related issues

-

### DIFF
--- a/code/go/internal/validator/semantic/validate_version_integrity.go
+++ b/code/go/internal/validator/semantic/validate_version_integrity.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
 	"github.com/elastic/package-spec/v3/code/go/internal/pkgpath"
 	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
@@ -33,6 +34,11 @@ func ValidateVersionIntegrity(fsys fspath.FS) specerrors.ValidationErrors {
 	}
 
 	err = ensureManifestVersionHasChangelogEntry(manifestVersion, changelogVersions)
+	if err != nil {
+		return specerrors.ValidationErrors{specerrors.NewStructuredError(err, specerrors.UnassignedCode)}
+	}
+
+	err = ensureChangelogLatestVersionIsGreaterThanOthers(changelogVersions)
 	if err != nil {
 		return specerrors.ValidationErrors{specerrors.NewStructuredError(err, specerrors.UnassignedCode)}
 	}
@@ -129,4 +135,25 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
 		}
 	}
 	return errors.New("current manifest version doesn't have changelog entry")
+}
+
+func ensureChangelogLatestVersionIsGreaterThanOthers(versions []string) error {
+	latestVersion, err := semver.NewVersion(versions[0])
+	if err != nil {
+		return fmt.Errorf("could not read package manifest version [%s]: %w", versions[0], err)
+	}
+
+	for i, v := range versions {
+		if i == 0 {
+			continue
+		}
+		changelogVersion, err := semver.NewVersion(v)
+		if err != nil {
+			return fmt.Errorf("could not read package manifest version [%s]: %w", changelogVersion, err)
+		}
+		if changelogVersion.GreaterThanEqual(latestVersion) {
+			return fmt.Errorf("changelog entry %s is greater than or equal to first changelog entry: %s", changelogVersion, latestVersion)
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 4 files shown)


mrodm:
test integrations

code/go/internal/validator/semantic/validate_version_integrity_test.go
```diff
@@ -5,103 +5,173 @@
...
-			"https://github.com/elastic/integrations/pull",
-			errGithubIssue,
+			"changelog next version manifest older version",
+			"1.0.0",
+			[]string{
+				"1.1.0-next",
+				"1.0.1",
+				"1.0.0",
+			},
+			false,
```
mrodm, line 93: This test case shows the current behaviour when using next versions in changelog (but not in manifest).
Here, the manifest version `1.0.0` is not the latest stable `1.0.1`.
Should this be allowed? Or, should it be revisited this code? @jsoriano

jsoriano: I think this would be a different change, and not sure if we want to be strict on that. We can keep it as is now.

code/go/internal/validator/semantic/validate_version_integrity.go
```diff
@@ -130,3 +136,24 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
 	}
 	return errors.New("current manifest version doesn't have changelog entry")
 }
+
+func ensureChangelogLatestVersionIsGreaterThanOthers(versions []string) error {
+	latestVersion, err := semver.NewVersion(versions[0])
```
jsoriano, line 141: Just in case, we should check if there is any version in `versions`.

code/go/internal/validator/semantic/validate_version_integrity.go
```diff
@@ -130,3 +136,24 @@ func ensureManifestVersionHasChangelogEntry(manifestVersion string, versions []s
...
+	if err != nil {
+		return fmt.Errorf("could not read package manifest version [%s]: %w", versions[0], err)
+	}
+
+	for i, v := range versions {
+		if i == 0 {
+			continue
+		}
```
jsoriano, line 149: Nit.
```suggestion
	for i, v := range versions[0:] {
```

jsoriano (approved):
LGTM, only a couple of nits

code/go/internal/validator/semantic/validate_version_integrity_test.go
mrodm: Ok, let's keep it as is :+1:

code/go/internal/validator/semantic/validate_version_integrity.go
mrodm: :+1: Applied change here: https://github.com/elastic/package-spec/pull/808/commits/02800e4f536cce2134c0f8a17d15bae3ef20ca9f

mrodm:
Just as a note, this new validation rule will affect all packages independently of their package spec version or package type @jsoriano

mrodm:
test integrations

mrodm:
Currently these are the packages failing due to package-spec related errors:
- aws: new version check
- network_traffic: new version check
- elastic_connectors: missing new fields

jsoriano:
There is already a PR open to fix AWS changelog https://github.com/elastic/integrations/pull/11307

mrodm:
> > * aws: new version check
> > * network_traffic: new version check
> 
> I guess these can be solved by fixing the order in the changelog? I think it would be fine to do that when we update elastic-package in integrations.

In case for instance of network_traffic, it has this version as the least 1.31.1, and the second one is 1.32.0. If it is re-ordered, 1.32.0 would have less changes/"code" than 1.31.1 version.

Another option would be to release a new version 1.32.1 with the same contents as 1.31.1. Not sure what it could be set in the changelog entry for this. Probably the same data as in 1.31.1?

jsoriano:
> Another option would be to release a new version 1.32.1 with the same contents as 1.31.1. Not sure what it could be set in the changelog entry for this. Probably the same data as in 1.31.1?

Yes, actually 1.32.0 will be "reverting" the change in 1.31.1. 
This was actually asked after merging. https://github.com/elastic/integrations/pull/10800#issuecomment-2325463137

The changelog could be fixed to something like this:
```
- version: "1.32.1"
  changes:
    - description: Add `event.module` to datastreams
      type: bugfix
      link: https://github.com/elastic/integrations/pull/10800
- version: "1.32.0"
  changes:
    - description: Removed `event.module` from datastreams
      type: breaking-change
      link: https://github.com/elastic/integrations/pull/????
- version: "1.31.1"
  changes:
    - description: Add `event.module` to datastreams
      type: bugfix
      link: https://github.com/elastic/integrations/pull/10800
    - description: Set `map_to_ecs` to enabled by default
      type: enhancement
      link: https://github.com/elastic/integrations/pull/10785
```

jsoriano:
PR open with the change I am propossing for network traffic: https://github.com/elastic/integrations/pull/11322

jsoriano (approved):

PR #825, opened 2024-10-24
Add setting to manage policy templates behaviour in Kibana

### LINKED ISSUE #802: [Change Proposal] Add flag to hide "parent tile" in integrations UI for integrations with multiple policy templates
For integrations like `elastic_connectors`, usage of of `policy_templates` is more about easing maintenance burden and less about reusing variables across different templates or leveraging any kind of information architecture. For integrations like this, we should have an option/flag in the manifest that allows Fleet to hide the "parent tile" for the integration, instead _only_ showing the policy templates cards.

mrodm:
## What does this PR do?

Adds a new setting `policy_templates_behaviour` to manage how the policy templates / tiles are shown to the user in Kibana.

## Why is it important?

There are some integrations that do not require to show a tile for the integration itself, or viceversa, there could be integrations that it is not desired to show each of the policy templates (child policy templates) defined as tiles, just the one refering to the integration itself (parent tile)

## Checklist

## Related issues

- Closes #802

### DIFF
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -205,7 +205,6 @@ spec:
                   - organization
                   - division
                   - team
-
     icons:
       description: List of icons for by this package.
       type: array
@@ -363,6 +362,20 @@ spec:
       $ref: "#/definitions/categories"
     conditions:
       $ref: "#/definitions/conditions"
+    # requires a conditional JSON schema to update the value depending
+    # on the policy_templates length
+    policy_templates_behaviour:
+      description: >
+        Behaviour to manage the policy templates (policy_templates field) defined that must be available in Kibana as tiles.
+        This is just applicable for those packages that define more than one policy template. By default for these packages, Kibana
+        shows all the policy templates defined in the manifest separately plus another one refering to the
+        whole integraton (that contains all the policy templates definitions) available to the user.
+
+        If the `integration_package_only` option is selected, then just one policy template for the whole integration containing all policy
+        templates will be available in Kibana. If the `policy_templates_only` option is selected, then just the policy templates defined
+        in the manifest will be available in Kibana without the specific policy template for the whole integration. If the `default` option
+        is selected, then Kibana keeps the same behaviour described above.
+      type: string
     policy_templates:
       description: List of policy templates offered by this package.
       type: array
@@ -487,6 +500,25 @@ spec:
     - version
     - type
     - owner
+  allOf:
+    - if:
+        properties:
+          policy_templates:
+            maxItems: 1
+      then:
+        properties:
+          policy_templates_behaviour:
+            enum:
+              - default
+            default: default
+      else:
+        properties:
+          policy_templates_behaviour:
+            enum:
+              - integration_package_only
+              - policy_templates_only
+              - default
+            default: default
 
 # JSON patches for newer versions should be placed on top
 versions:
--- a/test/packages/good_v3/manifest.yml
+++ b/test/packages/good_v3/manifest.yml
@@ -28,6 +28,7 @@ vars:
     title: A URL to get secrets, but not a secret itself.
     type: url
     show_user: true
+policy_templates_behaviour: policy_templates_only
 policy_templates:
   - name: apache
     title: Apache logs and metrics
@@ -68,7 +69,7 @@ policy_templates:
               - http://127.0.0.1
             hide_in_deployment_modes:
               - agentless
-  - name: apache
+  - name: apache-agentless
     title: Apache logs and metrics in agentless
     description: Collect logs and metrics from Apache instances in agentless
     deployment_modes:


spec/integration/manifest.spec.yml
mrodm, line 1: Added this new setting just for integrations.

Tried to look into kibana's code here: https://github.com/elastic/kibana/blob/e5b0b8108b392dae070b87175cf2d43522576c4d/x-pack/plugins/fleet/public/applications/integrations/sections/epm/screens/home/hooks/use_available_packages.tsx#L82

IIUC Kibana just shows different tiles in case of integrations packages. Input packages would just show one tile. Is my understanding correct?

Or, should it be added also for input packages ?

@jen-huang @kpollich

spec/integration/manifest.spec.yml
```diff
@@ -363,6 +362,20 @@ spec:
...
+        Behavior to manage the policy templates (policy_templates field) defined that must be available in Kibana as tiles.
+        This is just applicable for those packages that define more than one policy template. By default for these packages, Kibana
+        shows all the policy templates defined in the manifest separately plus another one referring to the
+        whole integration (that contains all the policy templates definitions) available to the user.
+
+        If the `integration_package_only` option is selected, then just one policy template for the whole integration containing all policy
+        templates will be available in Kibana. If the `policy_templates_only` option is selected, then just the policy templates defined
+        in the manifest will be available in Kibana without the specific policy template for the whole integration. If the `default` option
+        is selected, then Kibana keeps the same behavior described above.
+      type: string
```
mrodm, line 378: Not sure about the wording or even the naming for options here.
WDYT ? Any other suggestions?

spec/integration/manifest.spec.yml
```diff
@@ -363,6 +362,20 @@ spec:
...
+      description: >
+        Behavior to manage the policy templates (policy_templates field) defined that must be available in Kibana as tiles.
+        This is just applicable for those packages that define more than one policy template. By default for these packages, Kibana
+        shows all the policy templates defined in the manifest separately plus another one referring to the
+        whole integration (that contains all the policy templates definitions) available to the user.
+
+        If the `integration_package_only` option is selected, then just one policy template for the whole integration containing all policy
+        templates will be available in Kibana. If the `policy_templates_only` option is selected, then just the policy templates defined
+        in the manifest will be available in Kibana without the specific policy template for the whole integration. If the `default` option
+        is selected, then Kibana keeps the same behavior described above.
```
jsoriano, line 377: Simplified wording, removing references to the visual representation. Also proposing these three values:
* `combined_policy`: to indicate that there is a single policy template that combines them all.
* `individual_policies`: to indicate that policies are available individually, without a combined policy.
* `all`: the default value, where both the combined policy and the individual policies are available.

```suggestion
        Expected behavior when there are more than one policy template defined.
        When set to `combined_policy`, a single policy template is available, that 
        combines all the defined templates. When set to `individual_policies`, all
        policies are individually available, but there is no combined policy.
        The default value is `all`, where the combined policy template is available
        along with the individual policies.
```

spec/integration/manifest.spec.yml
jsoriano: Proposed rewording in the comment above.

spec/integration/manifest.spec.yml
```diff
@@ -487,6 +500,25 @@ spec:
...
+    - if:
+        properties:
+          policy_templates:
+            maxItems: 1
+      then:
+        properties:
+          policy_templates_behavior:
+            enum:
+              - default
+            default: default
```
jsoriano, line 513: Maybe we could even disable the option if there is a single policy template?

spec/integration/manifest.spec.yml
```diff
@@ -487,6 +500,25 @@ spec:
...
+      else:
+        properties:
+          policy_templates_behavior:
+            enum:
+              - integration_package_only
+              - policy_templates_only
+              - default
```
jsoriano, line 520: Options proposed above.
```suggestion
              - combined_policy
              - individual_policies
              - all
```

spec/integration/manifest.spec.yml
jsoriano: I wonder why we support multiple policy templates in input packages :thinking: What could be a use case? I think we don't have any input package with more than one. Maybe we should change the spec to be more explicit about this.

Some previous discussion about this here: https://github.com/elastic/package-spec/pull/325/files#r860573848

jsoriano (commented):
Codewise LGTM, added some suggestions about wording and the possible values for the setting.

spec/integration/manifest.spec.yml
mrodm: I tried to forbid/disable that option in case of a single policy template, with this conditional schema:
```yaml

  allOf:
    - if:
        properties:
          policy_templates:
            maxItems: 1
      then:
        # not need to define policy_templates_behaviour if there is just one policy template
        not:
          required:
            - policy_templates_behavior
```

But the error was not descriptive enough I think to help to debug the error:
> field (root): Must not be present

spec/integration/manifest.spec.yml
mrodm: At least, currently all input packages defined in the integrations repository have just one policy template.

Forcing input packages to have just one policy template could be done by adding `maxItems` setting (it could be done in this PR or in a separate PR):
```diff
--- spec/input/manifest.spec.yml
+++ spec/input/manifest.spec.yml
@@ -38,6 +38,7 @@ spec:
     policy_templates:
       description: List of policy templates offered by this package.
       type: array
+      maxItems: 1
       items:
         type: object
         additionalProperties: false
```

> What could be a use case?

I was thinking for instance in the "Custom logs" input package, would that make sense to add a new policy template there to read logs from another source? or would it be better to create a new input package?

spec/integration/manifest.spec.yml
mrodm: I like this proposal that does not mention any reference to tiles in Kibana.
And it also looks easier to understand what each option is aimed for.

Thanks! 

I'll do the required changes.

mrodm: Added in https://github.com/elastic/package-spec/pull/825/commits/b2ff18906289256c01d80e8e585175bf433e92d5

spec/integration/manifest.spec.yml
jen-huang: > IIUC Kibana just shows different tiles in case of integrations packages. Input packages would just show one tile. Is my understanding correct?

Kibana does not enforce the distinction between integration and input type packages when it comes to displaying the tiles, it only reads the number of `policy_templates`

jsoriano (approved):

spec/integration/manifest.spec.yml
mrodm: As it is not clear whether or not input packages should have a limitation in the number of policy templates, I would suggest not doing that for now.

Issue #832, opened 2024-11-18
[Change Proposal] Add support for "next steps" links to be referenced in Kibana


criamico:
There are some cases when an integration developer might want to prompt the user to navigate to another area of kibana. 

For instance, in the case of Agentless policy configuration, the developer might need to display some next steps links that link to the kibana areas where the additional configuration will happen, as shown in the screenshot:
![Image](https://github.com/user-attachments/assets/a8127b6d-4c62-4f47-8b01-208508a82e61)

To be able to do an integration developer should be able to enumerate "next step" options for each package/policy template in [kibana manifest](https://github.com/elastic/package-spec/blob/5e844901c7cf688f011aaeedb2a17cc0b38b6458/spec/integration/manifest.spec.yml#L4):

```
policy_templates:
  - name: foo
    ...
    links:
      - title: Findings
        type: post_install_step
        content: "Consolidated view of security and compliance issues identified within your environment"
        url: "kbn:/app/enterprise_search/content/connector/{{connector_id}}/overview"
      - title: Add more accounts
        type: action
        url: "kbn:..."
```

There is also a special case of next steps that will navigate the user to the Connectors UI in Kibana and will reference connector ids for this:
![Image](https://github.com/user-attachments/assets/0dca3976-3ca8-47e1-8d17-f16fb1cf7d19)

These Connector Ids are currently static for the lifetime of a connector instance, (as they are ES Index document `_id` values for the `.elastic-connectors-v1` index) and they are being added as configurable to the policy "advanced" configuration, so they'll show up [here](https://github.com/elastic/integrations/blob/c55e51d9c8cdc64e4e10ca9c176894a74855d5a4/packages/elastic_connectors/manifest.yml#L49-L55). About connectors, they could be defined in one of these two ways:

- In the package policy ( no changes to package-spec)
- in package-spec, with the above suggested change to policy templates links, adding a templated variable `{{connector_id}}` that references the `vars` content.

jsoriano:
> These Connector Ids are currently static for the lifetime of a connector instance, (as they are ES Index document `_id` values for the `.elastic-connectors-v1` index) and they are being added as configurable to the policy "advanced" configuration, so they'll show up [here](https://github.com/elastic/integrations/blob/c55e51d9c8cdc64e4e10ca9c176894a74855d5a4/packages/elastic_connectors/manifest.yml#L49-L55).

So they are variables in the policy, right? This looks good as would be the same templating used in agent policies.

[...]

PR #834, opened 2024-11-19
Adding support for links and connectors to be used by Kibana

criamico:
## What does this PR do?

Adding support for links to be displayed in Kibana.

## Why is it important?

This change has been requested to immediately support Agentless integrations that need to display "next steps" links in Kibana and it will be hepful for other developers as well.

## Checklist

## Related issues

https://github.com/elastic/package-spec/issues/832


spec/integration/manifest.spec.yml
```diff
@@ -410,6 +410,21 @@ spec:
                 - access
           deployment_modes:
             $ref: "#/definitions/deployment_modes"
+          links:
+            description: List of links related to policy template.
+            type: array
+            items:
+              type: object
+              properties:
```
mrodm, line 418: Probably it's interesting to add here `additionalProperties` to false, so just these entries are the ones to be allowed

```suggestion
              type: object
              additionalProperties: false
              properties:
```

spec/integration/manifest.spec.yml
```diff
@@ -410,6 +410,24 @@ spec:
                 - access
           deployment_modes:
             $ref: "#/definitions/deployment_modes"
+          links:
```
mrodm, line 413: I think some of the fields defined under `links` should be set as required to ensure all the information required by Kibana is present.


```yaml
      links:
        description: ...
        type: array
        minItems: 1
        items:
          type: object
          properties: 
            - ...
          required:
            - title
            - link
            - url
```
Not sure if it could be leave some as not required. At least in the examples, there is one entry without the `content` field.

spec/changelog.yml
```diff
@@ -10,6 +10,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add support for next steps links to kibana manifest.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/834
```
mrodm, line 15: @jsoriano do you think this feature should be just available starting in spec 3.4.0 ? These links would not be shown in Kibana until the support is added.

Would it be interesting to add a JSON patch to remove the new `links` field for versions before `3.4.0`?

test/packages/kibana_links/manifest.yml
```diff
@@ -0,0 +1,132 @@
...
+  - name: apache
+    title: Apache logs and metrics
+    description: Collect logs and metrics from Apache instances
+    deployment_modes:
+      agentless:
+        enabled: true
+        organization: elastic
+        division: observability
+        team: obs-infraobs-integrations
+    links:
```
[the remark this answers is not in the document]
mrodm: Usually, a new test package is added to test that errors are caught.

These entries could be added into the `good_v3` test package.
And probably, this test package could be used to check some possible errors (renaming it to `bad_kibana_links`), like:
- defining a `links` field without elements
- defining an entry in `links` without one of the required elements.

Once this is done, in order to be validated this new test package, this package should be added here with the expected errors: https://github.com/elastic/package-spec/blob/main/code/go/pkg/validator/validator_test.go#L236-L241

mrodm (commented):
Thanks @criamico !
Added a few comments and questions about this new `links` field

spec/integration/manifest.spec.yml
criamico: Yes I think content could be optional. It would be the description of the link.

spec/changelog.yml
jsoriano: > @jsoriano do you think this feature should be just available starting in spec 3.4.0 ? These links would not be shown in Kibana until the support is added.

Does it break anything in current versions of the stack?
If the only problem is that links will not be rendered, I would say to stay in 3.3.
If something breaks before adapting kibana, then we need to use 3.4.

> Would it be interesting to add a JSON patch to remove the new links field for versions before 3.4.0?

Yes, I would add this for the version we decide at the end.

spec/integration/manifest.spec.yml
```diff
@@ -410,6 +410,33 @@ spec:
...
+              additionalProperties: false
+              properties:
+                title:
+                  description: Link title
+                  type: string
+                url:
+                  description: Link url
+                  type: string
+                type:
+                  description: Type of link. `next_steps` for a regular link, `action` for a link button.
```
jsoriano, line 428: Maybe we should be more explicit about this, while keeping this agnostic of how this is implemented in the UI.
```suggestion
                  description: Type of link. `next_steps` for links to locations that can be relevant right after configuring the policy. `action` for actions that can be performed while the policy is in use.
```

spec/integration/manifest.spec.yml
```diff
@@ -410,6 +410,33 @@ spec:
...
+                  type: string
+                  enum:
+                    - action
+                    - next_steps
```
jsoriano, line 432: Nit. Maybe we should be consistent with the use of plural here.
```suggestion
                    - action
                    - next_step
```

spec/integration/manifest.spec.yml
```diff
@@ -410,6 +410,24 @@ spec:
...
+                content:
+                  description: Link content
+                  type: string
+                url:
+                  description: Link url
+                  type: string
```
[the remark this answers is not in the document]
criamico: @seanstory could you weigh in about the possible url formats to be used here?

spec/changelog.yml
criamico: > Does it break anything in current versions of the stack?
>If the only problem is that links will not be rendered, I would say to stay in 3.3.
>If something breaks before adapting kibana, then we need to use 3.4.

I think that it shouldn't break anything. The "worst" that can happen is that those links get ignored by Kibana, as it reads only those keys that we explicitly choose. So we could stay in 3.3 for this change.

test/packages/good_v3/manifest.yml
```diff
@@ -39,7 +39,11 @@ policy_templates:
         organization: elastic
         division: observability
         team: obs-infraobs-integrations
-
+    links:
+      - title: Security overview
+        type: next_step
+        content: "View security overview"
+        url: "/security/overview"
```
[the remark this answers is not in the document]
criamico: Thanks Mario, I was waiting to have the desired formats for the urls to create this "bad" package. I'll create some with your suggestions and once we are sure of the formats I'll add that as well.

spec/integration/manifest.spec.yml
```diff
@@ -410,6 +410,33 @@ spec:
                 - access
           deployment_modes:
             $ref: "#/definitions/deployment_modes"
+          links:
```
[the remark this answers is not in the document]
criamico: @mrodm I renamed `links` to `configuration_links` and moved it under the `definitions` field. It's now referenced both under inputs and policy templates.

test/packages/good_v3/manifest.yml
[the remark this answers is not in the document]
criamico: I created a `bad_configuration_links` package with your suggestions.

spec/input/manifest.spec.yml
```diff
@@ -69,6 +69,8 @@ spec:
               - Collect logs and metrics from Apache instances
           deployment_modes:
             $ref: "../integration/manifest.spec.yml#/definitions/deployment_modes"
+          configuration_links:
```
jsoriano, line 72: Nit. What is the reason to mark these as `configuration_links`? Is it because they are expected to appear around configuration steps? `links` could be used in the future for other things, like links to additional documentation and so on.

criamico: Yes there was another comments saying that `links` was too generic, but if you think it's better to leave it generic on purpose I can put it back.

test/packages/kibana_configuration_links/LICENSE.txt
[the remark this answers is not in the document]
criamico: I had added the "good" case to good_v3, but that test package has version 3.4.0, so when we decided to change the version to 3.3.1 that wasn't correct anymore. Should I remove this test package then? @jsoriano what do you suggest?

spec/input/manifest.spec.yml
jsoriano: Oh ok, looks good then, not a strong opinion from my side.

PR #844, opened 2024-12-12
Expand configuration_links to support internal/external links

criamico:
## What does this PR do?
Follow up to https://github.com/elastic/package-spec/pull/834

## Why is it important?
With https://github.com/elastic/package-spec/pull/834 I added support to configuration_links to kibana manifest. This PR expands upon the previous change setting explicit formats for the urls:

**External** links should have `url` field of type `http://...` or `https://..`
``` yml
      - title: Configure fleet
        url: "https://www.elastic.co/guide/en/fleet/master/fleet-overview.html"
        type: next_step
        content: "See docs"
```

**Internal** links should use the `kbn:/` prefix (similar to what it's done in Kiban dev tools) and of course the link should be a valid Kibana link in order to work . Example corresponding to `/app/integrations/browse`:
``` yml
      - title: Browse integrations
        url: "kbn:/app/integrations/browse"
        type: action
```

**NOTE** Kibana doesn't currently handle `configuration_links` field at all and just ignores this field, as I'm still working to the related changes. This change should be safe to merge.

## Checklist

## Related issues

Related to https://github.com/elastic/package-spec/issues/832

### DIFF
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -219,9 +219,7 @@ spec:
           url:
             description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:app/...` for links internal to Kibana.
             type: string
-            oneOf:
-              - type: string
-                pattern: '^(http(s)?://|kbn:/)'
+            pattern: '^(http(s)?://|kbn:/)'
           type:
             description: Type of link. `next_steps` for links to locations that can be relevant right after configuring the policy. `action` for actions that can be performed while the policy is in use.
             type: string
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -217,29 +217,26 @@ spec:
             description: Link title
             type: string
           url:
-            description: Link url
+            description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:app/...` for links internal to Kibana.
             type: string
+            oneOf:
+              - type: string
+                pattern: '^http(s)?://'
+              - type: string
+                pattern: '^kbn:'
           type:
             description: Type of link. `next_steps` for links to locations that can be relevant right after configuring the policy. `action` for actions that can be performed while the policy is in use.
             type: string
             enum:
               - action
               - next_step
-          appId:
-            description: The identifier of the Kibana app to navigate to. Set this field when `internal` is true.
-            type: string
           content:
             description: Link description
             type: string
-          internal:
-            description: Set to true for internal kibana links, false for external links. Defaults to false.
-            type: boolean
-            default: false
         required:
         - title
         - url
         - type
-        - internal
     icons:
       description: List of icons for by this package.
       type: array
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -221,9 +221,7 @@ spec:
             type: string
             oneOf:
               - type: string
-                pattern: '^http(s)?://'
-              - type: string
-                pattern: '^kbn:'
+                pattern: '^(http(s)?://|kbn:)'
           type:
             description: Type of link. `next_steps` for links to locations that can be relevant right after configuring the policy. `action` for actions that can be performed while the policy is in use.
             type: string
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -217,7 +217,7 @@ spec:
             description: Link title
             type: string
           url:
-            description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:app/...` for links internal to Kibana.
+            description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:/app/...` for links internal to Kibana.
             type: string
             pattern: '^(http(s)?://|kbn:/)'
           type:
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -221,7 +221,7 @@ spec:
             type: string
             oneOf:
               - type: string
-                pattern: '^(http(s)?://|kbn:)'
+                pattern: '^(http(s)?://|kbn:/)'
           type:
             description: Type of link. `next_steps` for links to locations that can be relevant right after configuring the policy. `action` for actions that can be performed while the policy is in use.
             type: string
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -225,13 +225,21 @@ spec:
             enum:
               - action
               - next_step
+          appId:
+            description: The identifier of the Kibana app to navigate to. Set this field when `internal` is true.
+            type: string
           content:
             description: Link description
             type: string
+          internal:
+            description: Set to true for internal kibana links, false for external links. Defaults to false.
+            type: boolean
+            default: false
         required:
         - title
         - url
         - type
+        - internal
     icons:
       description: List of icons for by this package.
       type: array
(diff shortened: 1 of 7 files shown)


test/packages/bad_configuration_links/manifest.yml
```diff
@@ -79,10 +79,22 @@ policy_templates:
...
       - title: Security overview
         type: next_step
+        internal: true
+      # bad format: missing field `internal`
+      - title: View Agents
+        url: "/agents"
+        type: next_step
+        content: "Check your agents in Fleet"
+        appId: "fleet"
+      # bad format: `internal: false` should match http or https format
```
criamico, line 93: I'm not sure how to enforce this rule. I found the `pattern` field but can I enforce it conditionally (based on `internal` field)?

spec/integration/manifest.spec.yml
```diff
@@ -225,13 +225,21 @@ spec:
             enum:
               - action
               - next_step
+          appId:
```
jsoriano, line 228: We use snake_case for attributes here.
```suggestion
          app_id:
```

spec/integration/manifest.spec.yml
```diff
@@ -225,13 +225,21 @@ spec:
...
             type: string
+          internal:
+            description: Set to true for internal kibana links, false for external links. Defaults to false.
+            type: boolean
+            default: false
         required:
         - title
         - url
         - type
+        - internal
```
jsoriano, line 242: Do we need to make it required? Maybe we can assume that this is an external link if internal is not set?

test/packages/good_v3/manifest.yml
```diff
@@ -40,10 +40,22 @@ policy_templates:
...
     configuration_links:
-      - title: Security overview
-        url: "/security/overview"
+      - title: View Agents
+        url: "/agents"
         type: next_step
-        content: "View security overview"
+        content: "Check your agents in Fleet"
+        appId: "fleet"
+        internal: true
```
jsoriano, line 48: Instead of adding the `internal` setting, could we use the urls prefixed by `kbn` that are used in many other places?
```
      - title: View Agents
        url: "kbn:/app/fleet/agents"
        type: next_step
        content: "Check your agents in Fleet"
```

test/packages/bad_configuration_links/manifest.yml
jsoriano: I think you would need to use `anyOf` or `oneOf`. Something like this:
```
          internal:
            description: Set to true for internal kibana links, false for external links. Defaults to false.
            type: boolean
            oneOf:
              - const: true
                pattern: "^/"
              - const: false
                pattern: "^https?://"
```

mrodm: I think it can be used `if - then` definitions here to set those conditions

I've tried to set these conditions in the manifest definition:
```diff
 $ git diff
diff --git spec/integration/manifest.spec.yml spec/integration/manifest.spec.yml
index cf39a4f..bd79026 100644
--- spec/integration/manifest.spec.yml
+++ spec/integration/manifest.spec.yml
@@ -235,6 +235,30 @@ spec:
             description: Set to true for internal kibana links, false for external links. Defaults to false.
             type: boolean
             default: false
+        allOf:
+          - if:
+              properties:
+                internal:
+                  const: true
+            then:
+              required:
+                - appId
+              properties:
+                url:
+                  type: string
+                  pattern: '^/'
+          - if:
+              properties:
+                internal:
+                  const: false
+            then:
+              not:
+                required:
+                  - appId
+              properties:
+                url:
+                  type: string
+                  pattern: '^http(s)?://'
         required:
         - title
         - url
```

Adding those definitions there are some errors, that probably are legit. 

These conditions enforce:
- a pattern for `url` depending on the `internal` value
- `appId` is set required or not depending on the `internal` value.
Is there any other condition that should be added ?

test/packages/good_v3/manifest.yml
jsoriano, line 48: Or, trying to do this a bit more generic, define something like Android intents, but then we need to know the possible intents beforehand.
```
      - title: View Agents
        intent: "agents_view"
        type: next_step
        content: "Check your agents in Fleet"
```

mrodm (commented):
Added a suggestion to achieve conditional rules in the JSON schema.

spec/integration/manifest.spec.yml
```diff
@@ -225,13 +225,21 @@ spec:
             enum:
               - action
               - next_step
+          appId:
+            description: The identifier of the Kibana app to navigate to. Set this field when `internal` is true.
```
jsoriano, line 229: How does a package developer know the app to navigate to? Do they need to get it from the URL?

spec/integration/manifest.spec.yml
criamico: yes I was unsure about this, I can remove since I added the default

spec/integration/manifest.spec.yml
criamico: yes it's basically the plugin, it's usually in the url

test/packages/good_v3/manifest.yml
criamico: I couldn't find any examples in package-spec, but maybe you refer to the Kibana dev tools?

We could use this format, it requires to do some more manipulation in kibana as the url needs to be split (it cannot be used directly by kibana this way) and it's also less explicit. However, reading your other comments I understand your concerns that this could be a bit too specific to Kibana internal way of handling urls.

test/packages/bad_configuration_links/manifest.yml
criamico: Thanks, I'm switching to use format `kbn:...` for internal links as per below [discussion](https://github.com/elastic/package-spec/pull/844#discussion_r1881798689). Example `kbn:app/integrations/browse`

test/packages/good_v3/manifest.yml
criamico: Updated the PR and the description based on this discussio.

jsoriano (dismissed):
I think I prefer this approach based on the url prefixes, thanks!

Added a suggestion about how patterns are defined.

spec/integration/manifest.spec.yml
```diff
@@ -221,9 +221,7 @@ spec:
             type: string
             oneOf:
               - type: string
-                pattern: '^http(s)?://'
-              - type: string
-                pattern: '^kbn:'
+                pattern: '^(http(s)?://|kbn:)'
```
jsoriano, line 224: Should we add then the slash also here?
```suggestion
                pattern: '^(http(s)?://|kbn:/)'
```

spec/integration/manifest.spec.yml
```diff
@@ -217,8 +217,11 @@ spec:
             description: Link title
             type: string
           url:
-            description: Link url
+            description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:app/...` for links internal to Kibana.
             type: string
+            oneOf:
+              - type: string
+                pattern: '^(http(s)?://|kbn:/)'
```
jsoriano, line 224: `oneOf` is not needed now, as there is a single option.
```suggestion
            type: string
            pattern: '^(http(s)?://|kbn:/)'
```

jsoriano (approved):
:+1:

spec/integration/manifest.spec.yml
```diff
@@ -217,8 +217,11 @@ spec:
             description: Link title
             type: string
           url:
-            description: Link url
+            description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:app/...` for links internal to Kibana.
```
mrodm, line 220: Nit.
```suggestion
            description: Link url. Format is `http://...` or `https://...` for external links,  `kbn:/app/...` for links internal to Kibana.
```

mrodm (approved):

PR #847, opened 2024-12-16
Teach pipeline validator about terminate processor

### LINKED ISSUE #846: [Change Proposal] Allow 'terminate' ingest processor
Support the [terminate](https://www.elastic.co/guide/en/elasticsearch/reference/8.16/terminate-processor.html) processor (added in 8.16.0) in Fleet integration ingest pipelines. This is necessary to implement https://github.com/elastic/integrations/issues/12083#issuecomment-2546875613 which replaces the `fail` processor with `terminate` for clearer error messages.

The area of the package-spec that needs enhanced is:

https://github.com/elastic/package-spec/blob/7eb29caa329d3adb3eeedfab756e716a63ce948d/spec/integration/elasticsearch/pipeline.spec.yml#L9-L62

efd6:
## What does this PR do?

Add terminate processor to list of valid processors.

## Why is it important?

Currently the processor cannot be used.

## Checklist

## Related issues

- Closes #846

### DIFF
--- a/spec/integration/elasticsearch/pipeline.spec.yml
+++ b/spec/integration/elasticsearch/pipeline.spec.yml
@@ -48,6 +48,7 @@ spec:
         # set_security_user: false # Applicable to packages?
         sort: { type: object }
         split: { type: object }
+        terminate: { type: object }
         trim: { type: object }
         uppercase: { type: object }
         urldecode: { type: object }
--- a/spec/integration/elasticsearch/pipeline.spec.yml
+++ b/spec/integration/elasticsearch/pipeline.spec.yml
@@ -116,3 +116,7 @@ versions:
         path: /definitions/processors/if # remove rename processor validation
       - op: remove
         path: /definitions/processors/then # remove rename processor validation
+  - before: 3.3.0
+    patch:
+      - op: remove
+        path: /definitions/processors/terminate # remove terminate processor validation
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -10,6 +10,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add `terminate` processor to list pipeline spec.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/847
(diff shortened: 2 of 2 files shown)


jsoriano (approved):

spec/integration/elasticsearch/pipeline.spec.yml
```diff
@@ -48,6 +48,7 @@ spec:
         # set_security_user: false # Applicable to packages?
         sort: { type: object }
         split: { type: object }
+        terminate: { type: object }
```
jsoriano, line 51: Btw, please add a patch below, so this processor is actually not available in older versions of the spec, something like this:
```
  - before: 3.3.0
    patch:
      - op: remove
        path: /definitions/processors/terminate # remove terminate processor validation
```

efd6:
@jsoriano It seems that that does not work, unless I have misunderstood your instructions.

spec/integration/elasticsearch/pipeline.spec.yml
```diff
@@ -115,3 +116,7 @@ versions:
         path: /definitions/processors/if # remove rename processor validation
       - op: remove
         path: /definitions/processors/then # remove rename processor validation
+  - before: 3.3.0
+    patch:
+      - op: remove
+        path: /definitions/processors/terminate # remove terminate processor validation
```
mrodm, line 122: This array item should be added as the first one (as it references a newer version in the `before` field).

And looking at the error:
```
failed to load schema for "integration/elasticsearch/pipeline.spec.yml": failed to apply patch: error in remove for path: '/definitions/processors/terminate': unable to remove nonexistent key: terminate: missing value
```
the path should be updated too.

I think it should be like this:

```
versions:
  - before: 3.3.0
    patch:
      - op: remove
        path: /definitions/processor/properties/terminate # remove terminate processor validation
  - before: 3.1.0
    patch:
      - op: remove
        path: /definitions/processors/if # remove rename processor validation
      - op: remove
        path: /definitions/processors/then # remove rename processor validation
  
```

jsoriano:
Continuing with this change in https://github.com/elastic/package-spec/pull/857.

Issue #744, opened 2024-05-03
[Change Proposal] Add support for required conditional groups of variables


jsoriano:
There are cases where there are sets of variables that are optional, but it is required to set at least one of them. This is frequent for example when there are different available authentication options.

For example in Cloudflare there are three variables defined as optional: `auth_token`, `auth_email` and `auth_key`. Either the token needs to be set, or the email and the key.

We need a way to express this in the package manifests.

There are some options to explore:

Option 1: Variable groups
---
So it is possible to define "groups" in a similar fashion to `select`, and the selected group can have its own restrictions, something like this:
```
      - name: auth
        type: select_vargroup
        required: true
        options:
          - name: key
            title: Authentication using email and password
            vars:
              - name: auth_email
                type: text
                title: Auth Email
                description: The Auth Email. Needs to be used with an Auth Key. Do not fill if you are using an Auth Token.
                multi: false
                required: true
                show_user: true
              - name: auth_key
                type: password
                title: Auth Key
                description: The Auth Key. Needs to be used with an Auth Email. Do not fill if you are using an Auth Token.
                multi: false
                required: true
                show_user: true
                secret: true
          - name: token
            title: Authentication using token
            vars:
              - name: auth_token
                type: password
                title: Auth token
                description: The auth token. If set, Auth Email and Auth Key will be ignored.
                required: true
                multi: false
                show_user: true
                secret: true
```

This would also require changes in Fleet UI.

Option 2: Groups of requirements
---
Define a list of groups that must be configured. Something like this would mean what is wanted in the Cloudflare integration:
```
required_vars:
  - [auth_email, auth_key]
  - [auth_token]
```
Understood as (auth_email AND auth_key) OR (auth_token).

This option wouldn't require breaking changes in packages. It will also require changes in Fleet UI.

criamico:
Kibana related PR: https://github.com/elastic/kibana/pull/182645

seanrathier:
- This would solve https://github.com/elastic/security-team/issues/10623

The proposal presented would help out with the CSPM integration.  The problem statement aligns with the CSPM challenge where we [need to have all the stream variables optional](https://github.com/elastic/integrations/blob/main/packages/cloud_security_posture/data_stream/findings/manifest.yml#L30) because we have the options`access_key` and `secret_key` need to be required together but `session_token` should be required by itself.  I think this case looks like **option 2**.

However,  I can see other scenarios getting more complex, so I am leaning on **option 1**

[...]

PR #855, opened 2025-01-14
[Cloud Security] Add support for required conditional groups of variables

seanrathier:
## What does this PR do?

Add support for conditionally required groups of variables.

## Why is it important?

In some cases, policy templates like CloudFlare and CSPM will have mutually exclusive input variables that are required.

For example,  access_key and secret_key or the session_token is required
```
required_vars:
    - [access_key,secret_key]
    - [session_token]
```

## Checklist

## Related issues
- Closes https://github.com/elastic/package-spec/issues/744
- Relates https://github.com/elastic/security-team/issues/10623


test/packages/bad_required_vars/manifest.yml
```diff
@@ -0,0 +1,47 @@
+format_version: 3.1.4
+name: bad_required_vars
```
jsoriano, line 2: Test packages need to be added to the list of packages in https://github.com/elastic/package-spec/blob/8f1292454250c378af0432606bb99418cc19c753/code/go/pkg/validator/validator_test.go#L28

spec/changelog.yml
```diff
@@ -4,6 +4,9 @@
 ##
 - version: 3.4.0-next
   changes:
+    - description: Add support for required conditional groups of variables.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/855
```
jsoriano, line 9: I think we could add this in 3.3.2. In versions of the stack that don't support this feature it will be likely ignored.

spec/integration/manifest.spec.yml
```diff
@@ -233,6 +233,14 @@ spec:
                   - organization
                   - division
                   - team
+    required_vars:
```
jsoriano, line 236: We should add a version patch below under `versions:`, it would be something like:
```
  - before: 3.3.2
    patch:
      - op: remove
        path: "/properties/policy_templates/items/properties/required_vars"
```

test/packages/good_v3/manifest.yml
```diff
@@ -93,6 +93,8 @@ policy_templates:
         organization: security
         division: engineering
         team: cloud-security
+    required_vars:
+      - [package_username,package_password]
```
jsoriano, line 97: It would be nice to validate that these variables actually exist. If this cannot be done in JSON schema we could add a validation in Go code, they are added around here: https://github.com/elastic/package-spec/blob/8f1292454250c378af0432606bb99418cc19c753/code/go/internal/validator/spec.go#L192

seanrathier: I agree,  I thought that also.

This will be a challenge because Go is quite new for me.  
I would like to test that this structure will work for Fleet to validate before investing in this

spec/integration/manifest.spec.yml
```diff
@@ -245,6 +245,14 @@ spec:
...
+      description: Required conditional variables for the package.
+      type: array
```
jsoriano, line 250: Should this array require at least two items? What would be the use of this list with a single list of variables?

test/packages/good_v3/manifest.yml
jsoriano: Ok, we can do this in a follow up.

jsoriano (dismissed):

test/packages/good_v3/manifest.yml
jsoriano: Follow up for this in https://github.com/elastic/package-spec/pull/856.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -32,6 +32,28 @@ spec:
             - 1
           - - array
             - 2
+    required_vars:
```
mrodm, line 35: This new field has not been added under "properties", so it is not used right now in data stream manifests.

As for instance, it is done with `vars`
https://github.com/elastic/package-spec/blob/8f1292454250c378af0432606bb99418cc19c753/spec/integration/data_stream/manifest.spec.yml#L524

spec/integration/manifest.spec.yml
```diff
@@ -515,6 +515,8 @@ spec:
                   description: Can input be defined multiple times
                   type: boolean
                   default: false
+                required_vars:
```
mrodm, line 518: Should this be added also into `input` packages ? @jsoriano @seanrathier

jsoriano: We could add there later if needed. @seanrathier do you need this for your current use case?

seanrathier: No,  I only need it in stream/inputs

jsoriano (approved):

PR #856, opened 2025-01-16
Add semantic validation for required variable groups

jsoriano:
## What does this PR do?

Add semantic validation for required variable groups added in https://github.com/elastic/package-spec/pull/855.

## Why is it important?

Adds some checks that can be useful to prevent mistakes when configuring required variable groups.

## Related issues

- Complementary to https://github.com/elastic/package-spec/pull/855.


code/go/internal/validator/semantic/validate_required_vargroups_test.go
```diff
@@ -0,0 +1,136 @@
...
+  - inputs:
+    - required_vars:
+        user_password:
+          - name: user
+          - name: password
+        api_key:
+          - name: api_key
+`,
+			errors: []string{
+				`required var "api_key" in optional group is defined as always required`,
```
mrodm, line 94: Why does it fail with error if variable api_key is defined ?

code/go/internal/validator/semantic/validate_required_vargroups.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+	"gopkg.in/yaml.v3"
+)
+
+// ValidateRequiredVarGroups validates lists of optional required variables.
+func ValidateRequiredVarGroups(fsys fspath.FS) specerrors.ValidationErrors {
+	d, err := fs.ReadFile(fsys, "manifest.yml")
+	if err != nil {
+		return specerrors.ValidationErrors{specerrors.NewStructuredErrorf("failed to read manifest: %w", err)}
+	}
```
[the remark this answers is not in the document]
jsoriano: Umm, yeah, not sure. Let's wait for the related discussion in the spec change. I remember some complexities merging these configurations in Fleet, we'd have to mimic the behavior here.

code/go/internal/validator/semantic/validate_required_vargroups_test.go
jsoriano: This check is to avoid conflicts between the definition of `required` at the var level, and its presence in an optional group in `required_vars`.
A field with `required: true` would be expected to be required in any case, while a field in a `required_var` groups would be only required depending on the other groups and variables.

The test fails here because `api_key` is both, defined as `required: true` and as a member of a `required_vars` optional group.

mrodm: Got it! Thanks!

seanrathier (approved):
Thanks for the follow up!  Much appreciated

code/go/internal/validator/semantic/validate_required_vargroups.go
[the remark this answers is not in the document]
jsoriano: Added support for data streams, not sure if it handles all the places where variables can be defined, but we can add more cases later if there is any. Let me know if you can think about some additional case.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -606,6 +606,10 @@ spec:
   - title
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.3.2
+    patch:
+      - op: remove
+        path: "/properties/streams/items/properties/required_vars"
```
mrodm, line 612: :+1:

mrodm (approved):
LGTM!

code/go/internal/validator/semantic/validate_required_vargroups.go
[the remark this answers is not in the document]
mrodm: > Added support for data streams, not sure if it handles all the places where variables can be defined, but we can add more cases later if there is any. Let me know if you can think about some additional case.

I don't recall any other additional case now, thanks for adding this support!

mrodm (approved):

seanrathier (approved):
Wow!  Thanks again.

PR #858, opened 2025-01-23
ci(buildkite): exclude files/folders that are not tested in Buildkite

v1v:
## What does this PR do?

Skip files that are not used in Buildkite

## Why is it important?

Run faster builds and avoid waste of CI cycles.

I decided to support the whole .github folder since there are some other files, such as CODEOWNERS or .dependabot.yml and the rest are not used by Buildkite

## Checklist

## Related issues

-


jsoriano (approved):
Thanks!

mrodm (approved):

PR #859, opened 2025-01-23
Prepare release 3.3.2

jsoriano:

PR #860, opened 2025-01-24
Extend synthetic source compliance test and test with 8.18

jsoriano:
## What does this PR do?

Enable compliance tests with 8.18.

In compliance tests for synthetic source, explicitly check that synthetic source is enabled apart from checking the field with `store: true`.

## Why is it important?

Test with latest 8.x development branch.

## Related issues

- Issues with this test discovered in https://github.com/elastic/package-spec/pull/859.

### DIFF
--- a/test/packages/logs_synthetic_mode/data_stream/synthetic/fields/some-fields.yml
+++ b/test/packages/logs_synthetic_mode/data_stream/synthetic/fields/some-fields.yml
@@ -1,3 +1,3 @@
 - name: decision_id
-  type: text
+  type: long
   store: true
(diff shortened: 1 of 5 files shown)


test/packages/logs_synthetic_mode/data_stream/synthetic/fields/some-fields.yml
```diff
@@ -1,3 +1,3 @@
 - name: decision_id
-  type: text
+  type: long
```
jsoriano, line 2: `store` is set to true by default for `text` fields since 8.18. Testing this setting with a `long` field now.

mrodm (approved):
:100:

PR #869, opened 2025-01-31
Disallow usage of commas in indices defined in transforms

mrodm:
## What does this PR do?

Disallow usage of commas to define indices in transform definition (`transform.yml`).

Example of definition disallowed:
```yaml
source:
  index: "metrics-elasticsearch.stack_monitoring.index*,.monitoring-es-8*,metricbeat-*"
```

This should be defined as:
```yaml
source:
  index:
  - "metrics-elasticsearch.stack_monitoring.index*"
  - ".monitoring-es-8*"
  - "metricbeat-*"
```

## Why is it important?

In Elastic stacks previous to 8.12.0, these definitions are not set correctly in Elasticsearch when the packages are installed.
For instance , in 8.10.1 the definition above would result in just one source index:

```json
  "source": {
    "index": [
      "metrics-elasticsearch.stack_monitoring.index*,.monitoring-es-8*,metricbeat-*"
    ],
```

Starting in 8.12.0, when using commas, these definitions look like this::
```json
  "source": {
    "index": [
      "metrics-elasticsearch.stack_monitoring.index*",
      ".monitoring-es-8*",
      "metricbeat-*"
    ],
```

This PR tries to enforce to define every index as a list (if more than one is needed).

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-package/issues/2341


code/go/pkg/validator/validator_test.go
```diff
@@ -290,25 +370,50 @@ func TestValidateFile(t *testing.T) {
...
+			if !ok {
+				require.Len(t, test, 1)
+				require.Len(t, test[0].expectedErrContains, 1)
+				require.Equal(t, errs.Error(), test[0].expectedErrContains[0])
+				return
+			}
+
+			require.Len(t, vErrs, expectedFailures)
+
+			for _, failedFile := range test {
```
jsoriano, line 397: Thanks for adding support for failures in multiple places, highly appreciated :tada:

spec/integration/elasticsearch/transform/transform.spec.yml
```diff
@@ -34,10 +34,11 @@ spec:
           - type: array
             items:
               type: string
+              pattern: "^[^,]+$"
           - type: string
```
jsoriano, line 38: Nit. Maybe the pattern below should go here.
```suggestion
          - type: string
            pattern: "^[^,]+$"
```

jsoriano, line 38: Nit. Maybe we could reuse the definition, to don't have to repeat the pattern.

jsoriano (approved):
:tada:

mrodm:
It looks like that even if the source indices in the transform are defined as a list of comma-separated strings, the documents are being able to be processed by the transform.

Closing by now this PR.

PR #870, opened 2025-02-11
## What does this PR do?

patrykkopycinski:
Adds `kibana/security_ai_prompt` to support security AI prompt assets.

## Why is it important?

In order to have the flexibility to tweak AI prompts outside of our regular ESS release schedule, `kibana/security_ai_prompt` assets introduce the ability to ship prompt updates for the security AI Assistant and Attack Discovery.

## Checklist

## Related issues

- <https://github.com/elastic/security-team/issues/11196>

## Checklist

## Related issues

-

### DIFF
--- a/spec/integration/kibana/spec.yml
+++ b/spec/integration/kibana/spec.yml
@@ -65,6 +65,15 @@ spec:
       type: file
       contentMediaType: "application/json"
       pattern: '^.+\.json$'
+  - description: Folder containing security AI prompt assets
+    type: folder
+    name: security_ai_prompt
+    required: false
+    contents:
+    - description: A security AI prompt asset file
+      type: file
+      contentMediaType: "application/json"
+      pattern: '^{PACKAGE_NAME}-.+\.json$'
   - description: Folder containing rules
     type: folder
     name: "security_rule"
@@ -135,7 +144,7 @@ spec:
       contentMediaType: "application/json"
       pattern: '^{PACKAGE_NAME}-.+\.json$'
       forbiddenPatterns:
-        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden 
+        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
 versions:
   - before: 3.4.0
     patch:
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -10,6 +10,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add kibana/security_ai_prompt to support security AI prompt assets.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/TODO_UPDATE_LINK_AFTER_OPENING_PR
 - version: 3.3.2
   changes:
     - description: Add support for required conditional groups of variables.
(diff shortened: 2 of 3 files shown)


spec/integration/kibana/spec.yml
```diff
@@ -65,6 +65,15 @@ spec:
...
       pattern: '^.+\.json$'
+  - description: Folder containing security AI prompt assets
+    type: folder
+    name: security_ai_prompt
+    required: false
+    contents:
+    - description: A security AI prompt asset file
+      type: file
+      contentMediaType: "application/json"
+      pattern: '^{PACKAGE_NAME}-.+\.json$'
```
jsoriano, line 76: How are these prompts related to the data collected by the integration?

If these prompts are not specially related to any data, maybe they should be supported only in content packages.

spec/integration/kibana/spec.yml
```diff
@@ -135,7 +144,7 @@ spec:
       contentMediaType: "application/json"
       pattern: '^{PACKAGE_NAME}-.+\.json$'
       forbiddenPatterns:
-        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden 
+        - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
 versions:
```
jsoriano, line 148: We should add a version patch so this cannot be used in older versions of the spec.
```suggestion
versions:
  - before: 3.4.0
    patch:
      - op: remove
        path: "/contents/6" # remove AI prompt definitions
```

spec/integration/kibana/spec.yml
```diff
@@ -65,6 +65,15 @@ spec:
       type: file
       contentMediaType: "application/json"
       pattern: '^.+\.json$'
+  - description: Folder containing security AI prompt assets
+    type: folder
+    name: security_ai_prompt
+    required: false
+    contents:
+    - description: A security AI prompt asset file
```
jsoriano, line 73: Are they specific to security or they could be used in other solutions?
```suggestion
  - description: Folder containing AI prompt assets
    type: folder
    name: ai_prompt
    required: false
    contents:
    - description: An AI prompt asset file
```

spec/integration/kibana/spec.yml
jsoriano: A sample package with a more real world use case could help to illustrate this.

jsoriano (commented):
I have some questions about how this is expected to be used. Thanks!

spec/integration/kibana/spec.yml
stephmilovic: By sample package... do you mean the example added in `test/packages/good/kibana/security_ai_prompt/good-security-ai-prompt-1.json` or something different?

spec/integration/kibana/spec.yml
stephmilovic: This is security only

spec/integration/kibana/spec.yml
jsoriano: I mean something like this, but with a complete example of how this would be used in a real package.

Adding things to the `good` package is good for testing, but for this case it may be good to have a more illustrative example.

spec/integration/kibana/spec.yml
jsoriano: Great, thanks for the ping.

So we can close this PR and continue in https://github.com/elastic/package-spec/pull/871? :)

stephmilovic:
Closing in favor of https://github.com/elastic/package-spec/pull/871

spec/integration/kibana/spec.yml
stephmilovic: @jsoriano I'm still confused about the `6`. Since I'm adding it after SLO, wouldn't it be `14`?

Issue #803, opened 2024-09-19
[Change Proposal] Support additional asset types in content packages


kpollich:
Follow-up for https://github.com/elastic/package-spec/issues/351

Today, content packages only supports dashboards or simple Kibana saved objects that can be directly imported. We should expand this support to allow for all Kibana asset types listed here: https://github.com/elastic/package-spec?tab=readme-ov-file#supported-assets

kpollich:
@jsoriano Now that content packages are GA, it'd be great to provide a path to get the prebuilt rules package moved over to `type: content`. I think we need to add security rules as a supported asset type and then things should work pretty much as expected? Not sure if I'm missing any complexity there. I'm going to schedule this for our next sprint to take a look.

jsoriano:
Yes, in principle we can start adding additional assets. For prebuilt rules I think we also needed to improve support for big files and packages in Fleet.

mrodm:
Currently, the Kibana assets that are allowed in content packages are dashboards and SLOs ([kibana folder spec definition](https://github.com/elastic/package-spec/blob/77bd5412585cbbce62afb7de1fe12168c55b3aa7/spec/content/kibana/spec.yml))

The current list of Kibana assets that are supported in the integration packages is defined [here](https://github.com/elastic/package-spec/blob/77bd5412585cbbce62afb7de1fe12168c55b3aa7/spec/integration/kibana/spec.yml):
- Dashboards
- Visualizations
- Searches
- Maps
- Lens
- Index Patterns
- Security Rules
- CSP Rule Templates
- ML modules
- Tags
- Osquery Pack Assets
- Osquery Saved Queries
- SLOs


As it is recommended to define visualizations, maps, lens, ... assets inside the same dashboards, probably those kind of assets could be skipped (related to https://github.com/elastic/package-spec/issues/316).
Currently, the following Kibana assets cannot be defined via references in dashboards ([package-spec validation rule](https://github.com/elastic/package-spec/blob/77bd5412585cbbce62afb7de1fe12168c55b3aa7/code/go/internal/validator/semantic/validate_visualizations_used_by_value.go#L78)):
- Lens
- Maps
- Searches
- Visualizations


Looking into the packages defined in the [integrations repository](https://github.com/elastic/integrations), there are some assets that it looks like they would require transforms or data stream definitions (or at least those packages contain those other resources). Examples found:
[...]

PR #885, opened 2025-03-21
Allow security rule assets in content packages

mrodm:
## What does this PR do?

Allow to define `security_rule` assets in content packages.

## Why is it important?

Add support in content packages for other different asset types.

## Checklist

## Related issues

- Relates #803


spec/content/kibana/spec.yml
```diff
@@ -29,3 +29,12 @@ spec:
       pattern: '^{PACKAGE_NAME}-.+\.json$'
       forbiddenPatterns:
         - '^.+-(ecs|ECS)\.json$' # ECS suffix is forbidden
+  - description: Folder containing rules
```
mrodm, line 32: Should we include a JSON patch to remove this kind of assets (e.g. remove for spec versions < 3.4.0 ? 
I think it could be added since that should be already supported by Kibana. WDYT?

test/packages/good_content/manifest.yml
```diff
@@ -14,6 +13,8 @@ conditions:
     version: '^8.16.0'
   elastic:
     subscription: 'basic'
+    capabilities:
+      - security
```
mrodm, line 17: As there are `kibana/security_rule/*.json` assets/files, it is required to add this capability.

spec/content/kibana/spec.yml
jsoriano: If it is already supported we would not need to add a JSON patch.

mrodm: Security rules are shown in the Assets tab when installing this `good_content` package:
Assets installed

mrodm: Added compliance tests to check about the existence of security rules.

Even if the installation of the package uploads those assets (as it is shown in the previous screenshot), it was necessary to load the prebuilt detection rules via API in order to be able to get the given rule using the API:
- https://www.elastic.co/guide/en/security/8.17/prebuilt-rules-api.html
- https://www.elastic.co/guide/en/security/8.17/rules-api-get.html

If the prebuilt detetion rules were not loaded, those rules were available in the UI to create new rules, but when querying via API those were not found.
View add Elastic rules
```
GET kbn:/api/detection_engine/rules?rule_id=12cea9e9-5766-474d-a9dc-34ef7c7677c6

{
  "message": """rule_id: "12cea9e9-5766-474d-a9dc-34ef7c7677c6" not found""",
  "status_code": 404
}
```

compliance/features/basic.feature
```diff
@@ -16,11 +16,15 @@ Feature: Basic package types support
   @3.3.0
```
jsoriano, line 16: Is it expected for this package to work with `3.3.0`?

compliance/features/basic.feature
```diff
@@ -16,11 +16,15 @@ Feature: Basic package types support
...
   Scenario: Basic content package can be installed
    Given the "basic_content" package is installed
+     And prebuilt detection rules are loaded
```
jsoriano, line 19: What are these prebuilt rules? Are they needed to install the packages or other rules?

compliance/kibana.go
```diff
@@ -398,13 +407,98 @@ func (k *Kibana) getDashboard(dashboardID string) (*dashboardResponse, error) {
...
+
+	return &detectionRule, nil
+}
+
 func (k *Kibana) newRequest(method string, path string, body io.Reader) (*http.Request, error) {
-	url, err := url.JoinPath(k.Host, path)
+	return k.newRequestWithQueryParams(method, path, map[string]string{}, body)
+}
+
+func (k *Kibana) newRequestWithQueryParams(method string, path string, queryParams map[string]string, body io.Reader) (*http.Request, error) {
```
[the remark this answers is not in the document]
jsoriano: Nit. In other cases where we need to modify the request we use `newRequest`, and then modify it before using `doRequest`. Cannot we do the same in this case?

I am ok with adding a method, but I would prefer to keep the interface simple, otherwise we may be tempted to add methods like `newRequestWithHeaders`, or `newRequestWithHeadersAndQueryParams`...

We could also add a helper to make it easier to add query params, so we can do something like:
```
request, err := newRequest(method, path, nil)
...
request = addRequestParams(request, queryParams)

spec/content/kibana/spec.yml
jsoriano: > it was necessary to load the prebuilt detection rules via API in order to be able to get the given rule using the API

This looks like something Fleet should take care of, right? In that case maybe the compliance tests should not install the prebuilt rules, and fail. Then we would need to move this to `3.4.0`, and request support in Fleet.

> If the prebuilt detetion rules were not loaded, those rules were available in the UI to create new rules, but when querying via API those were not found.

This is weird. Where does the UI get them from? Is it possible to list all rules using the API? Maybe they are listed there without needing to install the prebuilt rules?

test/packages/good_content/manifest.yml
jsoriano: Should we also add a semantic validation rule that checks this?

mrodm: There is already in place one semantic validation rule for that. It failed for me locally, that's why I added here.

spec/content/kibana/spec.yml
mrodm: This should be already supported at least `security_detection_engine` package is using these rules. In its readme it is also mentioned to install prebuilt rules assets:
https://github.com/elastic/integrations/blob/92a21690126ad7500f353f574ce7be89f8d92fe7/packages/security_detection_engine/docs/README.md?plain=1#L5

Not sure if this requires support in Fleet. WDYT @kpollich ?

compliance/features/basic.feature
mrodm: Those prebuilt rules are the ones present in the package plus other rules (in total there are more than 1000).

compliance/kibana.go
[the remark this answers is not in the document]
mrodm: Agreed, I'll add a helper as you mentioned, thanks!
Done in https://github.com/elastic/package-spec/pull/885/commits/51b102b9678b07c997c3dcdec0da5a97e0a9df4b

spec/content/kibana/spec.yml
mrodm: It looks like that by default all rules are disabled when installing these packages, as mentioned here:
https://www.elastic.co/guide/en/security/8.17/prebuilt-rules-api.html#_load_prebuilt_rules

> By default, all loaded prebuilt rules are disabled.

Just a few rules which are in promotion mode are enabled by default. And the rules created in this test packages are not of them

So, in order to load and enable the rules defined in these test packages, so they are present in the GET api, it is required (or I think so) to call first the prebuilt API
https://www.elastic.co/guide/en/security/8.17/prebuilt-rules-api.html

could you confirm if this the expected behavior @elastic/threat-research-and-detection-engineering ? thanks!!

test/packages/good_content/manifest.yml
mrodm: This is the semantic validation rule for this scenario:
https://github.com/elastic/package-spec/blob/622b3a8f000ea7f40d5c2ad64801b94be4fa2522/code/go/internal/validator/semantic/validate_capabilities_required.go#L26

spec/content/kibana/spec.yml
jsoriano: Well, if this is working this way, let's keep the same behavior in content packages, and we can reconsider later if needed.

Issue #89, opened 2020-11-26
Support for shared fields mapping and pipelines between data streams


narph:
We have several cases where we need  to define mapping for fields at package level,  atm we have to define them under each data stream, same goes for `ecs` fields like `message` or `error.message` etc.
Would be great to provide a mapping at package level or even higher when it comes to ecs fields.

Same situation for pipelines, in some cases same processing of fields is required in all data streams for an integration, atm we have to define the shared pipeline at each data stream level.

ruflin:
I would suggest to first solve this issue in integrations with automation, meaning during the build step these are copied / duplicated to the right place.

We should also discuss if we need it in package-spec and with it support it in Kibana but it will make things a bit more complicated.

ycombinator:
Yeah, I think we could generalize this notion to allow references to other files, like includes, from any package file. Then as part of `elastic-package build` we could resolve these includes by inlining the referenced content.

marc-gr:
Going to take a shot at this one

PR #888, opened 2025-04-01
Add spec for shared folder and link files

marc-gr:
## What does this PR do?

Adds a spec definition for a `shared` folder under `_dev` that will allow users to put in there any files that are going to be linked from multiple data_streams and/or packages.

Adds a spec definition to allow `*.link` files under `pipelines`, `agent`, and `fields` to let users reuse repeated files from other parts of the integrations repository.

Adds a test package for both.

## Checklist

## Related issues

- Closes https://github.com/elastic/package-spec/issues/89


code/go/pkg/validator/validator_test.go
```diff
@@ -813,6 +813,12 @@ func TestValidateForbiddenDataStreamName(t *testing.T) {
 	}
 }
 
+func TestSharedFolderAndIgnoreLinks(t *testing.T) {
+	pkgRootPath := filepath.Join("..", "..", "..", "..", "test", "packages", "with_links")
+	errs := ValidateFromPath(pkgRootPath)
+	require.NoError(t, errs)
```
jsoriano, line 819: Nit. No special handling is done for this package, I think it could be directly added to [TestValidateFile](https://github.com/elastic/package-spec/blob/7605ac8d476aba086fa791ef16c94b53b26ea19f/code/go/pkg/validator/validator_test.go#L24).

code/go/internal/validator/semantic/types.go
```diff
@@ -214,6 +215,9 @@ func readFieldsFolder(fsys fspath.FS, fieldsDir string) ([]string, error) {
 	}
 
 	for _, f := range fs {
+		if filepath.Ext(f.Name()) != ".yml" && filepath.Ext(f.Name()) != ".yaml" {
+			continue
+		}
```
jsoriano, line 220: I was thinking more in something like this for these validators:
```suggestion
		if filepath.Ext(f.Name()) == ".link" {
			link, err := linkedfiles.New(f.Name())
			if err != nil { ... }
			f, err = os.Stat(link.LinkFilePath)
			if err != nil { ... }
		}
```

Or even better, we could maybe override the `Open()` function in the `fspath.FS` filesystems that validators receive, to make this transparent for them.

code/go/pkg/linkedfiles/linkedfiles.go
jsoriano, line 1: Not sure if the spec needs all the logic, I think it would only need to be able to resolve links, maybe checking the checksums.

code/go/pkg/validator/validator.go
```diff
@@ -10,14 +10,31 @@ import (
...
+	"github.com/elastic/package-spec/v3/code/go/pkg/linkedfiles"
 )
 
 // ValidateFromPath validates a package located at the given path against the
 // appropriate specification and returns any errors.
 func ValidateFromPath(packageRootPath string) error {
+	root, err := linkedfiles.FindRepositoryRoot()
+	if err != nil {
+		return err
+	}
```
jsoriano, line 26: This function already receives the root of the package, we should be able to rely on it.
```suggestion
	root, err := os.OpenRoot(packageRootPath)
	if err != nil {
		return nil, fmt.Errorf("could not open root: %w", err)
	}
```

code/go/pkg/validator/validator.go
```diff
@@ -10,14 +10,31 @@ import (
...
+	links, err := linkedfiles.IncludeLinkedFiles(root, pkgPath, "")
+	if err != nil {
+		return err
+	}
+	defer func() {
+		for _, link := range links {
+			_ = root.Remove(link.TargetFilePath)
+		}
+	}()
 	return ValidateFromFS(packageRootPath, os.DirFS(packageRootPath))
```
jsoriano, line 38: We could explore the option of wrapping the FS with an `Open()` that transparently handles links.

code/go/pkg/validator/validator.go
```diff
@@ -10,14 +10,31 @@ import (
...
+	pkgPath, _ = filepath.Rel(root.Name(), pkgPath)
+	links, err := linkedfiles.IncludeLinkedFiles(root, pkgPath, "")
+	if err != nil {
+		return err
+	}
+	defer func() {
+		for _, link := range links {
+			_ = root.Remove(link.TargetFilePath)
+		}
+	}()
```
jsoriano, line 37: I would prefer not needing to modify the working copy for spec validation.

spec/integration/data_stream/agent/spec.yml
```diff
@@ -7,8 +7,12 @@ spec:
...
     contents:
-    - description: Folder containing agent stream definitions
+    - description: Config template file for inputs defined in the policy_templates section of the top level manifest
       type: file
       sizeLimit: 2MB
       pattern: '^.+.yml.hbs$'
       required: true
+    - description: Link to a config template file for inputs defined in the policy_templates section of the top level manifest
+      type: file
+      pattern: '^.+.yml.hbs.link$'
```
mrodm, line 17: It is needed to escape some dots:
```suggestion
      pattern: '^.+.yml\.hbs\.link$'
```

spec/integration/data_stream/fields/spec.yml
```diff
@@ -7,3 +7,7 @@ spec:
     required: true
     contentMediaType: "application/x-yaml"
     $ref: "./fields.spec.yml"
+  - description: Link to fields definition
+    type: file
+    pattern: '^[a-z0-9][a-z0-9_-]+[a-z0-9]\.yml.link$'
```
mrodm, line 12:
```suggestion
    pattern: '^[a-z0-9][a-z0-9_-]+[a-z0-9]\.yml\.link$'
```

spec/integration/data_stream/spec.yml
```diff
@@ -68,6 +68,10 @@ spec:
           contentMediaType: "application/json"
           required: false
           $ref: "../../integration/elasticsearch/pipeline.spec.yml"
+        - description: Link to ingest pipeline definition
+          type: file
+          pattern: '^.+\.(yml|json).link$'
```
mrodm, line 73:
```suggestion
          pattern: '^.+\.(yml|json)\.link$'
```

spec/integration/elasticsearch/spec.yml
```diff
@@ -30,6 +30,10 @@ spec:
           contentMediaType: "application/json"
           required: false
           $ref: "./pipeline.spec.yml"
+        - description: Link to ingest pipeline definition
+          type: file
+          pattern: '^.+\.(yml|json).link$'
```
mrodm, line 35:
```suggestion
          pattern: '^.+\.(yml|json)\.link$'
```

spec/integration/agent/spec.yml
```diff
@@ -12,3 +12,7 @@ spec:
       sizeLimit: 2MB
       pattern: '^.+.yml.hbs$'
```
mrodm, line 13: It looks like it was also missing to escape the dots from the extensions in this pattern.

```suggestion
      pattern: '^.+\.yml\.hbs$'
```

marc-gr:
I simplified this as suggested:

- Only logic in package-spec is now related to finding the included files on validation
- Removes any extra logic and does this transparently using a custom fs.FS
- As suggested in https://github.com/elastic/elastic-package/pull/2483 now included files are relative to the link location

For this to be effective I had to set `required: false` to agent and files specs. This is not ideal but I am not sure how to add an `anyOf` condition in there. I see a similar case in https://github.com/elastic/package-spec/blob/main/spec/integration/elasticsearch/spec.yml#L24 and as far as I can tell it does not seem addressed either, so any suggestions in that direction are appreciated.

I think once we solve that this seems enough from the spec perspective.

spec/integration/agent/spec.yml
```diff
@@ -12,3 +12,7 @@ spec:
...
       required: true
+    - description: Link to a config template file for inputs defined in the policy_templates section of the top level manifest
+      type: file
+      pattern: '^.+.yml.hbs.link$'
```
mrodm, line 17:
```suggestion
      pattern: '^.+\.yml\.hbs\.link$'
```

spec/integration/data_stream/agent/spec.yml
```diff
@@ -7,8 +7,12 @@ spec:
     required: true
     additionalContents: false
     contents:
-    - description: Folder containing agent stream definitions
+    - description: Config template file for inputs defined in the policy_templates section of the top level manifest
       type: file
       sizeLimit: 2MB
       pattern: '^.+.yml.hbs$'
```
mrodm, line 13: Same here:
```suggestion
      pattern: '^.+\.yml\.hbs$'
```

spec/integration/agent/spec.yml
mrodm: Created PR to update those regexes that I found in the spec
https://github.com/elastic/package-spec/pull/889

mrodm:
> For this to be effective I had to set `required: false` to agent and fields specs. This is not ideal but I am not sure how to add an `anyOf` condition in there. I see a similar case in https://github.com/elastic/package-spec/blob/main/spec/integration/elasticsearch/spec.yml#L24 and as far as I can tell it does not seem addressed either, so any suggestions in that direction are appreciated.

Not sure it is supported the `anyOf` in the spec related to fields or folders currently. IIRC the spec for folders and the files they could contain are not validated using the JSON schema library. It uses a syntax similar to JSONschema but the validation is coded here:

https://github.com/elastic/package-spec/blob/ed8dc58326d2ce9456de4800624cbb9b810449a8/code/go/internal/validator/folder_spec.go#L43

And here the syntax of each element (contents):
https://github.com/elastic/package-spec/blob/ed8dc58326d2ce9456de4800624cbb9b810449a8/code/go/internal/specschema/folder_item_spec.go#L124

For those elements defining files in `spec.yml`, the `$ref` field will load the corresponding JSON schema to be used in those files (using the JSONSchema library):
https://github.com/elastic/package-spec/blob/ed8dc58326d2ce9456de4800624cbb9b810449a8/code/go/internal/yamlschema/loader.go#L32

I would say `anyOf` (neither other conditions like `allOf`) is not supported in those `spec.yml` files. @jsoriano is my understanding correct? 

If it cannot be used `anyOf`, I guess it should be added new semantic validation rules to ensure that one of those two type of files is present.
Those semantic validation rules are added here:
https://github.com/elastic/package-spec/blob/ed8dc58326d2ce9456de4800624cbb9b810449a8/code/go/internal/validator/spec.go#L216

marc-gr:
@mrodm I added a semantic check for our case and the pipelines one, not sure this is the way you prefer them to be defined so lmk if any changes are required

code/go/pkg/validator/validator_test.go
```diff
@@ -813,6 +814,41 @@ func TestValidateForbiddenDataStreamName(t *testing.T) {
 	}
 }
 
+func TestValidateAnyOfContents(t *testing.T) {
+	tests := map[string][]string{
+		"missing_required_files": {
+			"item [.keep] is not allowed in folder [../../../../test/packages/missing_required_files/agent/input]",
+			"item [.keep] is not allowed in folder [../../../../test/packages/missing_required_files/data_stream/foo/agent/stream]",
+			"item [.keep] is not allowed in folder [../../../../test/packages/missing_required_files/data_stream/foo/elasticsearch/ingest_pipeline]",
+			"item [.keep] is not allowed in folder [../../../../test/packages/missing_required_files/data_stream/foo/fields]",
```
jsoriano, line 823: Are the `.keep` files necessary? Would the missing files be found if the directories would not exist?

code/go/pkg/linkedfiles/fs.go
```diff
@@ -0,0 +1,46 @@
...
+	inner   fs.FS
+}
+
+// NewLinksFS creates a new LinksFS.
+func NewLinksFS(workDir string, inner fs.FS) *LinksFS {
+	return &LinksFS{workDir: workDir, inner: inner}
+}
+
+// Open opens a file in the filesystem.
+func (lfs *LinksFS) Open(name string) (fs.File, error) {
```
jsoriano, line 32: Should we also wrap opening directories, or [`ReadDir()`](https://pkg.go.dev/io/fs#ReadDirFS)? 

I wonder if we may be missing files when walking directories.

The wrap would modify all `DirEntries` to refer to the linked files, so it would be transparent and we would not need `ValidateAnyOfRequiredContents`, or the changes in the spec files. We could also have here the list of allowed links.

In any case the current implementation makes sense to me and also seems to capture the needings we have. We can later iterate on this idea if needed.

spec/input/_dev/spec.yml
```diff
@@ -17,3 +17,7 @@ spec:
       name: test
       required: false
       $ref: "./test/spec.yml"
+    - description: Folder containing shared files.
+      type: folder
+      name: shared
+      $ref: "./shared/spec.yml"
```
jsoriano, line 23: This could refer to the definition in the integrations directory to avoid duplicated definitions, as they are going to be the same.
```suggestion
      $ref: "../../integration/shared/spec.yml"
```

spec/integration/_dev/spec.yml
```diff
@@ -22,3 +22,8 @@ spec:
       name: test
       required: false
       $ref: "./test/spec.yml"
+    - description: Folder containing shared files.
+      type: folder
+      name: shared
+      required: false
+      $ref: "./shared/spec.yml"
```
jsoriano, line 29: Should this be added also to content packages?

code/go/pkg/linkedfiles/fs.go
jsoriano: > I wonder if we may be missing files when walking directories.

Self-reply to this. Checked locally to move the required fields of a data stream of the `good_v3` package to a linked file, and current implementation correctly detects them. So we seem to be good.

Maybe we could add this test case and/or some unit test that ensures that at least the tests that use `validateFields` properly loop over all linked field files.

This is the patch of the local test I have tried:

```diff
diff --git a/test/packages/good_v3/data_stream/foo/fields/base-fields.yml b/test/packages/good_v3/_dev/shared/base-fields.yml
similarity index 100%
rename from test/packages/good_v3/data_stream/foo/fields/base-fields.yml
rename to test/packages/good_v3/_dev/shared/base-fields.yml
diff --git a/test/packages/good_v3/data_stream/foo/fields/base-fields.yml.link b/test/packages/good_v3/data_stream/foo/fields/base-fields.yml.link
new file mode 100644
index 00000000..94bff2ba
--- /dev/null
+++ b/test/packages/good_v3/data_stream/foo/fields/base-fields.yml.link
@@ -0,0 +1 @@
+../../../_dev/shared/base-fields.yml 092c60ec1f7725d278aa0564d946f6803736c436239c4ca20049013a4ce8e91c
```

jsoriano (commented):
Overall looks good, thanks for adding the suggestions. Added some small questions.

code/go/pkg/validator/validator_test.go
marc-gr: git does not commit empty folders otherwise, and for the test I feel is more useful that the paths exist

spec/integration/_dev/spec.yml
marc-gr: I thought about that, but content packages are only supposed to hold files that we do not support links from currently IIRC

code/go/pkg/validator/validator_test.go
jsoriano: Yes, I was wondering if we would be missing the detection of the missing files if the directory would not exist. But I think that for these cases we are setting `required: true` in the directories.

code/go/internal/validator/semantic/validate_anyof_required_contents.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return specerrors.NewStructuredErrorf("no file matching any of the patterns %v found in %s", a.anyOfPatterns, path)
+}
+
+// ValidateAnyOfRequiredContents validates that at least one file matching
+// any of the patterns in the given path exists in the package.
+// It checks the following paths:
+// - agent/input
+// - agent/stream
+// - fields
```
mrodm, line 78: Should we remark that this `fields` folder is just for the ones defined under `data_stream` ? Is that right?

If so, would it be better to re-phrase that item in the comment as this? WDYT @marc-gr ?
```suggestion
// - data_stream/*/fields
```

IIUC the fields folder in `transforms` is not taken into account here. If it would be needed those links in transforms, I guess it could be added support in a follow-up. Let's continue with the current links defined in this PR , is that ok @jsoriano?

marc-gr:
@mrodm I opted to add an `allowLink` option into the item spec after discussing it with @jsoriano as it seems more robust and not rely on the semantic validation, I think this is a cleaner approach together with the custom fs

code/go/pkg/validator/validator.go
```diff
@@ -13,12 +13,15 @@ import (
...
 	"github.com/elastic/package-spec/v3/code/go/internal/validator"
+	"github.com/elastic/package-spec/v3/code/go/pkg/linkedfiles"
 )
 
 // ValidateFromPath validates a package located at the given path against the
 // appropriate specification and returns any errors.
 func ValidateFromPath(packageRootPath string) error {
-	return ValidateFromFS(packageRootPath, os.DirFS(packageRootPath))
+	// We wrap the fs.FS with a linkedfiles.LinksFS to handle linked files.
+	linksFS := linkedfiles.NewLinksFS(packageRootPath, os.DirFS(packageRootPath))
```
jsoriano, line 23: Food for thought regarding the use of the link FS only on `ValidateFromPath`.

I think it is fine to use the link FS only when validating from path, but not when validating from zip. Zip files are in principle built packages and should not have links.
It also looks fine to don't use the link FS in `ValidateFromFS`, so the caller can decide if using it or not.

The potential problem I see is that even if we are validating without the link FS, the spec validation is going to use `allowLink`. This may lead to weird issues if a built package includes links. As the spec will allow the link, but the FS won't resolve it.

I see two potential approaches to solve this:
- Easier, create another FS wrapper, that always fails when trying to open a link. This wrapper would be used at least on `ValidateFromZip`.
- Trickier, in `ValidateFromFS`, check if `fsys` is a `LinkFS`. If it is not, disallow all links.

code/go/pkg/linkedfiles/fs.go
```diff
@@ -0,0 +1,47 @@
...
+// If a linked file is found, it reads the link file to determine the target file
+// and its checksum. If the target file is up to date, it returns the target file.
+// Otherwise, it returns an error.
+type LinksFS struct {
+	workDir string
+	inner   fs.FS
+}
+
+// NewLinksFS creates a new LinksFS.
+func NewLinksFS(workDir string, inner fs.FS) *LinksFS {
```
jsoriano, line 27: Nit. Maybe `Links` is redundant here given the name of the package.
```suggestion
// FS is a filesystem that handles linked files.
// It wraps another filesystem and checks for linked files with the ".link" extension.
// If a linked file is found, it reads the link file to determine the target file
// and its checksum. If the target file is up to date, it returns the target file.
// Otherwise, it returns an error.
type FS struct {
	workDir string
	inner   fs.FS
}

// NewFS creates a new FS that resolves links.
func NewFS(workDir string, inner fs.FS) *FS {
```

code/go/pkg/linkedfiles/fs.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+var _ fs.FS = (*LinksFS)(nil)
+
+// LinksFS is a filesystem that handles linked files.
+// It wraps another filesystem and checks for linked files with the ".link" extension.
+// If a linked file is found, it reads the link file to determine the target file
+// and its checksum. If the target file is up to date, it returns the target file.
+// Otherwise, it returns an error.
+type LinksFS struct {
```
jsoriano, line 21: We are going to publicly expose this FS, but not sure if this should be used out of the spec validation.

Do you think this can be useful in elastic-package? What do you think about moving it to an internal package by now?

In any case I think we can leave the `Link` type in a public package as it can be more clearly reused in elastic-package.

I would be fine also with letting everything here in a private package, and have its own implementation in elastic-package, as the needings in both places are not exactly the same.

jsoriano (commented):
> I opted to add an `allowLink` option into the item spec after discussing it with @jsoriano as it seems more robust and not rely on the semantic validation, I think this is a cleaner approach together with the custom fs.

Thanks for trying this approach, it looks clearer and I feel more confident merging this. I think there can still be some corner cases, but they will be more limited.

I think this could be merged, but added a couple of questions.
Before merging I would like to double-check what parts of `linkedfiles` we want/need to publicly expose.

code/go/pkg/linkedfiles/fs.go
marc-gr: I think makes sense to move it to a private package and let ep handle it on its own

test/packages/missing_required_files/manifest.yml
```diff
@@ -0,0 +1,28 @@
+format_version: 1.0.0
+name: good
```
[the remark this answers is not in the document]
marc-gr: Will remove it, now that we do not have the semantic validator this is not used

test/packages/with_links/manifest.yml
```diff
@@ -0,0 +1,50 @@
+format_version: 1.0.0
+name: good
```
jsoriano, line 2: Nit. Rename this package.

jsoriano (approved):
:+1:

spec/changelog.yml
```diff
@@ -18,6 +18,12 @@
     - description: Allow security_rule assets in content package.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/885
+    - description: Add support for _dev/shared folder.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/888
+    - description: Add support for *.link files in agent, pipelines, and fields folders.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/888
```
jsoriano, line 26: This should be moved to a new version block for `3.3.6-next`.

jsoriano (approved):
Thanks! :tada:

mrodm (approved):
Thanks!

PR #889, opened 2025-04-04
Update regexes used in patterns

mrodm:
## What does this PR do?

Update regexes used in patterns to escape the dots where it is needed (e.g. file extensions), or add missing characters to the regex (e.g. `^*` should be `^.*`). Moreover, some patterns have been replaced by name fields, since the spec refers to a specific name.

## Checklist

## Related issues

-


mrodm:
test integrations

jsoriano (approved):
:+1:

Issue #644, opened 2023-10-12
[Change Proposal] Add `number_of_shards` to `elasticsearch_index_template.settings`


kyungeunni:
It now strictly validates `elasticsearch_index_template.settings`, but the spec v3 is missing the definition for `number_of_shards` which blocks the APM package from migrating to v3. 

```
2023/10/12 14:41:13 Warning: package using an unreleased version of the spec (3.0.0-rc1)
Error: building package failed: invalid content found in built zip package: found 1 validation error:
   1. file "$HOME/apm-server/build/packages/apm-8.12.0.zip/data_stream/sampled_traces/manifest.yml" is invalid: field elasticsearch.index_template.settings: Additional property number_of_shards is not allowed
``` 

Related links:
- https://github.com/elastic/apm-server/issues/11643
- https://github.com/elastic/apm-server/pull/11822

PR #911, opened 2025-06-27
Add refresh_interval to spec

stevengoossensB:
## What does this PR do?

Add functionality to set the refresh_interval in the package

## Why is it important?

It allows to manage the refresh_interval settings inside the package, similar to the number_of_shards. This avoids having to create an @custom template.

## Checklist

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -10,6 +10,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add support for `refresh_interval` in data streams.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/911
 - version: 3.4.0
   changes:
     - description: Add kibana/security_ai_prompt to support security AI prompt assets.
--- a/spec/integration/data_stream/manifest.spec.yml
+++ b/spec/integration/data_stream/manifest.spec.yml
@@ -280,6 +280,13 @@ spec:
               type: integer
               minimum: 1
               default: 1
+            refresh_interval:
+              description: >
+                The interval at which the index is refreshed. A lower value means that
+                documents are searchable sooner, but it can impact indexing performance.
+                Set to `-1` to disable automatic refreshes.
+              type: string
+              default: 1s
         mappings:
           description: Mappings section of index template
           type: object
--- a/test/packages/good_v3/data_stream/foo/manifest.yml
+++ b/test/packages/good_v3/data_stream/foo/manifest.yml
@@ -63,6 +63,7 @@ elasticsearch:
           dimension_fields:
             limit: 32
       number_of_shards: 1
+      refresh_interval: 1s
     mappings:
       dynamic: strict
       dynamic_templates:


jsoriano:
Hey @stevengoossensB,

This is actually the kind of package that I would expect needing to configure per case by users, using `@custom`.

Do you have an example of a package that should default to an specific `refresh_interval`?

stevengoossensB:
Hi @jsoriano,

I agree to some extent since those settings depend on the volume of the log source and the specifics of the environment. However, the same could be said for settings like number_of_shards, ILM policies, etc., which are already supported.

Currently, I manage this by maintaining a set of custom templates and deploying them separately. While this works, it feels a bit clumsy, as the configuration is scattered and we have to ensure everything is deployed in the correct order and perform another validation ourselves.

Personally, I don’t see any harm in allowing all index settings to be defined in the spec and passed through to the templates. This would let us take full advantage of the validation provided during package builds and eliminate the need to duplicate these settings in custom templates.

An alternative—albeit more complex—solution would be to add an option for building packages with a flag pointing to a location containing custom configurations (e.g., additions to the manifest, custom pipelines, etc.). This location could then be used for validation and packaging, and the custom components could be automatically installed alongside the package with the same commands.

jsoriano:
> I agree to some extent since those settings depend on the volume of the log source and the specifics of the environment. However, the same could be said for settings like number_of_shards, ILM policies, etc., which are already supported.

At some point we allowed the configuration of any index setting, but we removed this as it was a backdoor for introducing potentially breaking elements. For example it used to be allowed to define any mapping configuration there, and we could not validate all the possible combinations.

`number_of_shards` is probably still there because there were packages using it, and we would break them if we remove it (but maybe we should at some point).

ILM policies are kind of different because it can be interesting to provide safer configurations for example if the package is expected to collect a lot of data, though I would say it is in the fence.

Package Spec is sometimes very restrictive on purpose, to avoid situations that can be difficult to support, or prone to errors.

> I manage this by maintaining a set of custom templates and deploying them separately. While this works, it feels a bit clumsy, as the configuration is scattered and we have to ensure everything is deployed in the correct order and perform another validation ourselves.

I think this is a very valid concern. Maybe we have to think in some way to provide overrides for packages. Though I think this falls in the fields of configuration management.

kpollich:
I'm moving this out of our project as it's a PR and we don't track them for triage. 

I don't think `refresh_interval` belongs in `package-spec` as there's simply no way for a package to know the best value to set by default. Index refreshes are by nature highly resource-intensive, and configuring a custom `refresh_interval` will strongly depend on your Elasticsearch cluster's resource availability, configuration, etc. It'd never make sense for us to publish a package that assumes the proper value for `refresh_interval` in all clusters. 

On `number_of_shards`, looking back through the Git blame lands us on https://github.com/elastic/package-spec/issues/644 which seems to indicate this is a setting relied upon by APM. Since APM is special cased to be auto-installed on Elastic Cloud deployments and manages its data streams in a very heavy-handed way, I think this was acceptable at the time. In hindsight this is absolutely something that should've been controlled through specific index templates controlled in Elastic Cloud's configuration management system. 

> I manage this by maintaining a set of custom templates and deploying them separately. While this works, it feels a bit clumsy, as the configuration is scattered and we have to ensure everything is deployed in the correct order and perform another validation ourselves.

I wonder if a way to package index templates with integrations could be a way to deal with this? Building custom integrations as an encapsulation mechanism for distributing common assets is a very common use case, so I think we should absolutely try to fix this pain point. Perhaps Fleet could spread an additional template object onto the `@package` index template?

jsoriano:
Issue created https://github.com/elastic/package-spec/pull/911.

PR #917, opened 2025-07-01
Add support for an optional value in discovery fields

mrodm:
## What does this PR do?

Allow to declare discovery fields rules matching specific values.
Example of discovery field defining a specific value for a field:
```yaml
discovery:
  fields:
    - name: "data_stream.dataset"
      value: "nginx.access"
```

## Why is it important?

It will allow to show packages that use certain fields and also show packages that use fields with some specific data.

## Checklist

- [x] The `discovery.fields` spec in `manifest.yml` supports an optional `value` field
- [x] Multiple discovery field/value pairs can be defined.
- [x] Backward compatibility is maintained for discovery entries that only specify `name`

## Related issues

-

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -15,7 +15,7 @@
     - description: Add fips_compatible boolean flag for input package policy templates.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/914
-    - description: Add support to define an optinal value for discovery fields.
+    - description: Add support to define an optional value for discovery fields.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/917
 - version: 3.4.0


test/packages/bad_discovery_fields/manifest.yml
```diff
@@ -0,0 +1,32 @@
...
+  - custom
+conditions:
+  kibana:
+    version: "^8.18.3"
+  elastic:
+    subscription: "basic"
+discovery:
+  fields:
+    - name: 12345
+    - value: nginx.access
```
mrodm, line 19: Examples of incorrect discovery fields:
- `name` must be a string.
- a discovery field must contain a `name` field. `value` field is the optional one.

spec/content/manifest.spec.yml
```diff
@@ -22,6 +22,19 @@ spec:
...
+                oneOf:
+                - type: string
+                - type: number
+                - type: boolean
+                examples:
+                - "nginx.access"
+                - true
+                - 12345
+            required:
+              - name
```
mrodm, line 37: Ensure that `name` is always set in each array item.

spec/content/manifest.spec.yml
```diff
@@ -22,6 +22,19 @@ spec:
               name:
                 description: Name of the field.
                 type: string
+              value:
```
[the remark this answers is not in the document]
kpollich: This field will be ignored by Kibana IIUC - there are no user-facing features that make use of the discovery fields in Kibana right now so we are free to make this change without considering impact to Kibana.

spec/content/manifest.spec.yml
```diff
@@ -22,6 +22,20 @@ spec:
...
+                description: >
+                  An optional expected value for the "name" field.
+                oneOf:
+                - type: string
+                  minLength: 1
```
mrodm, line 30: Here, I'm forcing that if it is a string, it should not be an empty string.

For instance, this would not be allowed:
```yaml
discovery
  fields:
    - name: event.dataset
      value: ""
```

Would that make sense ?
Depending on that, the validation in EPR could be different.

jsoriano: > should we allow to compare a field with an empty string?

I guess this is in theory possible, but I don't know what would be the use case. So feel free to keep this `minLength`.

spec/changelog.yml
```diff
@@ -15,6 +15,9 @@
     - description: Add fips_compatible boolean flag for input package policy templates.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/914
+    - description: Add support to define an optinal value for discovery fields.
```
jsoriano, line 18:
```suggestion
    - description: Add support to define an optional value for discovery fields.
```

jsoriano (approved):

jsoriano (approved):

PR #918, opened 2025-07-01
Add alert rule kibana asset to package spec

MichelLosier:
## What does this PR do?

Adds support for `alert_rule_template` Kibana assets

## Why is it important?

We want to support being able to include prebuilt alert rules in packages, this will help set users up operationally for their agents.

## Checklist

## Related issues

* Closes: https://github.com/elastic/kibana/issues/229687
* Requires: https://github.com/elastic/ingest-dev/issues/5901


spec/changelog.yml
```diff
@@ -10,6 +10,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add support for `alert` rule assets.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/918
```
jsoriano, line 15: Does Fleet need changes to support these new assets? If it does, we will also need to add a patch in the `versions` sections of the spec. If it doesn't, we might include this change in `3.4.2-next`.

jsoriano (commented):
Does kibana need changes to support these new assets?

spec/changelog.yml
MichelLosier: Sounds good, thanks for taking a look. Supporting this does require changes on the kibana end ([original PR](https://github.com/elastic/kibana/pull/226085)). We're pivoting on the direction for this and putting the work behind a feature flag. So I'm going to put this back into draft for now, and we'll ready this back up once the new kibana work is in place.

jsoriano (approved):

muthu-mps:
test integrations

PR #922, opened 2025-07-08
Add support for discovery content packages based on datasets

### LINKED ISSUE #919: [Change Proposal] Add a new discovery parameter based on datasets
Currently, content packages are allowed to define different fields to allow for discovery.

The goal of this issue is to add another parameter to allow discovering packages based on the datasets.

Example of how this new parameter could be defined in the manifest:
```yaml
discovery:
  fields:
    - name: host.ip
  datasets:
    - name: nginx.access
    - name: nginx.error
```

### Requirements
- [x] Add new discovery parameter `discovery.datasets` as an array.
- [x] Allow to define both `fields` and `datasets` for discovery independently

mrodm:
## What does this PR do?

Add discovery of content packages based on datasets .
It allows to define a new `discovery.datasets` key in the package manifests.

```yaml
discovery:
  datasets:
    - name: nginx.access
    - name: nginx.error
```

## Checklist

## Related issues

- Closes #919
- Depends on https://github.com/elastic/package-spec/pull/923

### DIFF
--- a/spec/content/manifest.spec.yml
+++ b/spec/content/manifest.spec.yml
@@ -14,7 +14,7 @@ spec:
       additionalProperties: false
       properties:
         datasets:
-          description: Description of the datasets this package can be used with.
+          description: List of the datasets this package can be used with. For a package to be used with an index, the `data_stream.dataset` field of this index should be one of the datasets listed here.
           type: array
           items:
             type: object
@@ -26,7 +26,7 @@ spec:
             required:
               - name
         fields:
-          description: Description of the fields this package can be used with.
+          description: List of fields this package expects to find in an index. For a package to be used with an index, the index should contain all the fields listed here.
           type: array
           items:
             type: object
--- a/spec/content/manifest.spec.yml
+++ b/spec/content/manifest.spec.yml
@@ -13,6 +13,17 @@ spec:
       type: object
       additionalProperties: false
       properties:
+        datasets:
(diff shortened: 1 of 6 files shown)


jsoriano (approved):

spec/content/manifest.spec.yml
```diff
@@ -13,11 +13,24 @@ spec:
...
+            type: object
+            additionalProperties: false
+            properties:
+              name:
+                description: Name of the dataset.
+                type: string
+            required:
+              - name
         fields:
           description: Description of the fields this package can be used with.
```
jsoriano, line 29: Btw, maybe we can extend the descriptions of the `datasets` and `fields`, so it is less confusing how queries are expected to match.
```suggestion
          description: List of fields this package expects to find in an index. For a package to be used with an index, the index should contain all the fields listed here.
```

spec/content/manifest.spec.yml
```diff
@@ -13,11 +13,24 @@ spec:
       type: object
       additionalProperties: false
       properties:
+        datasets:
+          description: Description of the datasets this package can be used with.
```
jsoriano, line 17:
```suggestion
          description: List of the datasets this package can be used with. For a package to be used with an index, the `data_stream.dataset` field of this index should be one of the datasets listed here.
```

jsoriano (commented):
Maybe we should extend documentation of these fields.

Issue #920, opened 2025-07-07
[Change Proposal] Add docs_structure_enforced flag in _dev/build/build.yml


haetamoudi:
Ad part of the [docs improvement project](https://docs.google.com/document/d/1suPRUVDYAqFa1LL-2RIWI7cnj9zI8-LYG0i1eiBfIII/edit?tab=t.0#heading=h.pfm1mse74cxb), we have created a new README structure template to be used for integrations.

To ensure consistency, we want to verify that new integration READMEs follow this structure.

Scope:
This validation will apply only to new integrations for now. Existing integrations will be updated gradually.

To avoid breaking existing integrations, the validation will only be enforced for selected ones. This will be controlled via a flag in the `_dev/build/build.yml` file.

The flag is named: `docs_structure_enforced`

Default value: `false`

jsoriano:
> To ensure consistency, we want to verify that new integration READMEs follow this structure.

How can we ensure that the READMEs follow a given structure?

haetamoudi:
@jsoriano the idea is to validate the section titles are present. We don't want to validate the content.

The section titles are coming from the template created by the docs team https://docs.google.com/document/d/1T05yEE_HJQN_9n5kMkoMZ9L00Msh1ribVoH50Z15pjg/edit?tab=t.0#heading=h.tzrqef91x5gk

Here is a draft PR on the changes on elastic-packagge https://github.com/elastic/elastic-package/pull/2716.
[Here](https://github.com/elastic/elastic-package/blob/2e157ec7418d336efd0c2c84ee0b743b25a26b7b/internal/docs/readme.go#L294) is the validation part

PR #924, opened 2025-07-09
Add docs_structure_enforced to validation.yml

haetamoudi:
[Initial discussion](https://github.com/elastic/package-spec/issues/920)
[RFC](https://docs.google.com/document/d/1suPRUVDYAqFa1LL-2RIWI7cnj9zI8-LYG0i1eiBfIII/edit?tab=t.0#heading=h.pfm1mse74cxb)

## What does this PR do?
This PR enable setting the `docs_structure_enabled` parameter in the `validation.yml`

## Why is it important?
If set, this will be sure to enforce the docs structure of the readme files. It will verify the documentation files contain a predefined set of titles. 
`docs_structure_enabled` contains:
- version: version of template that will be used
- skip: list of title that will be ignored from the validation

## Checklist

## Related issues

Implementation of changes in elastic-package https://github.com/elastic/elastic-package/pull/2716


spec/integration/validation.spec.yml
```diff
@@ -18,3 +18,26 @@ spec:
...
+      additionalProperties: false
+      properties:
+        enabled:
+          description: "Whether the documentation structure is enforced"
+          type: boolean
+          default: true
+        version:
+          description: "Version of the documentation structure"
+          type: integer
+          default: 1
```
jsoriano, line 33: This makes me wonder if the documentation structure also belongs here in the package spec.

With current plan this documentation structure is going to be defined and validated in elastic-package, right?

Making it part of the spec will make it easier to do things like enforcing that for a specific version of the spec, an specific version of the documentation structure is used. This may help on efforts to update documentation as the spec evolves. It would also help to test changes on this validation earlier, and in a more consistent way to other validations.

Though I also think that we can change this later, so not a strong opinion by now.

@mrodm I would also like to have your thoughts on this.

haetamoudi: I actually wanted to get your opinion on that, as I was wondering the same, if it should be in package-specs vs elastic-package. I am fine with either

spec/integration/validation.spec.yml
```diff
@@ -18,3 +18,26 @@ spec:
...
               - SVR00001
+    docs_structure_enforced:
+      description: "Rules to manage the documentation structure"
+      type: object
+      additionalProperties: false
+      properties:
+        enabled:
+          description: "Whether the documentation structure is enforced"
+          type: boolean
+          default: true
```
jsoriano, line 29: It is going to be disabled by default, right?

spec/integration/validation.spec.yml
jsoriano: Then I think I prefer to bring this validation to the spec. We could add it as a validation rule here: https://github.com/elastic/package-spec/blob/d9ff0bb4b78441b523f97a8b57a15d56b033711b/code/go/internal/validator/spec.go#L186

And we could also add some test packages in this repo too.

code/go/internal/pkgpath/files.go
```diff
@@ -84,3 +84,13 @@ func (f File) Values(path string) (interface{}, error) {
...
 	return f.path
 }
+
+// ReadAll reads and returns the entire contents of the file.
+func (f File) ReadAll() ([]byte, error) {
+	r, err := f.fsys.Open(f.path)
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
```
jsoriano, line 94: Is it needed to open the file? `r` is not really used.
```suggestion
```

code/go/internal/validator/semantic/validate_docs_structure.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read schema file: %w", err)
+	}
+
+	var spec EnforcedSections
+	if err := yaml.Unmarshal(data, &spec); err != nil {
+		return nil, fmt.Errorf("invalid schema YAML: %w", err)
+	}
+
+	if spec.Version != version {
```
jsoriano, line 156: How can this happen? Would it be an spec with an incorrect version?

I think we could rely on the name of the file, so we don't need the redundant `version` inside the spec.
```
schemaPath := fmt.Sprintf("enforced_sections_v%s.yml", version)
```

code/go/internal/validator/semantic/validate_docs_structure.go
```diff
@@ -0,0 +1,242 @@
...
+	files, err := pkgpath.Files(fsys, "docs/*.md")
+	if err != nil {
+		return fmt.Errorf("docs folder %s not found: %w", "docs/*.md", err)
+	}
+	if len(files) == 0 {
+		return nil
+	}
+
+	for _, file := range files {
+		if !file.IsDir() {
```
jsoriano, line 181: Nit.
```suggestion
		if file.IsDir() {
			continue
		}
```

code/go/internal/validator/semantic/validate_docs_structure.go
```diff
@@ -0,0 +1,242 @@
...
+			fullPath := path.Join("docs", file.Name())
+			content, err := fs.ReadFile(fsys, fullPath)
```
jsoriano, line 183: Could the new helper be used here?
```suggestion
			content, err := file.ReadAll()
```

code/go/internal/validator/semantic/validate_docs_structure.go
```diff
@@ -0,0 +1,242 @@
...
+		if section.Name != "" {
+			sections = append(sections, section.Name)
+		}
+	}
+
+	return sections, nil
+}
+
+func validateReadmeStructure(fsys fspath.FS, enforcedSections []string) error {
+	var errs DocsValidationError
```
jsoriano, line 171: Nit. We could get the same behaviour as this error (or very similar) with the stdlib, using a vanilla `[]error` here:
```suggestion
	var errs []error
```
And then [`errors.Join`](https://pkg.go.dev/errors#Join):
```
return errors.Join(errs...)
```

spec.go
```diff
@@ -13,7 +13,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-//go:embed spec spec/integration/_dev spec/integration/data_stream/_dev spec/input/_dev
+//go:embed spec spec/integration/_dev spec/integration/data_stream/_dev spec/input/_dev docsValidationSchema
```
jsoriano, line 16: Camel case is not usual in go projects, we could use something like `docs_schema`.
```suggestion
//go:embed spec spec/integration/_dev spec/integration/data_stream/_dev spec/input/_dev docs_schema
```

code/go/internal/validator/semantic/validate_docs_structure_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}{
+		{
+			name:          "Valid test",
+			pkgRoot:       "../../../../../test/packages/good_readme_structure",
+			expectError:   false,
+			expectedError: nil,
+		},
+		{
+			name:        "Invalid test",
+			pkgRoot:     "../../../../../test/packages/bad_readme_structure",
```
jsoriano, line 99: We use to add test packages to `code/go/pkg/validator/validator_test.go`, this helps ensuring for example with consistency on how error messages include the files that produce them.

code/go/internal/validator/semantic/validate_docs_structure.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	return false
+}
+
+func loadSectionsFromConfig(version string) ([]string, error) {
+	var schemaPath string
+	switch version {
+	case "1":
+		schemaPath = "enforced_sections_v1.yml"
+	default:
```
jsoriano, line 141: Should we have a default version? Maybe per spec version? We could add this later too though.

haetamoudi: I removed the `switch case` so not it will just throw an error if the file with the requested version does not exist

code/go/internal/validator/semantic/validate_docs_structure_test.go
haetamoudi: I added some tests

code/go/internal/validator/semantic/validate_docs_structure_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+package semantic
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+	"github.com/stretchr/testify/assert"
```
mrodm, line 13: Nit. Imports ordering

```suggestion
	"github.com/stretchr/testify/assert"

	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
```

code/go/internal/validator/semantic/validate_docs_structure.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load enforced sections from config: %w", err)
+	}
+
+	var sections []string
+	for _, section := range defaultSections {
+		if contains(config.DocsStructureSkip, section) {
+			continue
+		}
+		sections = append(sections, section)
```
mrodm, line 109: I think It could be used `slices.ContainsFunc()` here:
```go
	for _, section := range defaultSections {
		if !slices.ContainsFunc(config.DocsStructureSkip, func(s specerrors.DocsStructureSkip) bool {
			return s.Title == section
		}) {
			sections = append(sections, section)
		}
	}
```
or keeping the same condition:
```go
	for _, section := range defaultSections {
		if slices.ContainsFunc(config.DocsStructureSkip, func(s specerrors.DocsStructureSkip) bool {
			return s.Title == section
		}) {
			continue
		}
		sections = append(sections, section)
	}
```

WDYT?

If that makes sense, the `contains` function from line 115 could be removed

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/otiai10/copy v1.14.1
 	github.com/stretchr/testify v1.10.0
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415
+	github.com/yuin/goldmark v1.7.12
```
mrodm, line 19: There is a new patch version v1.7.13 published a few days ago
https://github.com/yuin/goldmark/releases/tag/v1.7.13

Could you try to update it here?

mrodm (approved):
Thanks!

PR #937, opened 2025-07-25
add description to the categories and remove unused categories

trisch-me:
Added description to categories 
Removed unused categories - I checked all integrations to find categories not present in any integration

This is needed to make an appropriate selection while developing new integration. We could also update PR to add it as a new field to be able to use it in UI if needed.

## Checklist

- ~~[ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~~

## Related issues

- Related 
  https://github.com/elastic/ingest-dev/issues/4788
  https://github.com/elastic/integrations/pull/14571

### DIFF
--- a/spec/integration/manifest.spec.yml
+++ b/spec/integration/manifest.spec.yml
@@ -24,74 +24,110 @@ spec:
       items:
         type: string
         enum:
+          # User and entity behavior analytics for advanced threat detection
           - advanced_analytics_ueba
+          # Tools for data analysis and processing
           - analytics_engine
+          # Monitoring and observability for applications
           - application_observability
-          - app_search
+          # Linux audit daemon monitoring
           - auditd
+          # Identity and access verification systems
           - authentication
+          # Amazon Web Services integrations
           - aws
+          # Microsoft Azure cloud integrations
           - azure
+          # Large-scale data processing systems
           - big_data
+          # Content delivery network security
           - cdn_security
+          # Cloud computing and services
           - cloud
+          # Cloud security and continuous data replication
           - cloudsecurity_cdr
+          # Configuration management tools and systems
           - config_management
+          # General connection interfaces
           - connector
-          - connector_client
-          - connector_package
+          # Container technologies like Docker
           - containers
-          - content_source
-          - crawler
+          # Password and credential management systems
           - credential_management
+          # Customer relationship management systems
           - crm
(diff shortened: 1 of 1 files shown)


trisch-me:
@nimarezainia We do have that list (for the items I have removed) changed 2 or even 4 years ago. If we don’t have packages that support those categories until now, I don’t see the need to have them. So it’s my attempt to clean up the table. What are the arguments to leave them?

As for you question to show up in UI it could be a tooltip on hover or something else? We shouldn’t do it, it was just a suggestion if we should do it based on yours or others opinion.

jsoriano:
Categories and its hierarchy, along with the titles shown in kibana are also defined in the package registry, here: https://github.com/elastic/package-registry/blob/main/categories/categories.yml

They are also listed in kibana:
* Here for categories that should be always listed in the UI, even if not returned by the registry: https://github.com/elastic/kibana/blob/main/src/platform/plugins/shared/custom_integrations/common/index.ts
* Here as a mapping of the possible values: https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/common/types/models/package_spec.ts#L55

I would be also hesitant of removing unused categories, there was some deliberation to end up with this list, and in principle having additional categories doesn't do any harm. Adding them back once removed would be more cumbersome.

I think we should centralize the definition of categories, probably in a similar fashion to the current definition in the registry, though they would make more sense here in the spec. Maybe the file in the registry should be moved here, and the registry should use it.

trisch-me:
> Categories and its hierarchy, along with the titles shown in kibana are also defined in the package registry, here: https://github.com/elastic/package-registry/blob/main/categories/categories.yml

It feels like we have multiple places for categories definition. I think it’s prone to errors if changed in one place without knowing about another one (like I did a change here)

> I think we should centralize the definition of categories, probably in a similar fashion to the current definition in the registry, though they would make more sense here in the spec. Maybe the file in the registry should be moved here, and the registry should use it.

I totally support it, it should be one place to define them all. How should we proceed with it?

I will revert removed categories if you strong disagree with their removal. But I still need factual review of descriptions. Thanks

jsoriano (dismissed):

trisch-me:
@jsoriano are you able to merge this request? It sits like this for a month already, thanks!

PR #945, opened 2025-08-01
Allow to use underscores in benchmark file names

jsoriano:
## What does this PR do?

Allow the use of underscores in benchmark file names.

## Why is it important?

To be used with packages or data streams that include underscores in their names.

## Checklist


mrodm (approved):
LGTM, just a comment about the new changelog entry

mrodm (approved):

PR #949, opened 2025-08-07
feat: add semantic validator for duration var type

andrewkroh:
## What does this PR do?

```
Add validation for duration type variables in package and data stream manifests
to ensure constraints are properly defined. The validation checks that duration
variables follow the ordering:

0 <= min_duration <= default <= max_duration.

Relates elastic/package-spec#948
```

## Why is it important?

This prevents developers from incorrectly specifying min, max, and default duration.

## Checklist

## Related issues

- Relates #948

## Screenshots

With line/col annotations, most IDEs will link you directly to the related code. For example, here is IntelliJ:

duration-var-errors-intellij

### DIFF
--- a/code/go/internal/validator/semantic/validate_duration_variables.go
+++ b/code/go/internal/validator/semantic/validate_duration_variables.go
@@ -128,7 +128,7 @@ type durationManifest struct {
 	} `yaml:"policy_templates"`
 }
 
-// allVars collects all duration variables from a manifest including those at the
+// allVars collects all variables from a manifest including those at the
 // top level, within policy templates, and nested inside inputs. It returns a
 // flattened slice of all variables for easier processing.
 func (m *durationManifest) allVars() []durationVar {
@@ -151,7 +151,7 @@ type durationDataStreamManifest struct {
 	} `yaml:"streams"`
 }
 
-// allVars collects all duration variables from a data stream manifest. It
+// allVars collects all variables from a data stream manifest. It
 // returns a flattened slice of all variables from all streams for easier
 // processing.
 func (m *durationDataStreamManifest) allVars() []durationVar {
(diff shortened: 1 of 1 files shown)


code/go/internal/validator/semantic/validate_duration_variables.go
```diff
@@ -0,0 +1,288 @@
...
+	Vars            []durationVar `yaml:"vars"` // Top-level variables
+	PolicyTemplates []struct {
+		Vars   []durationVar `yaml:"vars"` // Policy template variables
+		Inputs []struct {
+			Vars []durationVar `yaml:"vars"` // Input variables
+		} `yaml:"inputs"`
+	} `yaml:"policy_templates"`
+}
+
+// allVars collects all duration variables from a manifest including those at the
```
jsoriano, line 131: It is collecting all variables, not only duration ones, right?

code/go/internal/validator/semantic/validate_duration_variables.go
```diff
@@ -0,0 +1,288 @@
...
+			return fmt.Errorf("invalid max_duration value %q: %w", *v.MaxDuration, err)
+		}
+	}
+
+	// Check constraints: 0 <= min_duration <= default <= max_duration
+	var errs []error
+
+	// Check min_duration <= default (if both are defined)
+	if v.MinDuration != nil && v.Default != nil {
+		if minDuration.Nanoseconds() > defaultDuration.Nanoseconds() {
```
jsoriano, line 218: Durations can be directly compared, right?
```suggestion
		if minDuration > defaultDuration {
```

jsoriano (approved):

code/go/internal/validator/semantic/validate_duration_variables.go
andrewkroh: Absolutely 👍 . bda8b4c

PR #950, opened 2025-08-07
fix(internal/fspath.go): use filepath.Join instead of path.Join

andrewkroh:
## What does this PR do?

When joining file paths the 'path/filepath' package should be used instead of 'path' to apply the OS specific behavior.

This became apparent during Windows testing where the returned filepath in a validation error looked like

    C:\Users\buildkite\AppData\Local\Temp\TestValidate\001/data_stream/foo/manifest.yml:8:9 error in variable

## Why is it important?

File paths should be rendered correctly regardless of the user's OS.

## Checklist


jsoriano:
Most filesystem operations in the spec are done through `fs.FS`, that uses slash-separated paths even on Windows.

In principle we support Windows, and release elastic-package binaries for this OS. _Functionally_ it should be fine enough for the most common scenarios, though we don't have the same level of testing coverage as for other runtimes.
For the spec we [run unit tests on Windows](https://github.com/elastic/package-spec/blob/main/.buildkite/scripts/run-win-tests.ps1), and for elastic-package we run [some stack tests](https://github.com/elastic/elastic-package/blob/main/.buildkite/scripts/stack_tests_windows.ps1) and the [unit tests suite](https://github.com/elastic/elastic-package/blob/main/.buildkite/scripts/unit_tests_windows.ps1). From time to time I also do some manual testing on Windows.

But it is true that we don't pay a lot of attention to the paths built in error messages, that probably have mixed separators caused by directly joining paths from the native file systems and from `fs.FS`, without using `FromSlash` or anything.

Opening follow up issue about this: https://github.com/elastic/package-spec/issues/952

PR #936, opened 2025-07-25
Add 'lookup' as valid index mode for ES|QL lookup joins

JohannesMahne:
## What does this PR do?

It adds `lookup` as a valid value for index mode, so that an index created by an integration can be a lookup index, to be used in a ES|QL LOOKUP JOIN.

## Why is it important?

Chargeback integration (Draft integration - PR raised) needs.

## Checklist

- [ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.
- [x] I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).

## Related issues

- Relates https://github.com/elastic/integrations/pull/14545

-

PR #964, opened 2025-08-25
Fix transform index spec

jsoriano:
While attempting to follow with https://github.com/elastic/package-spec/pull/936#discussion_r2298336064, I found that no spec was being applied to the transform index manifests.

## What does this PR do?

Fix transform index spec.

We removed `spec` from https://github.com/elastic/package-spec/pull/936 by mistake. After adding it back it is discovered that several mappings are not being properly imported.

Duplicate them by now.

Also, drop `index.mode` from spec versions older than 3.4, as `lookup` mode is not supported in old versions of the stack.

## Why is it important?

Properly validate transform index definitions.

## Checklist

- [ ] ~~I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~~ There are tests covering this.
- [ ] ~~I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).~~ Fixes unreleased changes.

## Related issues

- Follow up to https://github.com/elastic/package-spec/pull/936

### DIFF
--- a/spec/integration/elasticsearch/transform/manifest.spec.yml
+++ b/spec/integration/elasticsearch/transform/manifest.spec.yml
@@ -11,6 +11,24 @@ spec:
           type: object
           additionalProperties: false
           properties:
+            analysis:
+              type: object
+              additionalProperties: false
+              properties:
+                analyzer:
+                  description: Definition of custom analyzer.
+                  type: object
+                  additionalProperties:
+                    type: object
+                    additionalProperties: false
+                    properties:
+                      type:
+                        description: Type of analyzer
+                        type: string
+                        enum:
+                          - pattern
+                      pattern:
+                        type: string
             index:
               type: object
               additionalProperties: false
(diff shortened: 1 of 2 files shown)


JohannesMahne (approved):
Thanks for picking this up and fixing it.

jsoriano:
test integrations

spec/integration/elasticsearch/transform/manifest.spec.yml
```diff
@@ -1,45 +1,61 @@
...
+                  description: >
+                    How often to perform a refresh operation, which makes recent changes to the index visible to search.
+                  type: string
+                  example: 1s
+                sort:
+                  $ref: "../../data_stream/manifest.spec.yml#/definitions/index_template_setting_index/properties/sort"
+                number_of_shards:
+                  $ref: "../../data_stream/manifest.spec.yml#/definitions/index_template_setting_number_of_shards"
+                number_of_routing_shards:
+                  description: Number used when splitting shards.
```
mrodm, line 39: Not sure that this `description` field is in the final JSON schema (after merging the `$ref`). But in any case, at least that description could help the developers as it is part of the code.

mrodm (approved):

PR #965, opened 2025-08-28
fix(data_stream manifest): move required_var minItems

andrewkroh:
## What does this PR do?

minItems was placed incorrectly on an object rather than on the array.

The JSON pointer to the invalid attribute is:

    /definitions/required_vars/patternProperties/^[a-zA-Z0-9_]+$/items/minItems

https://json-schema.org/understanding-json-schema/reference/array#length

## Why is it important?

The validation of required_vars may not be working as intended by the author.

## Checklist

## Related issues

-


jsoriano (approved):
Good one, thanks!

PR #966, opened 2025-08-28
fix(data_stream manifest): dynamic_date_formats example type

andrewkroh:
## What does this PR do?

The dynamic_date_formats examples values are arrays, but they are specified on a type=string. For the examples to be valid they need to be elevated up to the array.

The problematic JSON pointer is

    /definitions/elasticsearch_index_template/properties/mappings/properties/dynamic_date_formats/items/examples

## Why is it important?

The examples should be valid based on our specification.

## Checklist

## Related issues

-


jsoriano (dismissed):
Thanks!

jsoriano (approved):

PR #976, opened 2025-09-04
Fix alerting_rule_template typeMigrationVersion values in test packages

MichelLosier:
## What does this PR do?

Fixes the `typeMigrationVersion` values for `alerting_rule_template` fixtures. 

## Why is it important?

If a user were to test package installation using one of these test packages it would fail on the alerting rule templates due to the wrong typeMigrationVersion for the saved object.

## Checklist

## Related issues

Relates: https://github.com/elastic/kibana/issues/229687


mrodm (dismissed):
LGTM

mrodm (dismissed):

PR #988, opened 2025-09-26
[CI] Add staticcheck tool

mrodm:
## What does this PR do?

Include into check make target another step to run static checks

Errors found:
```
internal/specschema/folder_schema_spec.go:44:28: func (*folderSchemaSpec).patchForVersion is unused (U1000)
internal/validator/content.go:19:3: unnecessary assignment to the blank identifier (S1005)
internal/validator/semantic/validate_fields_limits.go:27:3: unnecessary assignment to the blank identifier (S1005)
internal/validator/semantic/validate_routing_rules_and_dataset.go:29:3: this value of err is never used (SA4006)
internal/yamlschema/loader.go:140:10: unnecessary use of fmt.Sprintf (S1039)
internal/yamlschema/schema_spec.go:50:26: func (*itemSchemaSpec).patchForVersion is unused (U1000)
pkg/validator/limits_test.go:347:2: should merge variable declaration with assignment on next line (S1021)
pkg/validator/validator_fields_test.go:541:9: assigning the result of this type assertion to a variable (switch content := content.(type)) could eliminate type assertions in switch cases (S1034)
	pkg/validator/validator_fields_test.go:545:10: could eliminate this type assertion
```

## Checklist

## Related issues

-


code/go/pkg/validator/validator_fields_test.go
```diff
@@ -538,11 +538,11 @@ func TestValidateFields(t *testing.T) {
 
 // encodeContent encodes the content as YAML, unless it is already an array of bytes, that is returned as is.
 func encodeContent(content any) ([]byte, error) {
-	switch content.(type) {
+	switch content := content.(type) {
```
mrodm, line 541: Should we set another name here ?

code/go/Makefile
```diff
@@ -19,6 +20,9 @@ check: lint check-license check-spec
 # which are not compliant with linter rules. The golint tool doesn't support ignore comments.
 lint:
 	@go list ./... | grep -v yamlschema | xargs -n 1 $(golint_cmd) -set_exit_status
+	@$(staticcheck_cmd) ./...
```
mrodm, line 23: Included in the same target `lint` following the example in `elastic-package` repository
https://github.com/elastic/elastic-package/blob/ddfa3a9027542a448e94d0e389563690e0c55c4b/Makefile#L33

mrodm:
> I have used in the past [golangci-lint](https://github.com/golangci/golangci-lint) which is a runner of linterns, customizable and configurable

Thanks! I'll take a look to it. 
For now, I thought to set the same tools that are used in other repositories like [elastic-package](https://github.com/elastic/elastic-package) and [package-registry](https://github.com/elastic/package-registry). It could be checked later if it is needed to switch to `golangci-lint`.

PR #994, opened 2025-10-07
Input packages don't require to define fields

jsoriano:
With OTel inputs we have seen that all fields required are already defined in the built-in templates, and requiring to add mappings can lead to conflicts with them.

For ECS inputs, required fields are also defined in ecs@mappings, so there would be no need to require them.

These packages can still define fields if needed.

Test package for OTel is updated for current recommendations.

Follow up to https://github.com/elastic/kibana/pull/237165.


spec/input/spec.yml
```diff
@@ -75,6 +75,11 @@ spec:
     $ref: "../integration/img/spec.yml"
 
 versions:
+  - before: 3.5.0
```
mrodm, line 78: Should it be 3.5.1 here ?
Or, is it expected here to set that folder as required just for versions `<3.5.0`?  

```suggestion
  - before: 3.5.1
```

spec/input/spec.yml
```diff
@@ -43,7 +43,7 @@ spec:
   - description: Folder containing field definitions
     type: folder
     name: fields
-    required: true
+    required: false
```
mrodm, line 46: For ECS input packages, there could be issues when running system tests if the ECS fields are not defined. 

Mappings and dynamic templates defined in `ecs@mappings` are not taken into account in that validation (in `elastic-package`). So, I think there will be fields reported as not defined for those input packages.

As those packages could still define the `fields` folder, there should not be any problem.

It needs to be tested also how `elastic-package` behaves when there is no `fields` folder in those input packages.

test/packages/good_input_otel/manifest.yml
mrodm, line 1: This test package is not part of the compliance tests.
It would be interesting to add this package to test that it can be installed at least.

https://github.com/elastic/package-spec/blob/18af7f05ce7186b13672b335b072c7e621ffb5fa/compliance/features/basic.feature

spec/input/spec.yml
mrodm: As in 3.5.0 the support for OTEL collector configuration was added into input packages looks good to me :+1: 

```
- version: 3.5.0
  changes:
    - ...
    - description: Add support for OTel collector configuration in input packages.
      type: enhancement
      link: https://github.com/elastic/package-spec/pull/975

File: spec/changelog.yml [hunk 3]
```diff
@@ -12,6 +12,11 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+- version: 3.5.1
```
mrodm, line 15: Nit

```suggestion
- version: 3.5.1-next
```

spec/input/spec.yml
jsoriano: > For ECS input packages, there could be issues when running system tests if the ECS fields are not defined.
> Mappings and dynamic templates defined in ecs@mappings are not taken into account in that validation (in elastic-package). So, I think there will be fields reported as not defined for those input packages.
> As those packages could still define the fields folder, there should not be any problem.

Yes, for these packages we still expect them to define `@timestamp`, `ecs.version` and so on. Though this wouldn't be currently required.

If the package has system tests, `elastic-package` will complain if these fields are not defined, if there are no system tests and no defined fields, this won't be detected, but shouldn't cause any issue (because of missing the usually required fields, there might be other issues undetected by the lack of tests...).

We may need to review this in `elastic-package`.

> It needs to be tested also how elastic-package behaves when there is no fields folder in those input packages.

Tests work with the `sql_input` test package in `elastic-package`, but there are test failures caused by missing fields definitions (what would be expected).

> Currently, `elastic-package create package` command to create an input package creates that `fields` folder:

We need to review `elastic-package create package` for OTel packages in general. For ECS input packages it would be fine if it creates the `fields` folder.

test/packages/good_input_otel/manifest.yml
jsoriano: Added in https://github.com/elastic/package-spec/pull/994/commits/d4e67cf47df2569dbffd248f073ece028a7b54cf.

PR #999, opened 2025-10-16
Add `conditions.agent.version` into integration and input manifest spec with validation

### LINKED ISSUE #970: Add a new `agent.version` condition in manifests
A new `agent.version` condition is introduced in package manifests in order to specify the minimum compatible version of Elastic Agent version.

Validation should be added if possible.

teresaromero:
## What does this PR do?

Adds a new property into the integration and input spec to add the minimum agent version a package is compatible with.

The new field is required after 3.6

## Why is it important?

The Elastic Agent running at Kibana, might not be compatible with the integration. This helps Fleet read the agent version an integration is compatible with and is able to warn the user if is not compatible.

## Checklist

- Added unit tests
- Updated test packages for 3.6 version to include the now required field

## Related issues

Resolves #970

### DIFF
--- a/code/go/pkg/validator/validator_test.go
+++ b/code/go/pkg/validator/validator_test.go
@@ -320,7 +320,7 @@ func TestValidateFile(t *testing.T) {
 		"bad_agent_version_v3": {
 			"manifest.yml",
 			[]string{
-				"field conditions: agent is required",
+				"field conditions.agent: version is required",
 			},
 		},
 	}
--- a/test/packages/bad_agent_version_v3/manifest.yml
+++ b/test/packages/bad_agent_version_v3/manifest.yml
@@ -14,6 +14,7 @@ conditions:
     capabilities:
       - observability
       - security
+  agent: {}
 vars:
   - name: package_password
     type: password


teresaromero:
test-integrations

code/go/internal/validator/semantic/validate_agent_version_required.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func getAgentVersionCondition(manifest pkgpath.File) (string, error) {
+	val, err := manifest.Values("$.conditions[\"agent.version\"]")
+	if err != nil {
+		val, err = manifest.Values("$.conditions.agent.version")
+		if err != nil {
+			return "", nil
+		}
+	}
```
jsoriano: We are not returning these errors, do they happen when the field is not found? Should we handle other cases? Though I see we do the same in other places.

code/go/internal/validator/semantic/validate_agent_version.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func getAgentVersionCondition(manifest pkgpath.File) (string, error) {
+	val, err := manifest.Values("$.conditions[\"agent.version\"]")
+	if err != nil {
+		val, err = manifest.Values("$.conditions.agent.version")
+		if err != nil {
+			return "", nil
+		}
+	}
```
jsoriano: Yeah, I would bring the check for requirement to the spec, and leave here only the validation of the format.

code/go/internal/validator/semantic/validate_agent_version.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return specerrors.ValidationErrors{specerrors.NewStructuredError(err, specerrors.UnassignedCode)}
+	}
+
+	agentVersionCondition, err := getAgentVersionCondition(*manifest)
+	if err != nil {
+		return specerrors.ValidationErrors{specerrors.NewStructuredError(err, specerrors.UnassignedCode)}
+	}
+
+	if _, err := semver.NewConstraint(agentVersionCondition); agentVersionCondition != "" && err != nil {
```
teresaromero: @jsoriano used here the validation of the format, if the value is empty it fails as it not compliant with semver contraint.

jsoriano (dismissed):

mrodm (dismissed):
Would it be possible to add a test package with `format_version: 3.6.0` failing ? 
Package manifest for that package could be like :
```yaml
conditions:
  agent: {}
```

PR #1000, opened 2025-10-17
Add validation for `template_path` at `input` packages

teresaromero:
## What does this PR do?

- Adds validation rule for `input` type packages regarding the `template_path` at the `policy_template`
- Changes the `input spec` to have the field `template_path` required

## Why is it important?

When an `input` package does not define its `template_path` at the `policy_templates` objects, fleet throws an error
[here](https://github.com/elastic/kibana/blob/293a1c0/x-pack/platform/plugins/shared/fleet/server/services/package_policy.ts#L3327) 

This is caused because, the stream created from the input [here](https://github.com/elastic/kibana/blob/293a1c0/x-pack/platform/plugins/shared/fleet/common/services/policy_template.ts#L99) has an undefined `template_path`. For inputs, there is no default value.

When a value is defined at `template_path`, `fleet` looks for the asset taking into account it ends with the given value [here](https://github.com/elastic/kibana/blob/293a1c0/x-pack/platform/plugins/shared/fleet/server/services/package_policy.ts#L3336) so it might be under the `agents/input` folder, or any other as long as it ends with the value. 

## Checklist

- Added unit tests for the validation rule

## Related issues

Resolves partially #703

### DIFF
--- a/code/go/internal/validator/semantic/validate_input_policy_template_template_path.go
+++ b/code/go/internal/validator/semantic/validate_input_policy_template_template_path.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	inputPackageType packageType = "input"
+	inputPackageType string = "input"
 )
 
 var (
@@ -29,15 +29,13 @@ var (
 	errInvalidPackageType    = errors.New("invalid package type")
 )
 
-type packageType string
-
 type inputPolicyTemplate struct {
 	Name         string `yaml:"name"`
 	TemplatePath string `yaml:"template_path"` // input type packages require this field
 }
 
 type inputPackageManifest struct { // package manifest
-	Type            packageType           `yaml:"type"`
+	Type            string                `yaml:"type"`
 	PolicyTemplates []inputPolicyTemplate `yaml:"policy_templates"`
 }
 


teresaromero:
test integrations

code/go/internal/validator/semantic/validate_input_policy_template_template_path.go
```diff
@@ -0,0 +1,100 @@
...
+
+type packageType string
+
+type inputPolicyTemplate struct {
+	Name         string `yaml:"name"`
+	TemplatePath string `yaml:"template_path"` // input type packages require this field
+}
+
+type inputPackageManifest struct { // package manifest
+	Type            packageType           `yaml:"type"`
```
mrodm: Could it be used `string` here instead of a custom type ?

```suggestion
	Type            string           `yaml:"type"`
```

code/go/internal/validator/semantic/validate_input_policy_template_template_path.go
```diff
@@ -0,0 +1,100 @@
...
+	"path"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+)
+
+const (
+	inputPackageType packageType = "input"
```
mrodm: Could it be used `string` instead?
```suggestion
	inputPackageType string = "input"
```

PR #1002, opened 2025-10-22
Add validation for `template_path` at integration packages

### LINKED ISSUE #703: [Linter] Ensure `stream.yml.hbs` exists if it will be fallen back on
Ref https://github.com/elastic/kibana/issues/151062
Ref https://github.com/elastic/integrations/pull/5267

In the case where a package references a stream that doesn't exist, the linter should error if the `{stream}.yml.hbs` file doesn't exist. Similarly, if no streams are defined, the linter should error if the `stream.yml.hbs` default file doesn't exist. 

This will prevent invalid packages from making their way to Fleet, though we'll also be adding some validation on Fleet's end to make sure this doesn't cause crashes.

teresaromero:
## What does this PR do?

This PR validates the existence of the `template_path` for integration packages.

## Why is it important?

Policy templates for an integration package have inputs references. This inputs have a property `template_path` which is optional. 
If the property is declared, then the file for this template has to be check under `agent/input/<template_path>`.
When the property is empty, the validation is deferred to the streams `input` `template_path`.

Any data stream, whose `streams` contain any `input` that matches the policy template input, has to check its `template_path` exist under /data_stream/<stream>/agent/stream. If the field `template_path` is empty a default value `streams.yml.hbs` is assigned to it.

Validation checks then that there is a file ending with `template_path`

This validation is mirroring what Fleet is doing with integration packages. A package can have inputs that dont match with data streams, in which case, there are no data streams to be checked, not input template_paths.

## Checklist

- added unit testing
- manual tests with fleet to verify this logic.

## Related issues

Partially resolves #703

### DIFF
--- a/code/go/internal/validator/semantic/validate_integration_policy_template_path.go
+++ b/code/go/internal/validator/semantic/validate_integration_policy_template_path.go
@@ -151,7 +151,15 @@ func validateInputWithStreams(fsys fspath.FS, input string, dsMap map[string]dat
 			_, err := fs.ReadFile(fsys, path.Join(dsDir, "agent", "stream", stream.TemplatePath))
 			if err != nil {
 				if errors.Is(err, fs.ErrNotExist) {
-					return errTemplateNotFound
+					// fallback to glob pattern matching in case the default template path is customized with a prefix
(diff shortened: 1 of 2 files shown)


mrodm (approved):
Thanks !! :rocket:

PR #1009, opened 2025-10-31
Fix template path validation including linked files

teresaromero:
## What does this PR do?

Includes a fix at the integration packages template path validation, taking into account the existence of linked files.
A template can be shared accross data streams by using a shared file and at the agent/stream file, place a `.link`.

The validation previously implemented did not take into account this occurence, as the iterator was looking for specific file names. It has been spotted while running the integration tests with `elastic-package`, as the test case for `with_links` package was correctly mapped. 

Why it was not detected at `package-spec` tests? The package representing a package with link files was not taking into account that streams are linked with the manifest policy template declaration. This PR modifies the package so the test is reliable and spots this occurance.

## Why is it important?

When linked files are used to share data streams templates, the validation should be able to check if the file exists, and the final "included" file exists too.

The fs handler already has embedded the link files system, so when a file is `.link` it looks for the included directly. 
https://github.com/elastic/package-spec/blob/1bd98cfc1b9ba37928c7e246f6fa10b496ee9fbd/code/go/internal/linkedfiles/fs.go#L45

## Checklist

tested replacing package-spec at elastic-package and testing the error

`replace github.com/elastic/package-spec/v3 v3.5.0 => ../package-spec
`

## Related issues

Related #703 
Bug introduced at #1002

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -17,7 +17,7 @@
       link: https://github.com/elastic/package-spec/pull/999
 - version: 3.5.2
   changes:
-    - description: Fix validation of integration package policy_template paths with linked files.
+    - description: Fix validation of policy_template paths with linked files.
       type: bugfix
       link: https://github.com/elastic/package-spec/pull/1009
 - version: 3.5.1


code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -168,3 +170,33 @@ func validateInputWithStreams(fsys fspath.FS, input string, dsMap map[string]dat
 
 	return nil
 }
+
+// findPathWithPattern looks for a file matching the templatePath in the given data stream directory
+// It checks for exact matches, files ending with the templatePath, or templatePath + ".link"
+func findPathWithPattern(fsys fspath.FS, dsDir, templatePath string) (string, error) {
```
mrodm: Another option here is using directly the `fsys` that will iterate over the linked files already "rendered".

https://github.com/elastic/package-spec/blob/b9e2c0484b01e2ca44fc8d1d5847ec8a8ac17855/code/go/pkg/validator/validator.go#L53-L62

So maybe it could be used `fs.WalkDir` and look for the template paths without checking explictely for `*.link`. Maybe something like this (it needs to be tested)?

```golang
func findPathWithPattern(fsys fspath.FS, dsDir, templatePath string) (string, error) {
	mainPath := path.Join(dsDir, "agent", "stream")

	foundFile := ""
	err := fs.WalkDir(fsys, mainPath, func(p string, d fs.DirEntry, walkErr error) error {
		if walkErr != nil {
			return walkErr
		}
		if d.IsDir() {
			return nil
		}
		base := filepath.Base(p)
		if base == templatePath || strings.HasSuffix(p, templatePath) {
			foundFile = base
			return fs.SkipAll // found a match, stop walking
		}
		return nil
	})
	if err != nil {
		return "", err
	}

	if foundFile == "" {
		return "", errTemplateNotFound
	}

	return foundFile, ""
}
```

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -168,3 +170,33 @@ func validateInputWithStreams(fsys fspath.FS, input string, dsMap map[string]dat
...
+	// 2. Ends with stream.TemplatePath
+	// 3. Equals stream.TemplatePath + ".link"
+	var foundFile string
+	for _, match := range matches {
+		base := filepath.Base(match)
+		if base == templatePath || base == templatePath+".link" {
+			foundFile = match
+			break
+		}
+		// fallbak to check for suffix match, in case the path is prefixed
```
mrodm: Nit
```suggestion
		// fallback to check for suffix match, in case the path is prefixed
```

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -148,18 +150,18 @@ func validateInputWithStreams(fsys fspath.FS, input string, dsMap map[string]dat
 				stream.TemplatePath = defaultStreamTemplatePath
 			}
 
-			_, err := fs.ReadFile(fsys, path.Join(dsDir, "agent", "stream", stream.TemplatePath))
+			foundFile, err := findPathWithPattern(fsys, dsDir, stream.TemplatePath)
```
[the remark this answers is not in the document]
teresaromero: I've updated the PR with this change. Made the finder function more broad with the dir param instead of data stream dir. Updated also test cases

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -148,18 +150,19 @@ func validateInputWithStreams(fsys fspath.FS, input string, dsMap map[string]dat
...
+			if err != nil {
+				return err
+			}
+			if foundFile == "" {
+				return errTemplateNotFound
+			}
+
+			_, err = fs.ReadFile(fsys, foundFile)
 			if err != nil {
 				if errors.Is(err, fs.ErrNotExist) {
```
jsoriano: Nit. This should not happen, right? Though we can leave this in case the file is somehow removed between finding and opening it.

jsoriano (commented):
LGTM, added a couple of small comments.

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
teresaromero: this was supposed to be removed, my bad i did not remove it when changing the code.

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -168,3 +171,33 @@ func validateInputWithStreams(fsys fspath.FS, input string, dsMap map[string]dat
 
 	return nil
 }
+
+// findPathWithPattern looks for a file matching the templatePath in the given directory (dir)
+// It checks for exact matches, files ending with the templatePath, or templatePath + ".link"
+func findPathWithPattern(fsys fspath.FS, dir, templatePath string) (string, error) {
```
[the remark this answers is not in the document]
teresaromero: i got the point for renaming, but was not sure about the suffix suggestion. The function also matches exact without being a suffix... i've renamed it to findPathAtDirectory... as the files should be inside the dir param ... WDYT?

spec/changelog.yml
```diff
@@ -15,6 +15,11 @@
     - description: Require defining agent version constraints in input and integration packages.
       type: breaking-change
       link: https://github.com/elastic/package-spec/pull/999
+- version: 3.5.2
+  changes:
+    - description: Fix validation of integration package policy_template paths with linked files.
```
mrodm: Nit.
This involves integration and input packages. WDYT ?

```suggestion
    - description: Fix validation of policy_template paths with linked files.
```

mrodm (approved):
LGTM!
Just a minor comment related to the changelog entry description.

jsoriano (approved):

mrodm (approved):

PR #1013, opened 2025-11-11
Allow to define the message field in an array when removing for event original

jsoriano:
## What does this PR do?

There is a required processor for removing `message` when handling `event.original`, if the `message` field is listed in an array, the processor should be equally valid.

## Why is it important?

This syntax serves the same purpose, and avoids issues when scripting generation of pipelines.

## Checklist

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -15,7 +15,7 @@
     - description: Require defining agent version constraints in input and integration packages.
       type: breaking-change
       link: https://github.com/elastic/package-spec/pull/999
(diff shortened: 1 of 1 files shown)


spec/changelog.yml
```diff
@@ -15,6 +15,11 @@
     - description: Require defining agent version constraints in input and integration packages.
       type: breaking-change
       link: https://github.com/elastic/package-spec/pull/999
+- version: 3.5.4-next
```
```suggestion
- version: 3.5.3-next
```

are we skipping a patch?

PR #1024, opened 2025-11-20
Fix rule for processors handing event.original so the check can be skipped

jsoriano:
In #1013 the spec for the processors required for event.original
handling was modified to allow to remove the message field when included
in an array. This modified the error message by adding the index in the
array, what doesn't match anymore with the rule that allows to skip this
test.
    
Update the rules to match with this new pattern of error messages.

Another rule is added to allow to skip this check in pipelines that only
have a remove processor, that doesn't have `if` or `ignore_missing`
properties.

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -19,7 +19,7 @@
   changes:
     - description: Fix rule matching processor for event.original handling.
       type: bugfix
-      link: https://github.com/elastic/package-spec/pull/1013
+      link: https://github.com/elastic/package-spec/pull/1024/
 - version: 3.5.3
   changes:
     - description: Make support for alert rule templates more permissive.


code/go/internal/validator/spec.go
```diff
@@ -127,13 +127,17 @@ func processErrors(errs specerrors.ValidationErrors) specerrors.ValidationErrors
...
 			new:     "%s: rename \"message\" to \"event.original\" processor requires remove \"message\" processor",
 		},
 		{
 			matcher: regexp.MustCompile(`(processors.[0-9]+.remove.if): processors.[0-9]+.remove.if does not match: "ctx\.event\?\.original != null"`),
 			new:     "%s: rename \"message\" to \"event.original\" processor requires remove \"message\" processor with if: 'ctx.event?.original != null'",
 		},
+		{
+			matcher: regexp.MustCompile(`(field processors.[0-9]+.remove): (ignore_missing|if) is required`),
+			new:     "%s: rename \"message\" to \"event.original\" processor requires remove \"message\" processor with if: 'ctx.event?.original != null'",
+		},
```
jsoriano: This rule is added to allow to ignore the error in pipelines that only have a `remove` processor and it doesn't have the `ignore_missing` or `if` properties. This is the case of the pipeline added here for testing.

Other pipelines are not failing because they use to contain some other remove with these parameters, and if the `if` doesn't contain `'ctx.event?.original != null'`, it is captured by the previous rule above.

jsoriano:
test integrations

PR #1042, opened 2025-12-16
Validate kibana tags duplicates

### LINKED ISSUE #977: [Change Proposal] Add validation to avoid duplicate tags
Packages have two ways of defining Kibana tags:
* Directly as a saved object under `kibana/tag`. These are used for package-specific tags.
* Using the `kibana/tags.yml` file, that is intended for shared tags, and includes rules about what saved objects to reference with this tag. It is created when needed if it doesn't exist yet.

Packages can currently define the same tag using both methods, what leads to having the same tag multiple times, for example for Crowdstrike:
* Defined as a package-specific tag (what manages `crowdstrike-security-solution-default`): https://github.com/elastic/integrations/blob/7414f0cee3cdb60cdbef9f5895556640fe57f916/packages/crowdstrike/kibana/tag/crowdstrike-security-solution-default.json#L5
* Defined in `kibana/tags.yml` (what manages the tagging of saved objects with `security-solution-default`: https://github.com/elastic/integrations/blob/7414f0cee3cdb60cdbef9f5895556640fe57f916/packages/crowdstrike/kibana/tags.yml#L1

Not sure, but it might be also possible to duplicate a tag if defined twice with the same name but different ids under `kibana/tag`.

Some tasks here:
* [ ] Add a validation rule that ensures that each tag is defined only once in each package. Two tags are the same when defined in a package if they have the same name, even if they have different ids.
  Shared tags should have preference, so if a package is defined under `kibana/tag` and `tags.yml`, the error should point to the definition in `kibana/tag`.
* [ ] When exporting resources with `elastic-package`, avoid creating tags under `kibana/tag` if they already exist in `kibana/tags.yml`, update if needed to include ids.

teresaromero:
## What does this PR do?

This PR implements validation into kibana tags assets. When a tag is defined inside the `tags.yml` is duplicated and an asset is created when exporting the dashboard with `elastic-package`.

## Why is it important?

Avoid duplicate tags at kibana UI, as .json assets when is already declared at `tags.yml`

## Checklist

## Related issues

- Resolves #977

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -21,6 +21,11 @@
     - description: Add support for Kibana `slo_template` assets
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1037
+- version: 3.5.6-next
+  changes:
+    - description: Add validation for Kibana tag duplicates.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1042
 - version: 3.5.5
   changes:
     - description: Run validation semantic rules also in transform fields.
@@ -29,9 +34,6 @@
     - description: Run static validation for handlebars templates in input and integration packages.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1030
-    - description: Add validation for Kibana tag duplicates.
-      type: enhancement
-      link: https://github.com/elastic/package-spec/pull/1042
 - version: 3.5.4
   changes:
     - description: Fix rule matching processor for event.original handling.


code/go/internal/validator/semantic/validate_kibana_tag_duplicates.go
```diff
@@ -0,0 +1,126 @@
...
+	err := fs.WalkDir(fsys, tagDir, func(filePath string, d fs.DirEntry, walkErr error) error {
+		if walkErr != nil {
+			if !errors.Is(walkErr, fs.ErrNotExist) {
+				return walkErr
+			}
+			return nil
+		}
+		if d.IsDir() || path.Ext(filePath) != ".json" {
+			return nil
+		}
```
jsoriano: Nit. Could this be simplified by using `Glob`? Or even `Readdir`, as we are only interested on the content of a directory.

code/go/internal/validator/semantic/validate_kibana_tag_duplicates.go
```diff
@@ -0,0 +1,126 @@
...
+			continue
+		}
+		tagMap[tag.Name] = struct{}{}
+	}
+	return tagMap, errs
+}
+
+func validateKibanaJSONTags(fsys fspath.FS, tagMap map[string]struct{}) specerrors.ValidationErrors {
+	// Collect all tags used in the package.
+	tagDir := path.Join("kibana", "tag")
```
jsoriano: Can tags be embedded directly in dashboards? 🤔 Probably don't, but it would be worth checking.

jsoriano (commented):
Overall approach LGTM 👍

code/go/internal/validator/semantic/validate_kibana_tag_duplicates.go
jsoriano: There are elements that can be embedded in dashboards, but they are mostly the panels and the searches. Probably tags can only be included by reference. Thanks for checking :+1:

teresaromero:
test integrations

code/go/internal/validator/spec.go
```diff
@@ -226,6 +226,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateIntegrationPolicyTemplates, types: []string{"integration"}},
 		{fn: semantic.ValidatePipelineTags, types: []string{"integration"}, since: semver.MustParse("3.6.0")},
 		{fn: semantic.ValidateStaticHandlebarsFiles, types: []string{"integration", "input"}},
+		{fn: semantic.ValidateKibanaTagDuplicates, types: []string{"integration"}},
```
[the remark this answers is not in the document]
jsoriano: > i am guessing at least all the json corresponding to the security solution tag can be removed (safely)

Yes, but we would also need to remove the references to these ids from the dashboards, right?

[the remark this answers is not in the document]
teresaromero: yes, the json and the reference to that id in the dasboards or any other object where they are referenced

code/go/internal/validator/semantic/validate_kibana_tag_duplicates.go
```diff
@@ -0,0 +1,134 @@
...
+				fmt.Errorf("file \"%s\" is invalid: duplicate tag name '%s' found", fsys.Path(tagsPath), tag.Name), specerrors.CodeKibanaTagDuplicates))
+			continue
+		}
+		tags = append(tags, tag.Name)
+	}
+	return tags, errs
+}
+
+func validateKibanaPackageTagsDuplicates(fsys fspath.FS, sharedTagNames []string) specerrors.ValidationErrors {
+
```
jsoriano: Nit.
```suggestion
```

jsoriano (approved):

mrodm (dismissed):

PR #1043, opened 2025-12-29
Prepare release 3.5.5

teresaromero:
- remove `next` from changelog 3.5.5
- update version for compliance test


spec/changelog.yml
jsoriano: We can remove this line as this is done.

jsoriano (dismissed):

PR #1053, opened 2026-01-08
Support `deprecated` packages or individual features

### LINKED ISSUE #1032: Add support for new 'deprecated' field
Add support for a new field that will allow package authors to deprecate an entire package or a part of it (policy template/input/datastream). The deprecation lifecycle has been discussed in this [RFC](https://docs.google.com/document/d/1daDzfoxlsPM-s9sMy0kizUcB2Ly91mUXORqw_uYy1Ys/edit?tab=t.0).

``` yml
deprecated:
   since: "2.2.0"
   description: "this package is deprecated, use XYZ instead"
      replaced_by: 
        package: "xyz"

```

- `since` and `description` mandatory fields
- `replaced_by` is optional and can be used if the authors want to indicate explicitly the replacement. Its content should match the package name or the input/datastream name

##  Entire package

In this case the fields will be added to the manifest:
``` yml
format_version: 3.1.5
name: log
title: Custom Logs (Deprecated)
description: >-
  Collect custom logs with Elastic Agent.
type: input
version: 2.4.4
deprecated:
    description: "This package is deprecated, please use the Custom Logs (Filestream) package instead."
    since: "vX.Y.Z"
```

## Individual feature
The deprecated field will need to appear at the input or data stream level.  Using as an example an [input from o365](https://github.com/elastic/integrations/blob/d11374338787fde1d6f15c1de8dde3759efac227/packages/o365/manifest.yml#L51-L53) that is currently deprecated:

``` yml
policy_templates:
  - name: o365
    title: Office 365 logs
    description: Collect logs from Office 365
    inputs:
      - type: o365audit
      title: "Collect audit logs"
      description: "This option collects audit logs using the Management Activity API through a deprecated method"
      deprecated:
        description: "This input is deprecated. Please use the CEL input instead." 
         replaced_by: 
	   package: "o365"
           input: "cel"

```

In the case of a data stream, the deprecated field would appear at the variable level inside the [manifest file](https://github.com/elastic/integrations/blob/f75ddbefe1c3a95648ba13b8f16441404874574b/packages/gcp/data_stream/cloudrun_metrics/manifest.yml#L14-L20) that contains the data stream:

``` yml
title: "GCP Redis Metrics"
type: metrics
streams:
  - input: gcp/metrics
    title: GCP Redis Metrics
    description: Collect GCP Redis Metrics
    vars:
      - name: zone
        ...
      - name: region
        type: text
        title: GCP Region
        description: deprecated, use regions instead
        multi: false
        required: false
        show_user: true
        deprecated:
        	description: "This variable is deprecated. Please use use `regions` instead." 
```

Same for policy template deprecations, the deprecation will be at that level.

teresaromero:
## What does this PR do?

This PR implements support for the new `deprecated` field in package specifications, enabling package authors to mark entire packages or individual features (policy templates, inputs, data streams and variables) as deprecated.

1. Added `deprecated` field to specifications:
   - Package-level: content, input and integration manifest specs
   - Policy template level: input and integration manifest spec
   - Inputs within policy templates at integrations manifest
   - Data stream and Data stream variables
2. Added deprecation as a new changelog type alongside enhancement, bugfix, and breaking-change
3. Implemented semantic validation that:
    - ensures if all inputs in an integration are deprecated, the package itself must also be marked as deprecated
    - ensures replaced_by object is aligned with feature deprecated
5. Version compatibility: Added JSON patches to remove the deprecated field for format_version < 3.6.0

## Why is it important?
This feature provides a standardized mechanism for package authors to communicate deprecation information to users, following the deprecation lifecycle defined in the [RFC](https://docs.google.com/document/d/1daDzfoxlsPM-s9sMy0kizUcB2Ly91mUXORqw_uYy1Ys/edit?tab=t.0)

## Checklist

## Related issues
- Closes #1032
- Related to https://github.com/elastic/package-registry/pull/1516

### DIFF
--- a/code/go/internal/validator/semantic/validate_deprecated_replaced_by.go
+++ b/code/go/internal/validator/semantic/validate_deprecated_replaced_by.go
@@ -14,8 +14,8 @@ import (
 )
 
 type deprecatedInfo struct {
-	Since       string `yaml:"since,omitempty"`
-	Description string `yaml:"description,omitempty"`
+	Since       string `yaml:"since"`
+	Description string `yaml:"description"`
 	ReplacedBy  *struct {
 		Package        string `yaml:"package,omitempty"`
 		PolicyTemplate string `yaml:"policy_template,omitempty"`


code/go/internal/validator/semantic/validate_integration_inputs_deprecated.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+)
+
+// ValidateIntegrationInputsDeprecation checks that if all inputs in an integration package are deprecated,
+// then the integration package itself must be marked as deprecated.
+func ValidateIntegrationInputsDeprecation(fsys fspath.FS) specerrors.ValidationErrors {
```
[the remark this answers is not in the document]
teresaromero: indeed i forgot to add it!! 
updated in last commit, added bas case test package too. thnks

test/packages/deprecated_integration/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.0.1"
+  changes:
+    - description: Initial draft of the package
+      type: enhancement
```
mrodm: This changelog entry type could be set as `deprecation` to test also that this value is valid here. At least in one package, I think it is no need to change it in all the test packages:

```suggestion
      type: deprecation
```

mrodm:
> I've not included the `replaced_by` field, as i am not clear if this is included in the first iteration @criamico @jsoriano

I see that in the issue description is added that field in the examples too. I guess it could be added now even if it is not going to be used currently, WDYT @criamico ?

It looks like that depending on the item to be deprecated, that `replaced_by` field is going to contain different fields:

Package
```yaml
deprecated:
    description: "This package is deprecated, please use the Custom Logs (Filestream) package instead."
    since: "X.Y.Z"
    replaced_by: 
      package: "filestream"
```

Input:
```yaml
policy_templates:
  - name: o365
    title: Office 365 logs
    description: Collect logs from Office 365
    inputs:
      - type: cel
        title: "Collect audit logs"
        description: "Collect audit logs using the Management Activity API"
      - type: o365audit
        title: "Collect audit logs"
        description: "This option collects audit logs using the Management Activity API through a deprecated method"
        deprecated:
          description: "This input is deprecated. Please use the CEL input instead." 
          since: X.Y.Z
          replaced_by: 
            package: "o365"
            input: "cel"
```

Following that, other deprecated elements like variable and policy template, would they need a new field too under `replaced_by` ? @criamico @jsoriano

criamico:
> I've not included the replaced_by field, as i am not clear if this is included in the first iteration

I think that it can be added later. This PR is already big so @teresaromero is up to you :)
if you don't add it now, please let's create a follow up ticket to avoid losing track of it. 


> other deprecated elements like variable and policy template, would they need a new field too under replaced_by ?

I think that it's possible that an integration developer decides to replace a var with another or a policy template with another. @jsoriano could you confirm?

teresaromero:
i've added as [last commit](https://github.com/elastic/package-spec/pull/1053/commits/6760df6c6d94777e8adb52a8d5c96f12ac8c9bb4) the replaced_by implementation. Although i though about doing it as a follow up PR to split the code i finally included this into this active PR.

Each deprecation should be replaced by its own "type", i've added semantic validation to make sure it does if the field is used, as the field is optional. 

cc @criamico @mrodm

spec/input/manifest.spec.yml
mrodm: Looking at the input manifest, there is a property to define policy templates. I guess it should be added there also the deprecated field.

spec/changelog.yml
```diff
@@ -24,6 +24,9 @@
     - description: Add pipeline global on_failure validations.
       type: breaking-change
       link: https://github.com/elastic/package-spec/pull/1038
+    - description: Add support for deprecating packages or individual features (datastreams, inputs, policies).
```
mrodm: Should it be policy templates, inputs and variables ?
Currently, in the data stream manifest there is no deprecated field to deprecated that. There is just the addition to deprecate variables.

Taking into account the issue description:
> The deprecated field will need to appear at the input or data stream level. Using as an example an [input from o365](https://github.com/elastic/integrations/blob/d11374338787fde1d6f15c1de8dde3759efac227/packages/o365/manifest.yml#L51-L53) that is currently deprecated:

Should it be able to deprecate a data stream too ? @criamico 

Something like (taken [this data stream as example](https://github.com/elastic/integrations/blob/d11374338787fde1d6f15c1de8dde3759efac227/packages/o365/data_stream/audit/manifest.yml#L1))

```
type: logs
title: Microsoft Office 365 audit logs
deprecated:
  since: x.y.z
  description: text
streams:
  - input: cel
    ...
  - input: o365audit
    ...
```

criamico: @mrodm Yes the datastream deprecation is a valid use case, in that case the `deprecated` field should go into the datastream manifest as you said.

spec/input/manifest.spec.yml
teresaromero: yes, i missed the policy template deprecation at input packages. i will add it. my doubt here would be that, in the case of integrations, if all inputs are deprecated -> the package should be marked as deprecated. so in case of an input, does this rule apply to policy_templates ?

mrodm (approved):
LGTM, thanks @teresaromero !

Issue #1062, opened 2026-01-21
[slo templates] allow packages targeting 8.x to ship with SLO templates


tommyers-elastic:
we want to be able to add SLO templates to packages without bumping the minimum required kibana version to 9+. for kibana versions which do not support slo templates, they should be silently ignored at install. 

we had the same requirement for rule templates, and we solved it by relaxing the package spec constraints for inclusion of the rules: https://github.com/elastic/package-spec/pull/1012

can we do the same for SLO templates?

tommyers-elastic:
@jsoriano do you see any issues with this?

tommyers-elastic:
@agithomas is going to double check

PR #1064, opened 2026-01-28
Relax package-spec constraints for slo templates

agithomas:
## What does this PR do?

To have the ability to add SLO templates to packages without bumping the minimum required kibana version to 9+. For kibana versions that do not support slo templates, they should be silently ignored at install.

We had the same requirement for rule templates, and is solved by relaxing the package spec constraints for inclusion of the rules: https://github.com/elastic/package-spec/pull/1012

## Why is it important?

With this change, integration packages can receive slo assets under slo_templates

## Checklist

## Related issues

- Relates https://github.com/elastic/package-spec/pull/1012
- Closes https://github.com/elastic/package-spec/issues/1062

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -18,23 +18,23 @@
     - description: Add support for ES|QL query assets in integration and content packages.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1028
-    - description: Add support for Kibana `slo_template` assets
-      type: enhancement
-      link: https://github.com/elastic/package-spec/pull/1037
     - description: Add pipeline global on_failure validations.
       type: breaking-change
       link: https://github.com/elastic/package-spec/pull/1038
     - description: Add support for deprecating packages or individual features (policy_templates, inputs, data_streams or variables).
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1053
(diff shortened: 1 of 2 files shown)


agithomas:
++ @kdelemme, who worked on the [PR](https://github.com/elastic/package-spec/pull/1037/), to highlight any issues or concerns.

spec/integration/kibana/spec.yml
```diff
@@ -144,12 +144,10 @@ spec:
         contentMediaType: "application/json"
         pattern: '^.+\.json$'
 versions:
-  - before: 3.6.0
+  - before: 3.4.0
```
mrodm: Should it be updated also the content packages ? 

https://github.com/elastic/package-spec/blob/c86e7f463bfff03c073f454d8c7d1f0fa5b0050f/spec/content/kibana/spec.yml#L60-L67

Currently, in content packages alert rule templates just can be used from 3.5.0 onwards. It was not changed as part of the #1012 

What should we do with slo templates there ? Removed that support for specs `<3.5.0` as alert rule templates ? 

cc @jsoriano

mrodm (approved):

kdelemme (approved):

teresaromero (approved):
thanks

Issue #1023, opened 2025-11-20
[Change Proposal] Support OTel input packages that collect data with multiple signal types


jsoriano:
Package policies for input packages are currently designed to collect a single signal type, and send it to a single data stream.

In the case of OTel, a single receiver can collect multiple signal types if it is connected to pipelines of different types, as seen in https://github.com/elastic/integrations/pull/16003

### Changes in package-spec
- [ ]  Define a new field `available_types` in the manifest that allows the author to specify the signal types available for the package, for instance:
``` yml
available_types: [logs, metrics, traces]
```

### Changes in Fleet

- [ ] Generate a different transform for each signal defined in `available_types` - example in case of all three signals present:

``` yml
processors:
  transform/foo
    log_statements: 
      - set(attributes["data_stream.type"], "logs")
      - set(attributes["data_stream.dataset"], "<policy_template_name>")
      - set(attributes["data_stream.namespace"], "<namespace>")
    metric_statements: 
      - set(attributes["data_stream.type"], "metrics")
      - set(attributes["data_stream.dataset"], "<policy_template_name>")
      - set(attributes["data_stream.namespace"], "<namespace>")
    trace_statements: 
      - set(attributes["data_stream.type"], "traces")
      - set(attributes["data_stream.dataset"], "<policy_template_name>")
      - set(attributes["data_stream.namespace"], "<namespace>")
 ```

- [ ] in case of for OTel input packages, update the UI element in "advanced options" that allows to select the datastream type so that the radio buttons for data stream type conform to the configured available_types.
e.g. a user can toggle between logs or metrics when `available_types: [logs, metrics]` is set, but can't select traces

<img width="505" height="314" alt="Image" src="https://github.com/user-attachments/assets/a34934a6-bdcc-4ad1-8485-e027bba182b1" />

The changes can be tested by compiling the test package defined in https://github.com/elastic/integrations/pull/16003 with the new `available_types` field.

<details>
  <summary>Original description</summary>
We have to design how this should work, as this has some implications, between them:
* There are currently UI elements to select the data stream type, should they be disabled for OTel input packages with multiple pipelines?
* Should support for multiple signal types in a package detected automatically or indicated by some flag in the manifest?
* How would routing work? Using fixed indexes in the exporter, different transforms for different dynamic mappings? Routing on ingestion? Should this be configurable per package?
...
</details>

cmacknz:
[...]

PR #1067, opened 2026-02-02
Define new field dynamic_sygnal_types for otel input packages

criamico:
## What does this PR do?

Add a new manifest field `dynamic_signal_types`  under policy_template.

## Why is it important?

Enable otelcol input packages to collect multiple signal types. `dynamic_signal_types` is a boolean field:
- Default behavior is false; the otel input package works as usual, collecting only one signal type define in the configuration;
- When enabled, the input package is able to collect all signal types defined under `config.pipelines`

## Checklist

## Related issues

- Relates https://github.com/elastic/package-spec/issues/1023
- Related Kibana PR: https://github.com/elastic/kibana/pull/250081
- Supersedes https://github.com/elastic/package-spec/pull/1061

### DIFF
--- a/spec/input/manifest.spec.yml
+++ b/spec/input/manifest.spec.yml
@@ -136,13 +136,13 @@ versions:
   - before: 3.6.0
     patch:
       - op: remove
+        path: "/properties/policy_templates/items/properties/dynamic_signal_types" # removes dynamic_signal_types field for policy templates
+      - op: remove
         path: "/properties/elasticsearch/properties/index_mode" # Fleet needs to check that it is not enabled in non-metrics data streams
       - op: remove
         path: "/properties/deprecated" # removes deprecated field for package
       - op: remove
         path: "/properties/policy_templates/items/properties/deprecated" # removes deprecated field for policy templates
-      - op: remove
-        path: "/properties/policy_templates/items/properties/dynamic_signal_types" # removes dynamic_signal_types field for policy templates
   # Reserve otelcol input name before 3.5.0.
   - before: 3.5.0
     patch:


code/go/internal/validator/semantic/validate_input_dynamic_signal_types_test.go
```diff
@@ -0,0 +1,144 @@
...
+		require.NoError(t, err)
+
+		errs := ValidateInputDynamicSignalTypes(fspath.DirFS(d))
+		require.NotEmpty(t, errs, "expected validation errors")
+		assert.Len(t, errs, 1)
+		assert.Contains(t, errs[0].Error(), "dynamic_signal_types is only allowed when input is 'otelcol'")
+		assert.Contains(t, errs[0].Error(), "got 'logfile'")
+	})
+
+	t.Run("valid_non_input_package_skipped", func(t *testing.T) {
```
teresaromero: i don`t think this case will reach the semantic validation. If an integration contains this field, the error is coming from the json spec file validation. however, for the function to be semantic validation, i feel we could interpret if an integration policy_template has the field, throw an error as this is not allowed, if i understood correctly? so firstly check the field under policy_template, and then validate is an input package, if not throw an error. WDYT?

code/go/internal/validator/semantic/validate_input_dynamic_signal_types.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, policyTemplate := range manifest.PolicyTemplates {
+		// Only check if dynamic_signal_types is explicitly set to true
+		if policyTemplate.DynamicSignalTypes != nil && *policyTemplate.DynamicSignalTypes {
+			if policyTemplate.Input != otelcolInputType {
+				errs = append(errs, specerrors.NewStructuredErrorf(
+					"file \"%s\" is invalid: policy template \"%s\": dynamic_signal_types is only allowed when input is 'otelcol', got '%s'",
+					fsys.Path(manifestPath), policyTemplate.Name, policyTemplate.Input))
+			}
+		}
+	}
```
teresaromero:
```suggestion
	

	for _, policyTemplate := range manifest.PolicyTemplates {
		// Only check if dynamic_signal_types is explicitly set to true
		if policyTemplate.DynamicSignalTypes != nil && *policyTemplate.DynamicSignalTypes {
		
		// first validate is an input package
		if manifest.Type != inputPackageType {
		   errs = append(errs, specerrors.NewStructuredErrorf(
					"file \"%s\" is invalid: policy template \"%s\": dynamic_signal_types is only allowed for input type, got '%s'",
					fsys.Path(manifestPath), manifest.Type))
					continue
	     }
		
			if policyTemplate.Input != otelcolInputType {
				errs = append(errs, specerrors.NewStructuredErrorf(
					"file \"%s\" is invalid: policy template \"%s\": dynamic_signal_types is only allowed when input is 'otelcol', got '%s'",
					fsys.Path(manifestPath), policyTemplate.Name, policyTemplate.Input))
			}
		}
	}
```

spec/input/manifest.spec.yml
```diff
@@ -88,6 +88,10 @@ spec:
           template_path:
             description: "Path to Elasticsearch index template for stream."
             type: string
+          dynamic_signal_types:
+            description: "When enabled, decides the transforms and index templates that need to be created depending on the pipelines specified in the configuration."
+            type: boolean
+            default: false
```
teresaromero: when using default, the pointer will always be set, is this expected?

jsoriano (dismissed):

Issue #1068, opened 2026-02-03
[Change Proposal] Support `_dev` directory for content-only packages


kpollich:
There are various developer-only artifacts that can be useful for integration development, and there's no real reason we exclude these from content-only packages at the moment. 

We should allow the `_dev` directory to exist for content-only packages. 

See https://github.com/elastic/integrations/pull/16969 for an example of this being disallowed today. This PR uses some experimental "dashboards as yml" files that don't actually impact the content pack's usage in Fleet/Kibana. This should be supported.

PR #1077, opened 2026-02-11
Add support for `_dev` directory in content packs + support for Kibana dashboards as YML in `_dev/build` directory

kpollich:
## What does this PR do?

Adds support for 

- `_dev/` directory existing in content-only packages
- `_dev/shared` as a home for dashboards-as-YML files in content-only packages

## Why is it important?

Allows experimental dashboards-as-code assets to be included with integrations such as those in https://github.com/elastic/integrations/pull/17358

## Checklist

## Related issues

Closes https://github.com/elastic/package-spec/issues/1068

_Generated (mostly) with Cursor_

### DIFF
--- a/spec/content/spec.yml
+++ b/spec/content/spec.yml
@@ -68,10 +68,6 @@ versions:
     patch:
       - op: remove
         path: "/contents/7" # remove elasticsearch spec
-  - before: 3.5.7
-    patch:
-      - op: remove
-        path: "/contents/7" # remove _dev folder spec
   - before: 3.3.1
     patch:
       - op: add


spec/integration/_dev/build/spec.yml
```diff
@@ -10,4 +10,9 @@ spec:
...
       name: docs
       required: false
-      $ref: "./docs/spec.yml"
\ No newline at end of file
+      $ref: "./docs/spec.yml"
+    - description: Folder containing Kibana dashboards as YML files
+      type: folder
+      name: kibana
+      required: false
+      $ref: "./kibana/spec.yml"
```
jsoriano: Do we need to create a new place for dashboards? How are these dashboards going to be used?

`elastic-package build` already does some processing of the dashboards under `kibana/dashboards`, they are not used as they are. We could add a step to convert from YAML to JSON.

Converting from YAML to JSON is something done also in other places as in ingest pipelines, that are defined as YAML for human friendliness.

jsoriano: Oh I see, it would be great to make `elastic-package` able to compile this format.

As this seems to be something temporary, maybe we could use a more general-purpose directory instead of one specific for this.

We already have the `_dev/shared` directory under `_dev`, that can be used to store any kind of file. In principle it was added to store files that can be reused among packages, but it could be really used for anything.

I would suggest to add support for `_dev/shared` in content packages. Or if we don't want to extend the responsibilities of this directory, create another directory intended to store any kind of development file. Maybe `_dev/sources`, if this is intended to store source files intended to generate other files with other tools.

spec/content/spec.yml
```diff
@@ -51,6 +51,12 @@ spec:
       name: "validation.yml"
       required: false
       $ref: "../integration/validation.spec.yml"
+    - description: Folder containing development resources
+      type: folder
+      name: _dev
+      required: false
+      visibility: private
+      $ref: "../integration/_dev/spec.yml"
```
jsoriano: This is importing the whole `_dev` from integration packages.

By now we should create a new `spec.yml`.
```suggestion
      $ref: "./_dev/spec.yml"
```

That imports only the definition for `_dev/shared`:
```yaml
spec:
  additionalContents: false
  developmentFolder: true
  contents:
    - description: Folder containing shared files.
      type: folder
      name: shared
      required: false
      $ref: "../integration/_dev/shared/spec.yml"
```

kpollich: Updated in https://github.com/elastic/package-spec/pull/1077/commits/d3f4f9752aef003db1cb93c16e9770b94a380ae6

code/go/pkg/validator/validator_test.go
```diff
@@ -440,6 +445,20 @@ func TestValidateFile(t *testing.T) {
...
+				require.True(t, ok)
+				var errMessages []string
+				for _, vErr := range vErrs {
+					errMessages = append(errMessages, vErr.Error())
+				}
+				devPath := filepath.Join(pkgRootPath, "_dev")
+				expectedMsg := "item [deploy] is not allowed in folder [" + devPath + "]"
+				require.Contains(t, errMessages, expectedMsg)
+				return
+			}
```
jsoriano: I think this test package could be moved to `TestValidateItemNotAllowed`, so this custom code is not needed.
```suggestion
```

It would be added as:
```go
                "bad_content_dev_deploy_variants": {
                        "_dev": []string{
                                "deploy",
                        },
                },
```

code/go/pkg/validator/validator_test.go
jsoriano:
```suggestion
		},
		"bad_content_dev_deploy_variants": {
			"_dev": []string{
				"deploy",
			},
		},
```

spec/content/_dev/spec.yml
```diff
@@ -0,0 +1,9 @@
+spec:
+  additionalContents: false
+  developmentFolder: true
+  contents:
+    - description: Folder containing shared files.
```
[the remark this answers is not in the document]
kpollich: https://github.com/elastic/package-spec/pull/1077/commits/80e5af2d0cb3605ba681c974df31df8a4b03f71c

PR #1090, opened 2026-02-17
Allow optional `_dev/scripts` directory in integration, input, and content packages for development tooling files.

marc-gr:
## What does this PR do?

Allow optional `_dev/scripts` directory in integration packages for development tooling files.

## Why is it important?

We have some cases where tooling is needed in order to generate files etc as part of a package development. This change would allow us to have them in the package instead of multiple repositories

## Checklist

Closes https://github.com/elastic/elastic-package/issues/950

### DIFF
--- a/spec/content/_dev/spec.yml
+++ b/spec/content/_dev/spec.yml
@@ -7,7 +7,7 @@ spec:
       name: shared
       required: false
       $ref: "../../integration/_dev/shared/spec.yml"
-    - description: Folder containing scripts required for development.
+    - description: Folder containing scripts used for development.
       type: folder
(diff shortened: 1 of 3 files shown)


spec/content/_dev/spec.yml
```diff
@@ -7,3 +7,8 @@ spec:
       name: shared
       required: false
       $ref: "../../integration/_dev/shared/spec.yml"
+    - description: Folder containing scripts required for development.
```
teresaromero: nit

```suggestion
    - description: Folder containing scripts used for development.
```

PR #1092, opened 2026-02-17
Otel profiles input package support

MichelLosier:
## What does this PR do?

* Adds `profiles` as a policy_template type for input packages
* Adds the index template settings `auto_expand_replicas`, `max_result_window`, and `refresh_interval`

Requires the kibana changes to add support for `profiles` otel transforms: https://github.com/elastic/kibana/pull/254090

## Why is it important?

Adds spec support for an OTel Profiling integration

## Checklist

## Related issues

- Closes: https://github.com/elastic/prodfiler/issues/6623
- Closes: https://github.com/elastic/prodfiler/issues/6641

### DIFF
--- a/compliance/features/basic.feature
+++ b/compliance/features/basic.feature
@@ -45,3 +45,4 @@ Feature: Basic package types support
   Scenario: OTel input package with profiles type can be installed
    Given the "good_input_profiles" package is installed
      And a policy is created with "good_input_profiles" package, "0.0.1" version, "profilingreceiver" template, "profilingreceiver" input, "otelcol" input type and dataset "spec.otel_input_test"
+     # We omit assertions here because 'profiles' OTel packages should not produce ES assets which is currently handled by an ES plugin. https://github.com/elastic/package-spec/pull/1092#discussion_r2857893043


spec/input/manifest.spec.yml
```diff
@@ -55,6 +55,7 @@ spec:
               - logs
               - synthetics
               - traces
+              - profiles
```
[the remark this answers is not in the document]
MichelLosier: I just noticed: https://github.com/elastic/prodfiler/issues/6646 

I'll need to step back on this one and regather context on the intent here with the `type` value we want to support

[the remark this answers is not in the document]
MichelLosier: @jsoriano 
Added the kibana changes here: https://github.com/elastic/kibana/pull/254090

For the compliance test if I have the test package format_version at 3.6.0 it fails on the spec its testing against not being released. If I switch to 3.5.0 it passes - is this a change I should keep locally for development or should I have the test package at 3.6.0?

test/packages/good_input_profiles/manifest.yml
```diff
@@ -0,0 +1,41 @@
...
+        description: Defines the sampling frequency.
+elasticsearch:
+  index_template:
+    settings:
+      number_of_shards: 1
+      auto_expand_replicas: "0-1"
+      max_result_window: 150000
+      refresh_interval: "10s"
+    mappings:
+      dynamic: false
```
florianl: I see that it should showcase the new settings like `max_result_window`. 
But can we drop this part as it conflicts with the ES profiling plugin or have it in a different example package?
```suggestion
```

teresaromero (approved):
🚀

compliance/features/basic.feature
```diff
@@ -34,6 +34,11 @@ Feature: Basic package types support
      # Missing support in Kibana (Fleet) https://github.com/elastic/kibana/pull/186974
      And there is an SLO "good_content-slo-abc-1"
      And there is a detection rule "12cea9e9-5766-474d-a9dc-34ef7c7677c6"
+  
+  @3.6.0
+  Scenario: OTel input package with profiles type can be installed
+   Given the "good_input_profiles" package is installed
+     And a policy is created with "good_input_profiles" package, "0.0.1" version, "profilingreceiver" template, "profilingreceiver" input, "otelcol" input type and dataset "spec.otel_input_test"
```
mrodm: I think it is missing the `Then` section to validate that the index template is created.


```suggestion
  @3.6.0
  Scenario: OTel input package with profiles type can be installed
   Given the "good_input_profiles" package is installed
     And a policy is created with "good_input_profiles" package, "0.0.1" version, "profilingreceiver" template, "profilingreceiver" input, "otelcol" input type and dataset "spec.otel_input_test"
    Then there is an index template "profiles-spec.otel_input_test" with pattern "profiles-spec.otel_input_test.otel-*"
```

mrodm: But I think that the comment was more related to the index template settings in the manifest, not about the assertions in the compliance test.

If there is no `Then` in this compliance test, it just installs the package but it does not perform any validation on that. I think it could be still be tested that if the package is installed with the specified variables (`And` line), an index template with a given pattern is going to be installed. Is that the expected behavior? 

Could you try to install that test package (it could be done via API) and create a policy in Kibana ? To validate that the index template is created.

florianl: >> But I think that the comment was more related to the index template settings in the manifest, not about the assertions in the compliance test.
>>
> Gotcha, would you be able to clarify here @florianl ?

With https://github.com/elastic/package-spec/pull/1092#discussion_r2839330190 I want to avoid conflicts of expectations. `elasticsearch.index_template` will currently not work and if they are used, they will conflicts with the ES plugin for profiling. Therefore, I think they should not be included in the `good_input_profiles` example.
As I'm not an expert on the compliance test, I can't speak for that.

> [..] sounds like there are still unknowns on whether the package should setup any ES assets, at least for now. 

The package should not and is not intended to, set up any ES assets as they would conflicts with the ES plugin for profiling. Once the ES plugin for profiling is replaced with something else (some time in the far future, as there are no ideas or plans for it), this could change.

> When installed in its current state, no index template is produced.

That is the expected outcome for the profiling team.

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -263,6 +263,12 @@ spec:
               $ref: "#/definitions/index_template_setting_index"
             number_of_shards:
               $ref: "#/definitions/index_template_setting_number_of_shards"
+            auto_expand_replicas:
+              $ref: "#/definitions/index_template_setting_auto_expand_replicas"
+            max_result_window:
+              $ref: "#/definitions/index_template_setting_max_result_window"
+            refresh_interval:
+              $ref: "#/definitions/index_template_setting_refresh_interval"
```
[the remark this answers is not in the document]
MichelLosier: These settings don't require any special handling downstream where we just copy over index template settings as is, this update mostly just makes sure we don't error on these already valid fields.

compliance/features/basic.feature
mrodm: Thanks @MichelLosier !

Just wondering if there is something that could be assert as a `Then` in the compliance test.

> we could setup a new assertion for compliance tests that validates the expected otel config.

Could it be added an assertion that downloads the policy and ensures that there is some pipeline with the prefix `profiles`?

Any other assertion that could be added here @jsoriano ?  

```
service:
  pipelines:
    profiles/profilingreceiver-otelcol-otelcol-good_input_profiles-profilingreceiver:
      receivers:
        - >-
          profiling/profilingreceiver-otelcol-otelcol-good_input_profiles-profilingreceiver
      processors:
        - >-
          transform/profilingreceiver-otelcol-otelcol-good_input_profiles-profilingreceiver-routing
```

compliance/kibana.go
```diff
@@ -15,11 +15,14 @@ import (
...
+
+	"gopkg.in/yaml.v3"
 )
 
 const (
-	apiAgentPolicyPath   = "/api/fleet/agent_policies"
-	apiPackagePolicyPath = "/api/fleet/package_policies"
+	apiAgentPolicyPath    = "/api/fleet/agent_policies"
+	apiPackagePolicyPath  = "/api/fleet/package_policies"
+	apiInputTemplatesPath = "/api/fleet/epm/templates/%s/%s/inputs"
```
mrodm: I think the results of this API returns the inputs information even when the package is not installed , just present in the package-registry or installed but without creating any policy.

For instance a request like this :
https://localhost:5601/api/fleet/epm/templates/nginx_otel_input/0.2.0/inputs
would return the input information even without installing nor creating any policy for that `nginx_input_otel` package.

So, not sure if this would be validating that the right policy (e.g. pipeline for the profiling is created) is created after the `Given` and `And` statements, WDYT ?

At least, defining this compliance test just as it was before validates that the policy is created (but without validating anything about it): 
```
@3.6.0
  Scenario: OTel input package with profiles type can be installed
   Given the "good_input_profiles" package is installed
     And a policy is created with "good_input_profiles" package, "0.0.1" version, "profilingreceiver" template, "profilingreceiver" input, "otelcol" input type and dataset "spec.otel_input_test"
```

So, I think it would be better to just keep the previous implementation and revert the latest changes.
Thanks for trying this approach in any case @MichelLosier !!

spec/changelog.yml
```diff
@@ -51,6 +51,13 @@
     - description: Add deployer option to system benchmark configuration.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1099
+    - description: Add support for auto_expand_replicas, max_result_window, and refresh_interval index settings in data stream templates.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1092
+    # Pending on https://github.com/elastic/kibana/pull/254090
```
mrodm: Is this already supported in Kibana ? I see that at least this Pull Request is already merged.

PR #1095, opened 2026-02-23
Add semantic validator for policy template datastream categories

JDKurma:
## What does this PR do?

Adds semantic validators to verify new datastream manifest categories are aligned and synced with policy template datastream categories when present as well as parent categories alongside package level categories.

## Why is it important?
Verify and prevent drift between datastream specific categorization

## Checklist

## Related issues
N/A


teresaromero (dismissed):

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, fmt.Errorf("failed to parse categories YAML from %s: %w", packageRegistryCategoriesURL, err)
+	}
+
+	if len(rc.Categories) == 0 {
+		return nil, fmt.Errorf("no categories found in response from %s", packageRegistryCategoriesURL)
+	}
+
+	parentCategories := make(map[string]struct{}, len(rc.Categories))
+	for id := range rc.Categories {
+		parentCategories[id] = struct{}{}
```
trisch-me: why are you throwing away subcategories? How will you make a connection afterwards between child category and parent category?

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return specerrors.ValidationErrors{
+			specerrors.NewStructuredErrorf("file \"%s\" is invalid: failed to load registry categories: %w", fsys.Path(manifestPath), err)}
+	}
+
+	var errs specerrors.ValidationErrors
+	for dsName, dsCats := range dsCategories {
+		var missingCats []string
+		for _, dsCat := range dsCats {
+			if _, isParent := parentCategories[dsCat]; isParent && !slices.Contains(pkgCategories, dsCat) {
```
trisch-me: I think this is wrong. As I stated before I don’t see where you read subcategories. So here only parent category is checked but datastreams are not defining parent categories 

There is following structure:

```
dsCategories: [“AuditD”]
parentCategories SHOULD HAVE : {“security”: “AuditD”} (btw are we checking case sensitivity?)
pkgCategories: [“a”] -> missingCats should get “security” as it’s missing parent category for this package
pkgCategories: [“a”, “security”] -> all is ok
pkgCategories: [“a”, “AuditD"] -> missingCats should get “security” as it’s missing parent category for this package
```

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -116,18 +120,27 @@ func ValidateDatastreamPackageCategories(fsys fspath.FS) specerrors.ValidationEr
...
 	var errs specerrors.ValidationErrors
 	for dsName, dsCats := range dsCategories {
+		seen := make(map[string]bool)
 		var missingCats []string
 		for _, dsCat := range dsCats {
-			if _, isParent := parentCategories[dsCat]; isParent && !slices.Contains(pkgCategories, dsCat) {
-				missingCats = append(missingCats, dsCat)
+			parent, known := categoryToParent[dsCat]
+			if !known {
+				continue
```
trisch-me, line 142: I think this should be raised as error. We shouldn’t allow any unknown categories

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -0,0 +1,161 @@
...
+	"sort"
+	"time"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+)
+
+const packageRegistryCategoriesURL = "https://raw.githubusercontent.com/elastic/package-registry/main/categories/categories.yml"
```
teresaromero: i understand this is fetched so to have the source of categories and sub-categories. i feel this should be living at the spec though... but these sub-categories, they are not reflected under the package manifest? 

would it make sense to support a subcategory field so we can do this validation at package-spec? this will go upstream to elastic-package validation and arrive package-registry with the correct mapping. 

that will mean supporting both files, as i can see package-registry file is also being pulled from kibana.
wdyt @mrodm ? i might be missing some context

code/go/internal/validator/semantic/validate_datastream_package_categories.go
JDKurma: Hmm, I was split on this during development since this particular validator is not necessarily responsible for category value validation, but rather parity between package and DS categories. Ideally, any invalid categories would be validated by the DS manifest file schema. 

Nevertheless, I'll raise an error

code/go/internal/validator/semantic/validate_datastream_package_categories.go
JDKurma: > but these sub-categories, they are not reflected under the package manifest?

Not directly. The categories field can contain either parent or sub-categories. The distinction between parent and sub-categories is implicit, at the package manifest file level, to your point. 

>would it make sense to support a subcategory field so we can do this validation at package-spec? this will go upstream to elastic-package validation and arrive package-registry with the correct mapping.
that will mean supporting both files, as i can see package-registry file is also being pulled from kibana.

Perhaps, I'm not entirely sure why `package-registry` is the source of truth for categories values. It makes more sense for the spec to have the valid schema definitions rather than the registry. 
However, I feel this is a larger architecture discussion requiring alignment between `package-spec`, `package-registry`, and `elastic-package`. Happy to discuss this in another issue!

code/go/internal/validator/semantic/validate_datastream_package_categories.go
trisch-me: it’s a datastream category validator, so I believe checking that category itself is valid still fits into it :) 
thanks for checking it

trisch-me (dismissed):

JDKurma:
@teresaromero Could I get a re-review, thanks!

teresaromero:
i would like some extra 👀  from @elastic/ecosystem

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -0,0 +1,168 @@
...
```
jsoriano: We should use a pinned version here, to avoid unexpected changes or breaking it when https://github.com/elastic/package-spec/issues/1157 is implemented.
```suggestion
const packageRegistryCategoriesURL = "https://raw.githubusercontent.com/elastic/package-registry/v1.38.0/categories/categories.yml"
```

And then we should include the update to this version when new versions of Package Registry are released (around https://github.com/elastic/elastic-package/blob/a1ef15cd1efccbe657101299fcf091688b58576a/.github/workflows/updatecli/updatecli.d/bump-package-registry-version.yml).

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -0,0 +1,168 @@
...
+					fsys.Path(manifestPath),
+					dsName,
+					dsCat,
+					fsys.Path(dsManifestPath),
+				))
+			}
+			if seen[parent] {
+				continue
+			}
+			seen[parent] = true
```
jsoriano: Nit. Is `seen` used only as an optimization to avoid checking categories that share the same parent category?

I think we can avoid this optimization, there won't be many categories on each data stream and maintaining the map also has an overhead.

code/go/internal/validator/semantic/validate_policy_template_datastream_categories.go
```diff
@@ -0,0 +1,160 @@
...
+func categoriesEqual(a, b []string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	aCopy := make([]string, len(a))
+	bCopy := make([]string, len(b))
+	copy(aCopy, a)
+	copy(bCopy, b)
+	sort.Strings(aCopy)
+	sort.Strings(bCopy)
```
jsoriano: Nit. `slices.Sorted` returns a ordered copy of the elements returned by an iterator, `slices.Values` returns an iterator for the values of an slice.
```suggestion
	aCopy := slices.Sorted(slices.Values(a))
	bCopy := slices.Sorted(slices.Values(b))
```

code/go/internal/validator/semantic/validate_policy_template_datastream_categories.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		for _, dsName := range pt.DataStreams {
+			dsCats, hasCats := dsCategories[dsName]
+			if !hasCats {
+				// data stream manifest has no categories field — nothing to validate
+				continue
+			}
+
+			if !categoriesEqual(pt.Categories, dsCats) {
```
jsoriano: So a policy template needs to have exactly the same categories as the data stream that contains it? What is the use of the categories at the policy template level then?

Btw, interestingly `categories` are not allowed in policy templates in input packages.

jsoriano:
test integrations

code/go/internal/validator/semantic/validate_policy_template_datastream_categories.go
JDKurma: Good catch! Yeah the intention was always to check that policy template categories (if present) are also in DS manifest categories. Exact equality was a mistake(I think I auto-piloted when creating that func) since DS manifest categories can have values that may not be in policy template DS categories which is a valid state. 

Or more precisely, if both policy template categories and DS manifest categories are present then the former must be a subset of the latter. If either is absent, no alignment or validation required.

JDKurma:
test integrations

code/go/internal/validator/semantic/validate_datastream_package_categories.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
 )
 
-const packageRegistryCategoriesURL = "https://raw.githubusercontent.com/elastic/package-registry/main/categories/categories.yml"
+const packageRegistryCategoriesURL = "https://raw.githubusercontent.com/elastic/package-registry/v1.38.0/categories/categories.yml"
```
jsoriano, line 22: Could we add an updatecli configuration to ensure this stays updated? It could be similar to this one: https://github.com/elastic/elastic-package/blob/09286eb1b16dea302739292ce353a8a1876be1b7/.github/workflows/updatecli/updatecli.d/bump-package-registry-version.yml

If not done in this PR please do it in a follow up or open an issue so we don't lose track of this.

jsoriano (approved):

code/go/internal/validator/semantic/validate_datastream_package_categories.go
JDKurma: Cool, I'll add it in a immediate follow-up PR. 

BTW, merging is blocked for me, any chance you could merge it or grant me permission to do so? Thanks!

jsoriano: Merged, thanks!

JDKurma: @jsoriano https://github.com/elastic/package-spec/pull/1171 for the updatecli configuration

PR #1096, opened 2026-02-24
Skip SVR00002 filter validation for ES|QL panels

LikeTheSalad:
## What does this PR do?

Skips the mandatory filter validation (SVR00002) for ES|QL panels in dashboards. The `findPanelsFilters` function now detects ES|QL panels by checking for the `esql` key in the panel's query object and skips them during the filter check.

A new helper `isESQLPanel` inspects known query paths (`embeddableConfig.attributes.state.query.esql` and `embeddableConfig.query.esql`) to identify ES|QL panels, along with a `nestedMapLookup` utility for traversing the panel's nested map structure.

## Why is it important?

ES|QL panels handle data filtering through `WHERE` clauses in the query itself rather than traditional Kibana filters. The SVR00002 validation was producing false positives for these panels because it looked for filter/query paths that don't apply to ES|QL (e.g. `embeddableConfig.attributes.state.query.query`). This forced packages like `otel_android_dashboards` to disable SVR00002 entirely, losing the validation benefit for their non-ES|QL panels as well.

## Checklist

## Related issues

- Closes https://github.com/elastic/elastic-package/issues/3300

Made with [Cursor](https://cursor.com)

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -50,7 +50,7 @@
       link: https://github.com/elastic/package-spec/pull/1091
     - description: Skip mandatory filter validation (SVR00002) for ES|QL panels, which handle data filtering via WHERE clauses.
       type: bugfix
-      link: TBD
+      link: https://github.com/elastic/package-spec/pull/1096
 - version: 3.5.7
   changes:
     - description: Allow _dev directory for content-only packages; use _dev/shared for development files (e.g. dashboard YML sources).


code/go/internal/validator/semantic/validate_kibana_filter_present.go
```diff
@@ -79,6 +79,9 @@ func findPanelsFilters(file pkgpath.File) error {
 		if !visualization.CanUseFilter() {
 			continue
 		}
+		if isESQLPanel(visualization) {
```
jsoriano: This logic should be implemented in https://github.com/elastic/kbncontent, so it knows if these visualizations `CanUseFilter`, and if they do, use `HasFilters` to know if it is defined.

In the case of ES|QL, I guess that it can use filter, and it will have filter if a `FROM` clause is used with a limited set of indexes.

Issue #1094, opened 2026-02-20
[Change Proposal] Make data stream type optional when dynamic signals are enabled


jsoriano:
Since the introduction of OTel collector inputs there can be inputs that can collect multiple signal types. With the introduction of [`dynamic_signal_types`](https://github.com/elastic/package-spec/pull/1067), the set of types that are going to be used and configured is automatically detected by Fleet, based on the resulting configuration.

In these cases, the data stream type is not meaningful, and could be optional in the policy, or even required to not be set to avoid confusion.

PR #1097, opened 2026-02-25
Disallow the use of an input type when dynamic signals are enabled

jsoriano:
## What does this PR do?

Disallow the use of an input type when `dynamic_signal_types` is enabled.

## Why is it important?

The input type is not relevant when `dynamic_signal_types` is used. In these cases, the signal and data stream types used depend on the result of rendering the template, what is done in runtime.

## Checklist

- [x] ~~I have added an entry in [`spec/changelog.yml`](https://github.com/elastic/package-spec/blob/main/spec/changelog.yml).~~
  - Not added because `dynamic_signal_types` hasn't been released yet.

## Related issues

- Fixes https://github.com/elastic/package-spec/issues/1094

### DIFF
--- a/code/go/internal/validator/semantic/validate_input_dynamic_signal_types_test.go
+++ b/code/go/internal/validator/semantic/validate_input_dynamic_signal_types_test.go
@@ -171,6 +171,28 @@ policy_templates:
 		assert.Contains(t, errs[0].Error(), "got 'logfile'")
 	})
 
+	t.Run("invalid_type_field_with_dynamic_signal_types_true", func(t *testing.T) {
+		d := t.TempDir()
+
+		err := os.WriteFile(d+"/manifest.yml", []byte(`
+format_version: 3.6.0
+type: input
+policy_templates:
+  - name: otel_logs
+    type: logs
+    input: otelcol
+    template_path: input.yml.hbs
+    dynamic_signal_types: true
+`), 0o644)
+		require.NoError(t, err)
+
+		errs := ValidateInputDynamicSignalTypes(fspath.DirFS(d))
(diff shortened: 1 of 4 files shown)


teresaromero:
Couple questions about making this field optional:

- system tests rely on this field to know what datastream they have to check? this change requires the system tests at `elastic-package` to be updated so they recognize the dynamic setting.

- With this change, the source of "types of signals" is purely defined at the input template pipeline, correct? 
Have we thought about making the field string or array if the dynamic option is enabled? so we can also validate the policy pipelines are aligned with what is defined at the manifest....

code/go/internal/validator/semantic/validate_input_dynamic_signal_types_test.go
```diff
@@ -86,107 +90,117 @@ policy_templates:
...
-		// Create empty data_stream directory so it doesn't fail on missing dirs
-		err = os.Mkdir(d+"/data_stream", 0o755)
-		require.NoError(t, err)
-
 		errs := ValidateInputDynamicSignalTypes(fspath.DirFS(d))
-		require.Empty(t, errs, "expected no validation errors for integration packages with otelcol")
+		require.Empty(t, errs, "expected no validation errors for non-input packages without field")
 	})
 
-	t.Run("invalid_integration_package_non_otelcol_with_dynamic_signal_types", func(t *testing.T) {
```
mrodm: This test `invalid_integration_package_non_otelcol_with_dynamic_signal_types` is not needed anymore ?

There is another test `invalid_integration_data_stream_non_otelcol_with_dynamic_signal_types` that validates the same but for data streams.

PR #1130, opened 2026-04-03
Add support for release field in agentless deployment mode

### LINKED ISSUE #1098: [Change Proposal] Allow to declare a maturity level in the agentless deployment mode
There are cases where the agentless deployment mode of a package has a different maturity level than the default one. In these cases, we would want to be able to define it, so UIs can make informed decisions on the versions of the packages used and how they are presented to users.

The change would be to add a `release` parameter to the agentless deployment mode, so for example we can have this configuration to indicate that even if the package is GA, the agentless deployment mode is beta:
```yaml
deployment_modes:
    agentless:
      enabled: true
      release: beta   # <-- new field
```

This new field is only intended for high-level UIs, and not intended to affect discoverability of packages through the Package Registry.

MichelLosier:
## What does this PR do?

* Adds support for a `release` field in the agentless deployment mode. This is to help indicate a different maturity level than that of the package semver.  Accepts values `beta`, `ga`. Absence implies `beta` by default

## Why is it important?

Agentless deployment modes for an integration can be at a different level of functional maturity than the default deployment of an integration. This will help communicate that expectation to users downstream in Kibana.

## Checklist

## Related issues

- Closes: https://github.com/elastic/package-spec/issues/1098

### DIFF
--- a/code/go/internal/validator/semantic/validate_deployment_modes.go
+++ b/code/go/internal/validator/semantic/validate_deployment_modes.go
@@ -19,9 +19,9 @@ type deploymentModesManifest struct {
 }
 
 type deploymentModesPolicyTemplate struct {
-	Name            string                    `yaml:"name"`
-	DeploymentModes deploymentModesSpec       `yaml:"deployment_modes"`
-	Inputs          []deploymentModesInput    `yaml:"inputs"`
+	Name            string                 `yaml:"name"`
+	DeploymentModes deploymentModesSpec    `yaml:"deployment_modes"`
+	Inputs          []deploymentModesInput `yaml:"inputs"`
 }
 
 type deploymentModesSpec struct {
--- a/code/go/internal/validator/semantic/validate_deployment_modes.go
+++ b/code/go/internal/validator/semantic/validate_deployment_modes.go
@@ -30,7 +30,8 @@ type deploymentModesSpec struct {
 }
 
 type deploymentModesDefault struct {
-	Enabled *bool `yaml:"enabled"` // pointer to detect if field was set; nil means enabled
+
+	Enabled *bool `yaml:"enabled"` // pointer to detect if field was set; when unset (nil) semantical meaning is default deployment is enabled
 }
 
 type deploymentModesAgentless struct {
--- a/code/go/internal/validator/semantic/validate_deployment_modes.go
+++ b/code/go/internal/validator/semantic/validate_deployment_modes.go
@@ -138,8 +138,8 @@ func validateAgentlessReleaseDeployment(manifest deploymentModesManifest) error 
 	}
 	tmpl := manifest.PolicyTemplates[0]
 	// Default.Enabled == nil means default mode is implicitly enabled, so agentless is not the only mode.
-	isSingleDeployment := tmpl.DeploymentModes.Default.Enabled != nil && !*tmpl.DeploymentModes.Default.Enabled
-	if isSingleDeployment && tmpl.DeploymentModes.Agentless.Release != "" {
+	defaultDeploymentDisabled := tmpl.DeploymentModes.Default.Enabled != nil && !*tmpl.DeploymentModes.Default.Enabled
+	if defaultDeploymentDisabled && tmpl.DeploymentModes.Agentless.Release != "" {
 		return fmt.Errorf("policy template \"%s\" sets agentless.release but agentless is the only deployment mode; use the package version to indicate maturity instead", tmpl.Name)
 	}
 	return nil
--- a/code/go/internal/validator/semantic/validate_deployment_modes.go
+++ b/code/go/internal/validator/semantic/validate_deployment_modes.go
@@ -31,7 +31,8 @@ func ValidateDeploymentModes(fsys fspath.FS) specerrors.ValidationErrors {
 					Enabled *bool `yaml:"enabled"` // Use pointer to detect if field was set, default is true
 				} `yaml:"default"`
 				Agentless struct {
-					Enabled bool `yaml:"enabled"`
+					Enabled bool   `yaml:"enabled"`
+					Release string `yaml:"release"`
 				} `yaml:"agentless"`
 			} `yaml:"deployment_modes"`
 			Inputs []struct {
@@ -47,6 +48,19 @@ func ValidateDeploymentModes(fsys fspath.FS) specerrors.ValidationErrors {
 	}
 
 	var errs specerrors.ValidationErrors
+
+	// Check that agentless.release is not set in a single-policy-template package
+	// where agentless is the only deployment mode. In that case the package version
+	// is the authoritative source of maturity and an explicit override would conflict.
+	if len(manifest.PolicyTemplates) == 1 {
+		tmpl := manifest.PolicyTemplates[0]
+		isSingleDeployment := tmpl.DeploymentModes.Default.Enabled != nil && !*tmpl.DeploymentModes.Default.Enabled
+		if isSingleDeployment && tmpl.DeploymentModes.Agentless.Release != "" {
+			err := fmt.Errorf("file \"%s\" is invalid: policy template \"%s\" sets agentless.release but agentless is the only deployment mode; use the package version to indicate maturity instead", fsys.Path(manifestPath), tmpl.Name)
+			errs = append(errs, specerrors.NewStructuredError(err, specerrors.UnassignedCode))
+		}
+	}
+
 	for _, template := range manifest.PolicyTemplates {
 		// Collect enabled deployment modes for this policy template
 		enabledModes := []string{}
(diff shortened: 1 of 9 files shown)


spec/integration/manifest.spec.yml
```diff
@@ -295,6 +295,21 @@ spec:
...
+                deployment mode.
+              type: string
+              enum:
+                - beta
+                - rc
+                - preview
+              examples:
+                - beta
+                - rc
+                - preview
```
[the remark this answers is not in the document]
MichelLosier: Given the issue we are trying to support, I think its permissible to allow this addition to backport to older spec versions. It's a non-breaking change (to my thinking), and allowing it wont force other packages to adopt latest spec to use it. WDYT @elastic/ecosystem ?

code/go/internal/validator/semantic/validate_deployment_modes.go
```diff
@@ -47,6 +48,19 @@ func ValidateDeploymentModes(fsys fspath.FS) specerrors.ValidationErrors {
 	}
 
 	var errs specerrors.ValidationErrors
+
+	// Check that agentless.release is not set in a single-policy-template package
+	// where agentless is the only deployment mode. In that case the package version
+	// is the authoritative source of maturity and an explicit override would conflict.
+	if len(manifest.PolicyTemplates) == 1 {
+		tmpl := manifest.PolicyTemplates[0]
+		isSingleDeployment := tmpl.DeploymentModes.Default.Enabled != nil && !*tmpl.DeploymentModes.Default.Enabled
```
teresaromero: will it be single too if `Default.Enabled == nil` ? 

maybe we can extract this to a smaller helper `validateAgentlessReleaseDeployment` so the main function is cleaner? we could just add unit tests based on the input (manifest struct), and then add test under the validation tests packages with fixtures, wdyt?

spec/integration/manifest.spec.yml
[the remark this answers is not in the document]
teresaromero: i think its okay to allow this without version constraint

code/go/internal/validator/semantic/validate_deployment_modes.go
MichelLosier: Good idea! Added this in https://github.com/elastic/package-spec/pull/1130/commits/789b7d3b4015e465a90724841edd60c1704c4adf

> will it be single too if Default.Enabled == nil ?

From my understanding if `Default.Enabled == nil` then absence get translated into implicitly supporting "agent" deployment, which means we have more than 1 deployment mode with agentless.

code/go/internal/validator/semantic/validate_deployment_modes.go
```diff
@@ -14,6 +14,35 @@ import (
...
+	Inputs          []deploymentModesInput `yaml:"inputs"`
+}
+
+type deploymentModesSpec struct {
+	Default   deploymentModesDefault   `yaml:"default"`
+	Agentless deploymentModesAgentless `yaml:"agentless"`
+}
+
+type deploymentModesDefault struct {
+	Enabled *bool `yaml:"enabled"` // pointer to detect if field was set; nil means enabled
```
teresaromero:
```suggestion

	Enabled *bool `yaml:"enabled"` // pointer to detect if field was set; when unset (nil) semantical meaning is default deployment is enabled
```

nit: just adding more context to the comment, i did not see it clear the first read :D I wonder if the field for Default should be "disabled", so the option actively disables it and leaves the default. I guess that implies a change on the spec.

MichelLosier: Yeah, I got turned around a bit on the meaning on that one too 😄 . I agree, an optional `Disabled` with absence implying default state of enabled is a clearer read, but yeah seems like it would be a breaking change if we included that with the additive change in this PR which we're allowing to previous versions.

MichelLosier:
@teresaromero Changed the direction here a little bit, to just accept `ga` and `beta` as values for `release, with `beta` as the implied default as to the comment here: https://github.com/elastic/package-spec/issues/1098#issuecomment-4208690525

MichelLosier:
test integrations

spec/changelog.yml
```diff
@@ -35,6 +35,9 @@
     - description: Add `show_divider` optional boolean to input objects to suppress horizontal dividers in the Fleet UI.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1133
+    - description: Add optional `release` field to agentless deployment mode to explicitly declare its release stage.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1130
```
mrodm: This should be moved to a new section 3.6.3-next, since 3.6.0-3.6.2 were already released/published.

test/packages/bad_agentless_release/manifest.yml
```diff
@@ -0,0 +1,27 @@
+format_version: 3.6.1
+name: bad_agentless_release
+title: Bad Agentless Release
+description: Test package where agentless.release is set but agentless is the only deployment mode
+version: 0.0.1
+type: integration
+policy_templates:
```
mrodm, line 7: Would it be good to add in this test example that it requires kibana 9.4.0 or 9.5.0 as minimum ?
In case it is used as a reference for others. 

```yaml
conditions:
  kibana:
    version: '^9.5.0'  # or maybe '^9.4.1' ?
```

test/packages/bad_agentless_release/manifest.yml
mrodm, line 1: Include this parameter in a good package too please? 
It could be added into `good_v3` or add a new `good_deployment_modes` test package if it is easier.

spec/integration/manifest.spec.yml
```diff
@@ -295,6 +295,20 @@ spec:
               type: string
               examples:
                 - "cloud-security-posture-management"
+            release:
+              description: >
+                The maturity level of the agentless deployment mode for this policy template.
+                When absent, the implicit maturity is "beta".
+                Only needs to be set to "ga" when the agentless deployment mode has reached general availability.
+                Not set in packages where agentless is the only deployment mode — in that case
+                defer to the package's top-level release.
```
mrodm: Even if there is no JSON patch for this new field, as mentioned in this thread https://github.com/elastic/package-spec/pull/1130#discussion_r3034638862

Would it be good to add a note here in the description to refer that this field would take effect in kibana version 9.5 (or the required version) ?

spec/integration/manifest.spec.yml
```diff
@@ -295,6 +295,20 @@ spec:
...
+              type: string
+              enum:
+                - beta
+                - ga
```
mrodm, line 307: At the moment of reviewing this PR, there are 70 packages using `deployment_modes` in their manifests in the integrations repository.

That would mean that all of them would be considered as `beta` when the kibana change is merged (and released). Could that be an issue ?

code/go/internal/validator/semantic/validate_deployment_modes.go
```diff
@@ -110,3 +127,20 @@ func ValidateDeploymentModes(fsys fspath.FS) specerrors.ValidationErrors {
...
+// case the package version is the authoritative source of maturity and an explicit
+// override would conflict.
+func validateAgentlessReleaseDeployment(manifest deploymentModesManifest) error {
+	if len(manifest.PolicyTemplates) != 1 {
+		return nil
+	}
+	tmpl := manifest.PolicyTemplates[0]
+	// Default.Enabled == nil means default mode is implicitly enabled, so agentless is not the only mode.
+	isSingleDeployment := tmpl.DeploymentModes.Default.Enabled != nil && !*tmpl.DeploymentModes.Default.Enabled
+	if isSingleDeployment && tmpl.DeploymentModes.Agentless.Release != "" {
```
mrodm: Nit. 
What about renaming this variable to something like `defaultDeploymentDisabled` ?

spec/integration/manifest.spec.yml
MichelLosier: Ya that is currently the intent, all agentless deployments are considered beta except for CSPM.

That said, currently re-visiting the default value behavior here: https://github.com/elastic/package-spec/issues/1098#issuecomment-4326028945

I've staged up a PR for CSPM as GA: https://github.com/elastic/integrations/pull/18552

teresaromero (dismissed):
👍🏻 . i wonder if this is not a patch but a minor version update, as we are adding a new field 🤔

MichelLosier:
> i wonder if this is not a patch but a minor version update, as we are adding a new field

Good catch, it does require a Kibana version to be interpreted, but that said it does not require packages to constrain to that Kibana version. Docs will consume this separately. The next minor does seem to be the better home.

I've also updated the description of the field regarding the defaulting behavior in Kibana as discussed in the most recent comments here: https://github.com/elastic/package-spec/issues/1098

This should be all the changes intended from my end.

teresaromero (approved):
👍🏻 thanks!

PR #1132, opened 2026-04-07
Add support for `geo_shape` field type

### LINKED ISSUE #1131: Add support for `geo_shape` field types
An integration under development needs to use `geo_shape` field type. Cannot build and validate package without adding `geo_shape` as a valid type in the package spec.

tehbooom:
## What does this PR do?

Added `geo_shape` as a valid field type for packages.

## Why is it important?

An integration under development needs to use `geo_shape` field type. Cannot build and validate package without adding `geo_shape` as a valid type in the package spec.

## Checklist

- [X] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.

## Related issues

- Closes #1131


teresaromero (dismissed):
could we add into the pr a related link to the integration relevant for this change?

tehbooom:
> could we add into the pr a related link to the integration relevant for this change?

I have not submitted a PR for this integration yet. Once `geo_shape` is supported I can finish the integration and validate its correct before submission.

Can we move merge this PR still and I link back the integration later? Thanks!

teresaromero:
@tehbooom could you update changelog? thanks!

spec/integration/data_stream/fields/fields.spec.yml
```diff
@@ -612,6 +613,10 @@ spec:
 
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.6.1
+    patch:
+      - op: remove
+        path: "/items/properties/type/enum/36" #remove geo_shape type
```
mrodm: As this `geo_shape` field data type is supported in Elasticsearch for a few years (correct me if I'm wrong @tehbooom ), probably it could be made available for previous versions too. If so, it would not be needed this JSON patch.

WDYT @teresaromero @jsoriano ?

If it is needed this JSON patch, it should be moved to be after the JSON related to 3.7.0 (around line 628).

teresaromero: i think we can remove the version constraint. as we dont have a clear reference i dont see that much of a risk, probably the first integration using this will be this one

jsoriano: Agree, it looks like we can remove the version constraint. We would need to check in any case end to end that `geo_shape` fields end up correctly configured in the index templates.

tehbooom: Removed the JSON patch

compliance/features/field-types.feature
```diff
@@ -6,3 +6,9 @@ Feature: Field types
    Given the "counted_keyword" package is installed
      And a policy is created with "counted_keyword" package
     Then index template "metrics-counted_keyword.foo" has a field "foo.count" with "type:counted_keyword"
+
+  @3.6.1
```
mrodm: As the JSON patch has been removed, this package can be set to use `format_version: 3.6.0` 

and then set also here


```suggestion
  @3.6.0
```

That would make sure that the compliance test runs.

test/packages/good_geo_shape/manifest.yml
```diff
@@ -0,0 +1,23 @@
+format_version: 3.6.1
```
mrodm: As the JSON patch is removed, it can be set here a spec already released like 3.6.0


```suggestion
format_version: 3.6.0
```

compliance/features/field-types.feature
mrodm: The compliance test needs to be updated too.
So it runs according to this:
https://github.com/elastic/package-spec/blob/8c7fcac97545da93310e4edd301a697fe271d113/.buildkite/pipeline.trigger.compliance.tests.sh#L31

test/packages/good_geo_shape/manifest.yml
```diff
@@ -0,0 +1,23 @@
...
+name: good_geo_shape
+title: Good Geo Shape Package
+description: This package demonstrates valid geo_shape field type usage.
+version: 0.0.1
+type: integration
+source:
+  license: "Apache-2.0"
+conditions:
+  kibana:
+    version: '^8.0.0'
```
mrodm: As the compliance test runs with 9.4.0-SNAPSHOT for 3.6.0, it would be better to ensure tha the kibana constraint allows that stack version:


```suggestion
    version: '^8.0.0 || ^9.0.0'
```

compliance/features/field-types.feature
```diff
@@ -6,3 +6,9 @@ Feature: Field types
    Given the "counted_keyword" package is installed
      And a policy is created with "counted_keyword" package
     Then index template "metrics-counted_keyword.foo" has a field "foo.count" with "type:counted_keyword"
+
+  @3.6.0
+  Scenario: Package uses the "geo_shape" type
+   Given the "good_geo_shape" package is installed
+     And a policy is created with "good_geo_shape" package
```
mrodm: If not specified, the compliance test assumes that the version to be used is 1.0.0.

I think it should be like this this `And` condition:

```suggestion
     And a policy is created with "good_geo_shape" package and "0.0.1" version
```

mrodm (commented):
Added some suggestions to try to fix the compliance test

test/packages/good_geo_shape/data_stream/events/agent/stream/stream.yml.hbs
```diff
@@ -0,0 +1 @@
+---
```
mrodm, line 1: I think it is required some agent template here:

```suggestion
paths:
{{#each paths}}
  - {{this}}
{{/each}}
```

test/packages/good_geo_shape/data_stream/events/manifest.yml
```diff
@@ -0,0 +1,6 @@
+title: Events
+type: logs
+streams:
+  - input: logfile
+    title: Geo shape events
+    description: Collect geo shape event data
```
mrodm, line 6: Adding variable used in the agent template (stream.yml.hbs)

```suggestion
    description: Collect geo shape event data
    vars:
      - name: paths
        type: text
        title: Paths
        multi: true
        required: true
        show_user: true
        default:
          - /var/log/*.log
```

mrodm (commented):
There are still some failures while testing locally:

```
  Scenario: Package uses the "geo_shape" type                                                   # features/field-types.feature:11
    Given the "good_geo_shape" package is installed                                             # compliance_test.go:431 -> github.com/elastic/package-spec/compliance.thePackageIsInstalled
    And a policy is created with "good_geo_shape" package and "0.0.1" version                   # compliance_test.go:433 -> github.com/elastic/package-spec/compliance.aPolicyIsCreatedWithPackageAndVersion
    failed to create package policy: request failed with status 400, body: {"statusCode":400,"error":"Bad Request","message":"Cannot convert undefined or null to object"}
    Then index template "logs-good_geo_shape.events" has a field "region" with "type:geo_shape" # compliance_test.go:430 -> github.com/elastic/package-spec/compliance.indexTemplateHasAFieldWith
    suite.go:640: failed to create package policy: request failed with status 400, body: {"statusCode":400,"error":"Bad Request","message":"Cannot convert undefined or null to object"}
```

I think it is missing to add some variables as well as the agent template to make it work. I added suggestions for those too.

Let's see if that works in the CI too.

mrodm (dismissed):

teresaromero (dismissed):
🚀 🗺️

mrodm:
@tehbooom sorry, could you update the changelog ? There are some conflicts due to a some PRs merged in the meantime.

Once solved, let us know to merge this PR if you don't have permissions for that.

teresaromero (approved):

mrodm (approved):

PR #1133, opened 2026-04-08
[Fleet] Add support for section headers, divider toggling, and reorganization

Supplementing:
## What does this PR do?

This PR is part of the Kibana changes in https://github.com/elastic/kibana/pull/262129. It adds support for section headers (which allow Kibana to render arbitrary EuiTitles), support for divider visibility toggles, and better control over ordering of fields when using var groups

## Why is it important?

With the introduction of var_groups, there was a lot of constraints around the layout, we needed to give the integration developers a little more control at the manifest level to impact how things show up at the Kibana level. 

## Checklist

## Related issues

Relates https://github.com/elastic/kibana/pull/262129
Relates https://github.com/elastic/integrations/pull/17495

commit message: added support for sections and divider toggling


spec/changelog.yml
```diff
@@ -8,6 +8,15 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add `section_header` as a valid var type for decorative section headings in the Fleet UI.
```
teresaromero: can we add as a comment the kibana pr that is related? that way we know when the related feature is released
```suggestion
    # Pending on xyz
    - description: Add `section_header` as a valid var type for decorative section headings in the Fleet UI.
```

test/packages/good_var_groups/manifest.yml
teresaromero, line 1:
```suggestion
format_version: 3.7.0
```

as the feature is intended for 3.7.0 the errors on unit test are because the package has a version that does not support the new field.

spec/changelog.yml
Supplementing: Done in latest commit

test/packages/good_var_groups/manifest.yml
Supplementing: Also done in latest commit

teresaromero (dismissed):

jsoriano (changes requested):
Hey, sorry for chiming it a bit late.

I think current proposal is good, but I have some concerns, specially about `section_header`:
* It overloads the meaning of `type`, it starts containing UI elements apart from actual data types.
* How would it work once we start composing lists of variables (https://github.com/elastic/elastic-package/issues/3279).
* It can be error prone, a developer may reorder some variable, without taking into account the section it belongs to. Sections depend too much on the order.
* In general we may be mixing too much variable definition with its representation.

What do you think about an alternative approach to solve this? Instead of a `section_header` variable type, we could:
* Add a top-level `sections` list to define the section headers, with any specific metadata they may need, and their display order.
* Add a `section: <section_name>` attribute to individual variables to assign them to a specific group.

This would keep the `type` field strictly focused on data types rather than presentation, and would be more explicit on what variables belong to each section.

What do you think?

(Requesting changes only to discuss this alternative option, again, sorry for chiming in late 🙏)

Supplementing:
> Hey, sorry for chiming it a bit late.
> 
> I think current proposal is good, but I have some concerns, specially about `section_header`:
> 
> * It overloads the meaning of `type`, it starts containing UI elements apart from actual data types.
> * How would it work once we start composing lists of variables ([elastic/elastic-package#3279](https://github.com/elastic/elastic-package/issues/3279)).
> * It can be error prone, a developer may reorder some variable, without taking into account the section it belongs to. Sections depend too much on the order.
> * In general we may be mixing too much variable definition with its representation.
> 
> What do you think about an alternative approach to solve this? Instead of a `section_header` variable type, we could:
> 
> * Add a top-level `sections` list to define the section headers, with any specific metadata they may need, and their display order.
> * Add a `section: <section_name>` attribute to individual variables to assign them to a specific group.
> 
> This would keep the `type` field strictly focused on data types rather than presentation, and would be more explicit on what variables belong to each section.
> 
> What do you think?
> 
> (Requesting changes only to discuss this alternative option, again, sorry for chiming in late 🙏)

Yeah, I agree with your points. In hindsight, I think the implementation is a bit to flexible and could allow the user to move things around without realizing what they are doing fully. I like the idea of a sections list and using the attributes on the variables to link them to the sections. 

However, the issue is now that the Kibana PR has been merged so we will need to circle back and open another PR to address the changes. Not a biggie though, and since all this is ongoing and packages havent been released anyway, theres no churn to really worry with. 

I can start working on making changes on this PR then open a follow up PR in Kibana to utilize the changes.

jsoriano:
> However, the issue is now that the Kibana PR has been merged so we will need to circle back and open another PR to address the changes. Not a biggie though, and since all this is ongoing and packages havent been released anyway, theres no churn to really worry with.

Yeah, I think it is worth to do it at this point as this hasn't been used yet in actual packages.

> I can start working on making changes on this PR then open a follow up PR in Kibana to utilize the changes.

Great, thanks!

test/packages/good_var_groups/manifest.yml
```diff
@@ -13,31 +13,40 @@ conditions:
...
+  - name: auth_section
+    title: Authentication
+    description: Configure AWS authentication credentials.
 vars:
   - name: access_key_id
     type: text
     title: Access Key ID
     show_user: true
     secret: false
+    section: auth_section
```
teresaromero, line 26: 🤔 have you thought on referencing the variables at `sections` and keep all sections fields wrapped into it? i am just thinking to avoid repetitive field `section` at the vars, but haven't evaluated how this will look like operating at fleet

```
  - name: auth_section
    title: Authentication
    description: Configure AWS authentication credentials.
    vars:
       - access_key_id
       - session_token
    var_groups:
       - whatever 
```

Supplementing: I think its something worth discussing for sure. A couple of thoughts I have: 

- Moving to the approach you recommended introduces a little more complexity in validation. You could remove a var and if you forget to remove it from the section list, then you have a dangling reference. We would need to update the validation logic to make sure that referenced vars exist, match exactly, and are only present in one section. The current approach only allows one section per var, so we only have to check that the section they are mapped to exists. 
- The next thing to consider is ordering. With the current approach, ordering is intuitive. This means its based on the order in the manifest. If we move to the approach recommended, we would need to explicitly set the logic in the Kibana side to render in order, or add some sort of way to control the order. 

Its worth noting that var_groups uses the approach you recommended, but they are behavioral and control visibility, this is purely visual, so i dont know if it warrants the same. 

I'll also add that I expect most of these sections to be rather small as theyre more of subsections if that makes sense. So theyre likely to be a few variables at most. 

In my opinion, I don't know if it's worth the added complexity, especially on the Kibana side to implement the recommended approach. The current approach, although a little more repetitive in some cases, is a little more intuitive when it comes to ordering etc. The developer experience though should be considered. WDYT?

jsoriano: I think I prefer having `section` in the variable definition itself, so when a variable is created, reorganized or deleted everything goes together.

Having the list of variables in the section would allow syntactically to have a variable in multiple sections, but this doesn't make a lot of sense in this case, what would force us to have semantic validations to avoid that.

I think current approach represents better what we want to define.

spec/input/manifest.spec.yml
```diff
@@ -168,6 +172,13 @@ spec:
 
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.7.0
+    patch:
+      # sections support added in 3.7.0.
+      - op: remove
+        path: "/properties/sections"
+      - op: remove
+        path: "/properties/policy_templates/items/properties/sections"
```
jsoriano: Maybe we don't need a version patch. On current versions of Kibana sections will be just ignored, right?

This is probably fine for this feature, and would allow to start using it earlier for packages that support a broad range of versions of Kibana.

test/packages/good_var_groups/manifest.yml
teresaromero: makes sense!

spec/input/manifest.spec.yml
Supplementing: I went ahead and did this and also applied the same for the `show_divider`, since its only a UI change and can just be ignored or optionally used on older (current) versions, its probably fine to not remove it. WDYT?

spec/integration/manifest.spec.yml
```diff
@@ -586,6 +586,33 @@ spec:
...
+            description: Display title for this section header in the Fleet UI.
+            type: string
+            examples:
+              - Authentication
+          description:
+            description: Optional help text displayed below the section header.
+            type: string
+        required:
+          - name
+          - title
```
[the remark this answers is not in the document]
Supplementing: The warning here can be ignored, the removal was intentionally omitted, see our convo above https://github.com/elastic/package-spec/pull/1133#discussion_r3099226330

code/go/internal/validator/spec.go
```diff
@@ -225,6 +225,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateCapabilitiesRequired, since: semver.MustParse("2.10.0")}, // capabilities definition was added in spec version 2.10.0
 		{fn: semantic.ValidateRequiredVarGroups},
 		{fn: semantic.ValidateVarGroups, since: semver.MustParse("3.6.0")},
+		{fn: semantic.ValidateSections, since: semver.MustParse("3.6.0")},
```
jsoriano: We should not keep a version here if this is going to be supported on all spec versions.
```suggestion
		{fn: semantic.ValidateSections},
```

spec/integration/manifest.spec.yml
```diff
@@ -586,6 +586,36 @@ spec:
           type: array
           items:
             $ref: "#/definitions/package_dependency"
+    show_divider:
+      description: When false, suppresses the automatic horizontal divider rendered after this section.
+      type: boolean
```
jsoriano, line 591: Should we include here its default value? 
```suggestion
      type: boolean
      default: true
```

spec/integration/manifest.spec.yml
```diff
@@ -586,6 +586,36 @@ spec:
           type: array
           items:
             $ref: "#/definitions/package_dependency"
+    show_divider:
+      description: When false, suppresses the automatic horizontal divider rendered after this section.
```
jsoriano, line 590: Is this a property related to sections or to inputs?
```suggestion
      description: When false, suppresses the automatic horizontal divider rendered after the configuration UI for this input.
```

Supplementing: 'Section' is used loosely here, it means section in terms of the section of inputs in the UI. If you look at this screenshot, its the divider(s) at the bottom after 'advanced options' 
image

jsoriano (dismissed):
LGTM, thanks!

teresaromero (dismissed):
thanks!!

spec/changelog.yml
```diff
@@ -8,6 +8,16 @@
...
+    # pending on https://github.com/elastic/kibana/pull/262129
+    - description: Add top-level `sections` list for defining named section headers, and `section` attribute on vars for assigning variables to a section in the Fleet UI.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1133
+    - description: Add `show_divider` optional boolean to input objects to suppress horizontal dividers in the Fleet UI.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1133
+    - description: Add `var_groups` optional array to input, stream, and input-package policy template objects for Fleet UI grouping.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1133
```
mrodm: I don't see in this PR the changes mentioned in this changelog entry @Supplementing . Is it missing something in the PR? or, could this changelog entry be removed ?

Supplementing: Its not needed, upstream already handled this actually so it became redundant. Removed!

mrodm (approved):

teresaromero (approved):

PR #1134, opened 2026-04-08
Add fleet reserved variable validation

### LINKED ISSUE #1100: [Change Proposal] Support providing a default value for `use_apm` variable in integration type packages
The `use_apm` variable is a special-cased variable injected by Fleet for OTel input packages that support ingesting traces. We should support this variable for integration packages as well that include an `otelcol` input. Integration packages should be able to set a default value for this variable of true/false. 

cc @nchaulet @jsoriano

MichelLosier:
## What does this PR do?

- Adds validation around the inclusion of `use_apm`, and `data_stream.dataset` variable overrides, which fleet already injects automatically if not present.
- `use_apm`
  - is of type `bool`
  - is used only on `otelcol` inputs of streams that are traces or dynamic_signal_types
- `data_stream.dataset`
  - is of type `text`

Where we do injection in fleet, but defer to existing declarations:
* https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/common/services/policy_template.ts#L212-L221

## Why is it important?

While already allowing integrations to decide on the default value of these special handled variables, these validations make sure that the expected variable types and constraints are observed.

## Checklist

## Related issues

- Closes: https://github.com/elastic/package-spec/issues/1100


MichelLosier:
test integrations

code/go/internal/validator/semantic/validate_fleet_reserved_vars.go
```diff
@@ -0,0 +1,218 @@
...
+	var errs specerrors.ValidationErrors
+
+	dataStreams, err := listDataStreams(fsys)
+	if err != nil {
+		return nil, specerrors.ValidationErrors{
+			specerrors.NewStructuredErrorf("failed to list data streams: %w", err)}
+	}
+
+	for _, ds := range dataStreams {
+		manifestPath := dataStreamDir + "/" + ds + "/manifest.yml"
```
teresaromero: i always struggle with path construction between linux/windows. use path.Join instead to avoid slash errors between platforms

teresaromero (commented):
looks good! couple thoughts:

- variables for integrations can be defined also at package root level (vars) and policy template input level (vars), i think we are missing these cases. for input packages is also a vars at root level and the one at the policy templates
- i was thinking if perhaps instead of creating the streams variables and then running again a loop, we could run validation on the only loop when normalizing. this way we just go through once and we can append validation errors on the go, skipping the variables we are not interested. if i read correctly, the streams list have variables we dont even care. wdyt?

code/go/internal/validator/semantic/validate_fleet_reserved_vars.go
teresaromero: i see the tests passed with this config 🤔 interesting

MichelLosier: oof yes, good catch! 😄

MichelLosier:
Thanks!
> variables for integrations can be defined also at package root level (vars) and policy template input level (vars), i think we are missing these cases. for input packages is also a vars at root level and the one at the policy templates

Good callout, I've added to the validation of these vars constraints at what scope they can be used, since `data_stream.dataset` and `use_apm` are stream level concerns, and also where fleet currently expects them. So with that validation will now error if they are found at integration: package root, and policy template input level, or input: root level.

> i was thinking if perhaps instead of creating the streams variables and then running again a loop, we could run validation on the only loop when normalizing. this way we just go through once and we can append validation errors on the go, skipping the variables we are not interested. if i read correctly, the streams list have variables we dont even care. wdyt?

Sounds good to me, I've refactored it run on normalization and validation in the same pass.

code/go/internal/validator/semantic/validate_fleet_reserved_vars.go
jsoriano: > i always struggle with path construction between linux/windows. use path.Join instead to avoid slash errors between platforms

For local filesystems `filepath.Join` should be used, that is the one taking into account the different platforms. `path.Join` always uses `/`.

> i see the tests passed with this config 🤔 interesting

Here we are using a `fs.FS`, that always uses unix paths, so `path.Join`, or just joining with `/` is fine. If we were using a local filesystem and supporting multiple platforms, we would need to use `filepath.Join`.

jsoriano (dismissed):

MichelLosier:
Addressed merged conflicts ^

jsoriano (approved):

teresaromero (approved):

PR #1135, opened 2026-04-09
Add support for inputs with qualified names

jsoriano:
## What does this PR do?

Add support for inputs with qualified names. See fleet implementation in https://github.com/elastic/kibana/pull/262138.

This field aims to fix cases of multiple inputs of the same type in policy templates, what starts being common with `otelcol`. In these cases a `name` can be used to explicitly correlate inputs in data streams.

## Why is it important?

With the introduction of OTel integrations in 3.6 it starts to be common having multiple inputs with the same type in the same policy template. This leads to ambiguous situations, as data streams referring to these inputs by type cannot distinguish between them.

## Checklist

## Related issues

- Kibana implementation in https://github.com/elastic/kibana/pull/262138.


jsoriano:
test integrations

code/go/internal/validator/spec.go
```diff
@@ -237,6 +237,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateKibanaTagDuplicates},
 		{fn: semantic.ValidatePipelineOnFailure, types: []string{"integration"}, since: semver.MustParse("3.6.0")},
 		{fn: semantic.ValidateIntegrationInputsDeprecation, types: []string{"integration"}, since: semver.MustParse("3.6.0")},
+		{fn: semantic.ValidateIntegrationInputQualifier, types: []string{"integration"}, since: semver.MustParse("3.6.0")},
```
teresaromero, line 240: would this be 3.6.1?

teresaromero (approved):

code/go/internal/validator/spec.go
jsoriano: I would like to allow this from 3.6.0.

PR #1140, opened 2026-04-15
Support configuring dynamic dataset and namespace in input packages

jsoriano:
## What does this PR do?

Add support to configure dynamic dataset and namespace in input packages.

## Why is it important?

There can be generic inputs that can store documents, or their documents can be rerouted, to multiple datasets or namespaces.

Currently Kibana is adding permissions for input packages depending on the kind of input, instead of that, it should give the minimum set of permissions, and developers should use `dynamic_dataset`/`dynamic_namespace` if permissions for more datasets are required.

## Checklist

## Related issues

-


jsoriano:
I think it was actually intentional to have all input packages work as if they always had dynamic dataset and namespace. To support dynamic routing.

See discussions about this in:
* https://github.com/elastic/package-spec/pull/327
* https://github.com/elastic/kibana/pull/157897

PR #1149, opened 2026-04-22
add ml_module asset type to content package spec

### LINKED ISSUE #1147: [Change Proposal] Support ML modules in content packages
`ml_module` asset type is currently supported in integration packages, but not content packages; this proposal suggests we support it there too.

https://github.com/elastic/package-spec/blob/main/spec/integration/kibana/spec.yml#L86

tommyers-elastic:
## What does this PR do?

add ml_module asset type to content package spec

## Why is it important?

fixes #1147 

## Checklist

## Related issues

-

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -8,6 +8,11 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+- version: 3.6.2-next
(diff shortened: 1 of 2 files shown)


mrodm (approved):
LGTM!
As content packages were added in 3.4.0, at that time `ml_module` asset should be supported too, so there is no need to add any JSON patch for that new entry.

PR #1162, opened 2026-05-06
Add migrate from variable support

Supplementing:
## What does this PR do?

Closes https://github.com/elastic/kibana/issues/266829

This PR adds support for `migrate_from` at the variable level allowing package maintainers to move variables between scopes (for instance input <-> stream). 

As we already has an existing field named `migrate_from`, it was rolled into an object form. You now must use `anyOf` name or scope in both places you use `migrate_from`.

## Why is it important?

We needed this in order to stop vars from being dropped when maintainers moved them in scope. See https://github.com/elastic/kibana/issues/264085

## Checklist

## Related issues

-

## Summary by CodeRabbit

* **New Features**
  * Added support for a migrate_from object on variables to indicate migrating values between scopes (input ↔ stream).

* **Tests**
  * Expanded validation tests and fixtures to cover valid and invalid migrate_from scenarios (including scope validation).

* **Documentation**
  * Added documentation and sample package scaffolds illustrating migrate_from usage and examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elastic/package-spec/pull/1162)


spec/changelog.yml
```diff
@@ -11,6 +11,9 @@
     - description: Add optional `release` field to agentless deployment mode to explicitly declare its release stage.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1130
+    - description: Add `migrate_from` object to variable definitions to declare that a variable moved between input-level and stream-level scope within the same input type.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1162
```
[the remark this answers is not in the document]
Supplementing: Upstream is actually depending on this, not the other way around

teresaromero (commented):
hi! regarding this change i have couple questions:

- will the variables coming from an input package (composables) be in the scope of this change? when an integration uses packages on its data-streams, the variables from the input package are merged into the integration, and whatever variable is declared at the integration overrides the one coming from the input. With this said, how is this change affecting this cases? 

-  i understand fleet reads this field from the manifest an uses the migrated value instead of the "original". How does this affect elastic-package? I believe system tests would need to mirror the behavior if, for example, some system test has variables declared?

Supplementing:
> * will the variables coming from an input package (composables) be in the scope of this change? when an integration uses packages on its data-streams, the variables from the input package are merged into the integration, and whatever variable is declared at the integration overrides the one coming from the input. With this said, how is this change affecting this cases?

This wasnt in scope for this change. Input packages themselves cant declare `migrate_from` on their vars, and the existing merge happens upstream of this change. If this is something we want to support, then we should to discuss if its something that is a blocker for this particular change or should be a follow up. 

> * i understand fleet reads this field from the manifest an uses the migrated value instead of the "original". How does this affect elastic-package? I believe system tests would need to mirror the behavior if, for example, some system test has variables declared?

By my understanding, no elastic-package changes are needed here, the package-spec release step will resolve any issues. Once the spec release happens, elastic-package will pick up the changes required. System tests install one version, the migration is upgrade-only.

spec/changelog.yml
```diff
@@ -11,6 +11,9 @@
     - description: Add optional `release` field to agentless deployment mode to explicitly declare its release stage.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1130
+    - description: Add `migrate_from` object to variable definitions to declare that a variable moved between input-level and stream-level scope within the same input type.
```
teresaromero, line 14: will this wait until 3.7 or shall it be 3.6.3 ??? cc @jsoriano if we need a quick release of this maybe we can add it to next elastic-package release?

teresaromero (approved):
its fine for 3.7 if packages from this version forward could include this field. I dont see how this is blocking kibana, as the spec will affect packages from version 3.7 kibana should be able to check if the field is there and act on it 🤔  so until there are no packages with version 3.7 kibana wont be seeing this? is this correct?

the spec goes to elastic-package so integration devs can update their packages with the new spec fields, and bump the version. but i spec is not injected at fleet (as far as i know)

spec/changelog.yml
jsoriano: This needs to go in 3.7, as this requires changes in Kibana.

Supplementing:
Quick update so theres a trail, this is no longer a blocker for the Kibana side! It can (and likely will) merge before this is ready. It has been verified to pass CI and will work without these changes.

PR #1168, opened 2026-05-12
Prepare release 3.6.3

jsoriano:
## What does this PR do?

Prepares the release of package-spec **3.6.3**.

Changes in this release:

- **Enhancement**: Add optional `release` field to agentless deployment mode to explicitly declare its release stage. ([#1130](https://github.com/elastic/package-spec/pull/1130))

## Why is it important?

Cuts the 3.6.3 patch release so that packages can use the new `release` field in agentless deployment mode.

## Checklist

- ~~[ ] I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~~

## Related issues

- Relates https://github.com/elastic/package-spec/pull/1130

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Chores**
  * Bumped compliance spec to 3.6.3 and updated compliance test generation.
  * Restored changelog structure for version 3.6.3 and removed stale pending markers.

* **New Features**
  * Added a minimal OpenTelemetry test integration with logs and metrics streams and package manifest.

* **Documentation**
  * New README for the OpenTelemetry test package and clarified manifest comments.

* **Tests**
  * Enabled an OTel integration scenario and added related test fixtures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elastic/package-spec/pull/1168)


mrodm (dismissed):
🚀

teresaromero (commented):
I think we need to fix good_v3 agentless deployment modes as its the shared package for the compliance tests. From claudes investigation:

```
Single root cause (shared by both jobs)

  compliance_test.go runs 27 scenarios; 26 pass, 1 fails — the same one in both stack versions:

  Scenario: Integration package with OTel input can be installed   # features/basic.feature:39
    And a policy is created with "good_v3" package, "1.1.0" version,
        "otel" template, "otel_logs" input, "otelcol" input type and dataset ""
    Error: failed to create package policy: request failed with status 400, body:
      {"statusCode":400,"error":"Bad Request",
       "message":"Input aws/s3 in good_v3 is not allowed for deployment mode 'default'"}

  Why this is happening

  This scenario was un-skipped on this branch in 709e41a — "Enable named inputs compliance test now that blockers are merged" (the named-inputs
  blockers were package-spec#1135 and kibana#262138).

  The test creates a policy for the otel policy template in good_v3 (test/packages/good_v3/manifest.yml:160). The package also defines an
  apache-agentless template (line 116) whose aws/s3 input is restricted to deployment_modes: ['agentless'] (line 149-152). Kibana 9.4.0 and
  9.5.0-SNAPSHOT are rejecting the request because they validate the package's aws/s3 input against the policy's deployment mode (default) — even
   though that input belongs to a different policy template that the test isn't using.

  So the "blockers merged" assumption in 709e41a doesn't hold for these stack versions: either the corresponding Kibana fix hasn't actually
  landed in 9.4.0 / today's 9.5.0-SNAPSHOT, or it only handles named inputs but not the cross-template deployment-mode check that the existing
  fixture exercises.
```

jsoriano:
`good_v3` triggers a rare error condition in Fleet. It tries to validate all policy templates when creating a policy, even the ones that won't be used. This makes it impossible to have pure agentless policy templates in the same package where non-agentless templates are also available, as this can lead to contradictory situations.

Issue created about this: https://github.com/elastic/kibana/issues/268930

I am leaving `good_v3` as is, and adding a new simpler OTel integration package to be used in the compliance tests.

compliance/testdata/packages/good_integration_otel/manifest.yml
```diff
@@ -0,0 +1,31 @@
...
+  - src: /img/sample-logo.svg
+    title: Sample logo
+    size: 32x32
+    type: image/svg+xml
+policy_templates:
+  - name: otel
+    title: OTel logs
+    description: Collect logs via OTel receiver
+    inputs:
+      - type: otelcol
```
jsoriano: There should be two otelcol inputs here, to actually check resolution of ambiguous names, adding it.

jsoriano: Ah no, this cannot work yet, till compliance tests build packages with an elastic-package with https://github.com/elastic/elastic-package/pull/3480.

I will add the second input after elastic-package is released and used here.

mrodm (approved):

teresaromero (approved):

test/packages/good_integration_otel/data_stream/otel_logs/manifest.yml
```diff
@@ -0,0 +1,6 @@
+title: OTel logs
+type: logs
+streams:
+  - input: otel_logs
```
teresaromero, line 4: shouldnt this be `otelcol` ??? 
also, this case is not yet supported as we need elastic-package release with the name support

jsoriano: Ah yes, this is only passing now I think because it is not using the policy. I have to review the compliance tests in any case after elastic-package release.

jsoriano: Ah no, sorry, this is fine. This is not a composable package, so the name should be already resolved.

PR #1175, opened 2026-05-28
Add source and build validation modes

### LINKED ISSUE #549: [Change Proposal] Add different validation modes for source and built packages

We would need to:

teresaromero:
## What does this PR do?

Introduces three explicit validation modes — `legacy`, `source`, and `build` — so that source packages and built packages are validated against distinct rule sets.

**New public API** (`pkg/validator`):
- `NewFromPath(mode, path)`, `NewFromFS(mode, location, fsys)` — mode-aware constructors with eager path/FS validation at construction time (clear errors rather than a confusing failure deep inside `Validate()`)
- `NewFromZip(path)` — always `ModeBuild`; zip files are by definition built packages
- `ValidateFromPath`, `ValidateFromZip`, `ValidateFromFS` preserved as deprecated wrappers using `ModeLegacy`

**Source mode** rejects `_embedded_ecs` keys in `dynamic_templates` (resolves the long-standing TODO at `manifest.spec.yml:350`) and skips build-mode-specific checks.

**Build mode** rejects source-only artifacts:
- any `_dev/` directory
- any `.link` file
- fields with `external: ecs`
- data stream manifest `streams[]` with `package:` or missing `input:`
- policy template inputs with `package:` instead of `type:`

The `legacy` mode reproduces today's behaviour byte-for-byte and is what all existing callers get through the deprecated wrappers.

**Composable package support (source mode):**

Four false-positive categories that fired when linting composable integration packages (ones that reference input packages via `package:`) are now fixed:

- **Template validation skipped for dependency-provided templates**: streams and policy template inputs that set `package:` but declare no local `template_path`/`template_paths` skip template-existence checks in source mode. Templates that come from the dependency are not present in source and only materialise at build time. Streams and inputs that *do* declare explicit overlay `template_paths` still have those files validated.
- **Partial var definitions allowed on composable streams/inputs**: vars on composable streams (`package:` set) and composable policy template inputs (`package:` set) are treated as partial overrides of vars defined in the referenced input package. `name`, `type`, and type-specific constraints (`options`, `secret`, duration patterns) are not required in source — they are assigned or enforced by elastic-package when it materialises the composable reference. Non-composable streams/inputs (`input:`/`type:` set) retain full var validation.
- **SVR00010 (input qualifier) restricted to build mode**: `ValidateIntegrationInputQualifier` now runs in build mode only. In source, composable inputs carry `package:` instead of `type:`, leaving `input.Type == ""` for every composable input; when a policy template has multiple such inputs SVR00010 fired spuriously. In build mode `ValidateStreamInputMaterialized` has already replaced every `package:` with a real `type:`, so the name-uniqueness check is meaningful.

## Why is it important?

The validator today applies a union of all rules regardless of whether a package is a checked-out source tree or a built artifact ready for distribution. This lets malformed source packages build without error and malformed built packages reach the registry undetected. Closes #549.

## Checklist

## Things to review

- **Mode filtering** (`internal/validator/spec.go` → `rules()`): rules with `modes: nil` run in all three modes; rules tagged with specific modes are filtered out. Confirm the tagging on `ValidateExternalFieldsWithDevFolder` and `ValidateTestPackageRequirements` (source+legacy only), and `ValidateIntegrationInputQualifier` (build only) is correct.
- **`NewFromZip` always `ModeBuild`**: this is a new function — no existing callers break. The design decision is that `NewFromZip` enforces `ModeBuild` at the API level rather than leaving it to the caller. The deprecated `ValidateFromZip` stays on `ModeLegacy` via an internal helper.
- **`NewFromFS(ModeSource)`**: wraps with `linkedfiles.NewFS(location, fsys)` so `.link` files are resolved transparently. Callers passing an in-memory FS should be aware that link resolution hits the real OS filesystem at `location`.
- **`good_v3/data_stream/ecs_import_mappings/manifest.yml`**: the large deletion (466 lines) removes `_embedded_ecs-*` dynamic template entries that were build artifacts committed into a source fixture. They now live in `test/packages/bad_embedded_ecs/` which is the fixture that proves source mode rejects them. The `ecs_import_mappings` data stream is left with ordinary `dynamic_templates` entries.
- **`good_built` fixture** (`test/packages/build_mode/good_built/`): hand-crafted to represent a valid built package. It is the reference for all build-mode passing tests; if it is wrong, the build-mode rules will be miscalibrated.
- **Composable var validation** (`data_stream/manifest.spec.yml` + `integration/manifest.spec.yml`): the `if: required: [input]` / `if: required: [type]` guards on stream/input items are the mechanism that switches between permissive (composable) and strict (non-composable) var validation. The four `allOf` conditions in `#/definitions/vars/items` now include `required: [type]` guards to prevent spurious cascades when `type` is absent.

## How to test

Automated: all existing tests pass against the deprecated `ValidateFrom*` wrappers, which use `ModeLegacy` and preserve byte-for-byte identical behaviour with `main`.

Manual: `elastic-package` was run with this package-spec pinned against the full integrations repository. All integrations validated and built successfully — no breaking changes to existing callers.

A follow-up PR to `elastic-package` will migrate its `check`/`lint` commands to `ModeSource` and its `build` command to `ModeBuild` using the new constructor API.

## Related issues

- Closes #549

## Summary by CodeRabbit

* **New Features**
  * Added validation modes (Legacy, Source, Build) and mode-aware spec selection.
  * New public Validator API with mode-aware constructors and mode exports.
  * New build-mode validators: reject embedded ECS mappings, external ECS refs, `_dev` folders, `.link` files, and require stream input materialization.

* **Bug Fixes**
  * Data stream listing now ignores non-directory entries.

* **Tests**
  * Added extensive tests covering modes, validators, and fixture scenarios.

* **Chores**
  * Updated .gitignore to ignore .scratch/ directory.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elastic/package-spec/pull/1175?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


code/go/internal/validator/modes/modes.go
```diff
@@ -0,0 +1,24 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package modes
+
+// Mode represents the validation mode used when validating a package.
+type Mode string
+
+// Validation modes. The public API re-exports these as validator.Mode* constants.
+const (
+	Legacy Mode = "legacy"
```
Supplementing, line 12: We originally discussed that legacy was functionally equivalent to source mode, but then there was some additional discussion about maybe keeping it for some time until we deprecate it and it collapses into source by default. Just wanted to see what the plan was here and if we really even need legacy at all if theyre functionally eq.

code/go/internal/validator/semantic/validate_no_embedded_ecs.go
```diff
@@ -0,0 +1,79 @@
...
+	} `yaml:"elasticsearch"`
+}
+
+// ValidateNoEmbeddedEcsInDynamicTemplates rejects data stream manifests that contain
+// dynamic_templates entries whose keys match the "^_embedded_ecs" pattern.
+//
+// Keys starting with "_embedded_ecs" are auto-injected by elastic-package at build time
+// when import_mappings is enabled. They must not appear in source packages and are
+// rejected in source validation mode. Built packages (where these keys are expected)
+// should be validated with ModeBuild, not ModeSource.
```
Supplementing, line 36: This is a general question due to my own lack of understanding, in Jaime's comment [here](https://github.com/elastic/package-spec/issues/549#issuecomment-4440628334) he mentions that we should treat embedded_ecs as a one off as its not usual, so is this something we want to validate against?

code/go/internal/validator/spec.go
```diff
@@ -44,8 +47,13 @@ type validationRules []validationRule
 // GASpecCheckVersion represents minimum version to start checking for unreleased version of the spec
 var GASpecCheckVersion = semver.MustParse("3.0.1")
 
-// NewSpec creates a new Spec for the given version
-func NewSpec(version semver.Version) (*Spec, error) {
+// NewSpec creates a new Spec for the given version and validation mode.
+// If mode is empty, it defaults to modes.Legacy.
+func NewSpec(version semver.Version, mode modes.Mode) (*Spec, error) {
+	if mode == "" {
+		mode = modes.Legacy
```
Supplementing, line 54: Depending on the answer above about whether we need legacy or not (im unbiased here FWIW), this would default to source. 

Also, looks like Coderabbit called it out below, but adding a guard clause against invalid modes would be good. In case someone passes 'src' or something not supported.

Supplementing (commented):
Just a few clarification questions/comments, but overall it looks good to me and seems to meet the criteria we decided on as a team 🙂

code/go/internal/validator/modes/modes.go
teresaromero: I've kept it for the shake of migrating. I felt it was safer to have all options available when using it on elastic-package. We could have a followup cleanup issue to remove it.

code/go/internal/validator/semantic/validate_no_embedded_ecs.go
teresaromero: i understood it was some validation needed for the build bundle... so that is why i addressed as a build validation rule. @jsoriano perhaps i have not understood correctly what you meant with treating it as "one off" ? should this not be validated?

code/go/internal/validator/spec.go
teresaromero: I accepted to have the Legacy as default to reinforce the idea that the validators in the "new" form should explicitly pass source or build. 

yup, i still have to address 🐰 feedback. thanks!

jsoriano (commented):
Could we split this change in two, or more? At least one only with the new validation API, and then another one, or more, for the new semantic validations.

Let's confirm first the new API, and then we can eventually introduce the changes in the validations.

teresaromero:
closing this in favor of:

- https://github.com/elastic/package-spec/pull/1177
- https://github.com/elastic/package-spec/pull/1178 (follow-up, draft until first is merged)

cc @jsoriano

PR #1177, opened 2026-06-02
validator: introduce mode-aware constructors and validation API

### LINKED ISSUE #549: [Change Proposal] Add different validation modes for source and built packages

teresaromero:
## What does this PR do?

Introduces a mode-aware validation API in `pkg/validator` and wires validation mode through the internal `Spec`.

**Public API (`code/go/pkg/validator/validator.go`)**

- Adds `Mode` (re-exported from internal) with three values: `ModeLegacy`, `ModeSource`, and `ModeBuild`.
- Adds `Validator` configured via `NewValidator(mode Mode, opts ...Option) (*Validator, error)`.
- Adds functional option `WithWarningsAsErrors(bool)` to control warning promotion independently of the `PACKAGE_SPEC_WARNINGS_AS_ERRORS` environment variable.
- Adds mode-aware methods on `*Validator`:
  - `ValidateFromPath(path string)` — wraps the disk FS per mode (`linkedfiles.NewFS` for legacy/source, `linkedfiles.NewBlockFS` for build).
  - `ValidateFromZip(zipPath string)` — supported in `ModeLegacy` and `ModeBuild` only; always validates through `BlockFS`.
  - `ValidateFromFS(location string, fsys fs.FS)` — validates caller-supplied FS with mode-specific linked-file semantics and rejects incompatible FS/mode pairings (e.g. `linkedfiles.FS` in build mode, `BlockFS` in source mode).
- Deprecates the existing package-level functions `ValidateFromPath`, `ValidateFromZip`, and `ValidateFromFS`; each now delegates to `NewValidator(ModeLegacy)` and preserves legacy behavior for existing callers.

**Internal plumbing (`code/go/internal/validator/`)**

- Adds `Mode` type and `Valid()` in `modes.go`.
- Extends `NewSpec(version, mode)` and stores mode on `Spec`.
- Extends `Spec.rules()` with an optional per-rule `modes []Mode` filter so follow-up PRs can gate semantic rules without further API changes.

**Changelog**

- Adds a `3.7.0-next` entry for mode-aware constructors and validation APIs.

Invalid mode values (including zero-value `Mode("")`) are rejected at `NewValidator` construction time.

## Why is it important?

The validator currently has no concept of validation mode — all callers get the same rules regardless of whether they are validating a source checkout or a built artifact. This PR lays the API foundation so future PRs can attach mode-specific rules (and enforce distinct source vs. build package shapes per #549) without breaking existing callers.

Existing integrations (`elastic-package`, etc.) continue to work via the deprecated entry points, which run in `ModeLegacy`.

## Checklist

- [ ] ~~I have added test packages to [`test/packages`](https://github.com/elastic/package-spec/tree/main/test/packages) that prove my change is effective.~~

## Related issues

- Closes #549 (partially — API foundation and linked-file mode behavior; mode-gated semantic rules and caller migration follow in subsequent PRs)

## Tests

New and updated tests cover the mode-aware API:

| Test | File | What it verifies |
|------|------|------------------|
| `TestValid` | `internal/validator/modes_test.go` | `Mode.Valid()` accepts legacy/source/build and rejects invalid/empty strings |
| `TestNewValidator_RejectsInvalidMode` | `pkg/validator/validator_test.go` | Invalid mode strings are rejected at construction |
| `TestLinksBehaviorAcrossModes` | `pkg/validator/validator_test.go` | Build rejects `.link` files; source and legacy accept them on disk |
| `TestValidateFromFS_rejectsIncompatibleFS` | `pkg/validator/validator_test.go` | Build + `linkedfiles.FS` and source + `BlockFS` are rejected; compatible pairings succeed |
| `TestValidateFromFS_legacyPlainFSBlocksLinks` | `pkg/validator/validator_test.go` | Legacy auto-wraps plain FS with `BlockFS`; legacy + `linkedfiles.FS` resolves links |
| `TestValidateFromZip_modeRestrictions` | `pkg/validator/validator_test.go` | Zip validation rejected in source mode; allowed in legacy and build |
| `TestValidateFromZip_validatesPackage` | `pkg/validator/validator_test.go` | Valid zip succeeds; malformed zip root layout errors |
| `TestDeprecatedValidateFromZip` | `pkg/validator/validator_test.go` | Deprecated `ValidateFromZip` wrapper still works |
| `TestWithWarningsAsErrors_option` | `pkg/validator/validator_test.go` | Option overrides env var for warning promotion |
| `TestLinksAreBlocked` | `pkg/validator/validator_test.go` | Deprecated `ValidateFromFS` with plain mock FS blocks links (legacy behavior) |

Uses existing test package `test/packages/with_links` for link-file scenarios. Internal `spec_test.go` updated for `NewSpec(..., ModeLegacy)`.

## Follow-ups / open questions

- **`ModeLegacy` vs `ModeSource`**: kept as distinct modes for migration safety; may collapse once source-only semantics are fully defined and callers have migrated.
- **Mode-gated semantic rules**: infrastructure is in place (`modes []Mode` on rule registration); individual rules are not yet filtered — that lands in a follow-up PR.
- **Caller migration**: `elastic-package` and other consumers still use deprecated entry points; reference migration to `NewValidator(ModeSource|ModeBuild)` is planned separately.


code/go/pkg/validator/api_test.go
```diff
@@ -0,0 +1,192 @@
...
+func TestLegacyPreservation_FromZip(t *testing.T) {
+	packagePath := filepath.Join(testPackagesDir, "good")
+	zipPath := createPackageZip(t, packagePath)
+
+	// NewFromZip always validates in ModeBuild (zip = built package).
+	// Constructor must succeed even though the source fixture may not pass
+	// build-mode validation.
+	_, err := NewFromZip(zipPath)
+	require.NoError(t, err)
+}
```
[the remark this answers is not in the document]
teresaromero: Fixed. The test was renamed to `TestNewFromZip_ConstructorSucceeds` to reflect what it actually does. A proper legacy-preservation comparison was added but ultimately removed because `ValidateFromZip` was simplified to delegate directly to `NewFromZip` — making the comparison tautological. The zip path is now covered by the constructor smoke test and `TestNewFromFS_BuildModeBlocksLinksEvenWithPreWrappedLinkedFS` for behavioural coverage.

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,183 @@
...
+	// that must be released after Validate returns.
+	closer io.Closer
+}
+
+// Option configures a Validator.
+type Option func(*Validator)
+
+// WithWarningsAsErrors makes validation warnings count as errors when v is true.
+func WithWarningsAsErrors(v bool) Option {
+	return func(va *Validator) { va.warningsAsErrors = v }
```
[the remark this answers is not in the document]
teresaromero: Partially addressed. `va` was renamed to `v` (idiomatic Go for option-closure receivers) and the `WithWarningsAsErrors` parameter renamed to `enabled` for clarity. `fsys` was intentionally kept — it is the established convention throughout this codebase and renaming it introduced more noise than signal.

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	var fsys fs.FS
+	if mode == ModeBuild {
+		fsys = linkedfiles.NewBlockFS(os.DirFS(packageRootPath))
+	} else {
+		// ModeLegacy and ModeSource: resolve linked files transparently.
+		fsys = linkedfiles.NewFS(packageRootPath, os.DirFS(packageRootPath))
+	}
+	return buildValidator(mode, packageRootPath, fsys, nil, opts), nil
```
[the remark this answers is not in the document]
teresaromero: Fixed. `NewFromPath` and `NewFromFS` now call `mode.Valid()` at the top and return `fmt.Errorf("invalid validation mode %q", mode)` immediately. `NewFromZip` is hardcoded to `ModeBuild` so no guard is needed there. Covered by `TestNewFromPath_RejectsInvalidMode` and `TestNewFromFS_RejectsInvalidMode`.

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, fmt.Errorf("invalid package filesystem at %q: root is not a directory", location)
+	}
+
+	if _, ok := fsys.(*linkedfiles.FS); !ok {
+		if mode == ModeSource {
+			fsys = linkedfiles.NewFS(location, fsys)
+		} else {
+			fsys = linkedfiles.NewBlockFS(fsys)
+		}
+	}
```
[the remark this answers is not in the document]
teresaromero: Partially fixed, with a correction. The original fix unconditionally applied `BlockFS` for both `ModeLegacy` and `ModeBuild` when a `*linkedfiles.FS` was passed. On review this was wrong for `ModeLegacy`: the old `validateFromFS` preserved a pre-wrapped `*linkedfiles.FS` (links resolved), only blocking plain filesystems. The fix was corrected to match that behaviour: `ModeLegacy` preserves `*linkedfiles.FS`, blocks everything else; `ModeBuild` always applies `BlockFS` unconditionally. Covered by `TestNewFromFS_BuildModeBlocksLinksEvenWithPreWrappedLinkedFS` and `TestNewFromFS_LegacyAndSourceModePreservePreWrappedLinkedFS`.

teresaromero:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > code/go/pkg/validator/api_test.go (1)> `29-54`: _🛠️ Refactor suggestion_ | _🟠 Major_ | _⚡ Quick win_
> > > **Rename `pkgName` to `packageName` per naming guideline.**
> > > `pkgName` (lines 29, 32, 54) is the exact abbreviation called out in the guidelines. Use the full name.
> > > 
> > > ♻️ Proposed rename
> > > -	pkgName := filepath.Base(packagePath)
> > > +	packageName := filepath.Base(packagePath)
> > >  
> > >  	tmpDir := t.TempDir()
> > > -	zipPath := filepath.Join(tmpDir, pkgName+".zip")
> > > +	zipPath := filepath.Join(tmpDir, packageName+".zip")
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > -		entryName := filepath.ToSlash(filepath.Join(pkgName, rel))
> > > +		entryName := filepath.ToSlash(filepath.Join(packageName, rel))
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > As per coding guidelines: "Go variable and function names must use full names, not abbreviations (e.g., `packageName` not `pkgName`...)".
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@code/go/pkg/validator/api_test.go` around lines 29 - 54, Rename the
> > > abbreviated variable pkgName to packageName throughout this test (declaration
> > > and all uses) so it follows naming guidelines; specifically change the value
> > > assignment using filepath.Base(packagePath), update the zipPath construction
> > > that uses pkgName+".zip", and update the entryName computation that joins
> > > pkgName with rel (and any other references such as zip entry prefixing) to use
> > > packageName instead, ensuring you update variable references in os.Create,
> > > filepath.Join, and filepath.ToSlash calls within api_test.go.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In `@code/go/pkg/validator/api_test.go`:
> - Around line 29-54: Rename the abbreviated variable pkgName to packageName
> throughout this test (declaration and all uses) so it follows naming guidelines;
> specifically change the value assignment using filepath.Base(packagePath),
> update the zipPath construction that uses pkgName+".zip", and update the
> entryName computation that joins pkgName with rel (and any other references such
> as zip entry prefixing) to use packageName instead, ensuring you update variable
> references in os.Create, filepath.Join, and filepath.ToSlash calls within
> api_test.go.
> ```
> 
> ℹ️ Review info

i will keep it as is thanks!

code/go/internal/validator/modes.go
jsoriano, line 1: It uses to be better to keep flatter packages. This file could belong directly to validator, and the modes could be LegacyMode, SourceMode and so on.
It is also easier for callers, the only need to import a single package, validator and use it together with other methods, like `validator.NewSpec(semver, validator.BuildMode)`.
This is actually done like this in the public package.

code/go/internal/validator/spec.go
```diff
@@ -44,8 +47,16 @@ type validationRules []validationRule
 // GASpecCheckVersion represents minimum version to start checking for unreleased version of the spec
 var GASpecCheckVersion = semver.MustParse("3.0.1")
 
-// NewSpec creates a new Spec for the given version
-func NewSpec(version semver.Version) (*Spec, error) {
+// NewSpec creates a new Spec for the given version and validation mode.
+// If mode is empty, it defaults to modes.Legacy.
+func NewSpec(version semver.Version, mode modes.Mode) (*Spec, error) {
```
jsoriano: Design question. Is the mode a property of the `Spec`?

A single spec for a given version can in principle validate in both modes, as well as it can validate any package type it supports.

If the mode is a property of the `Spec`, we would have the current API:
```
spec, _ := NewSpec(version, mode)
errs := spec.ValidatePackage(package)
```

If a single `Spec` can validate both modes, the API would be like this:
```
spec, _ := NewSpec(version)
errs := spec.ValidatePackage(package, mode)
```

Not sure what option is better, I think second option looks more natural, a spec could validate in both modes.
For the current uses we have it doesn't matter a lot, so we can keep it as is if you prefer, in any case this is internal and we can change it later.

code/go/pkg/validator/api.go
jsoriano, line 1: Nit. "api" is redundant here, any public method is going to be part of the package API 🙂

This could belong to `validator.go`, as it actually defines the `Validator` object.

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,203 @@
...
+// WithWarningsAsErrors makes validation warnings count as errors when enabled is true.
+func WithWarningsAsErrors(enabled bool) Option {
+	return func(v *Validator) { v.warningsAsErrors = enabled }
+}
+
+// NewFromPath returns a Validator for the package rooted at packageRootPath.
+//
+// For ModeLegacy and ModeSource the filesystem honours linked (.link) files;
+// for ModeBuild linked files are blocked (matching a built package artifact).
+func NewFromPath(mode Mode, packageRootPath string, opts ...Option) (*Validator, error) {
```
jsoriano: Nit. Configuration structs are usually easier to discover than functional options, and use to be simpler.
```suggestion
type Config struct {
    WarningsAsErrors bool
}

func NewFromPath(mode Mode, packageRootPath string, config Config) (*Validator, error) {
```

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,203 @@
...
+//
+// Linked-file handling depends on mode:
+//   - ModeSource: if fsys is not already a *linkedfiles.FS it is wrapped with
+//     linkedfiles.NewFS so .link files are resolved transparently.
+//   - ModeLegacy: a pre-wrapped *linkedfiles.FS is preserved as-is (links
+//     resolved); any other filesystem is wrapped with BlockFS. This matches the
+//     behaviour of the deprecated ValidateFromFS.
+//   - ModeBuild: BlockFS is always applied, even if fsys is already a
+//     *linkedfiles.FS, so linked files are unconditionally rejected.
+func NewFromFS(mode Mode, location string, fsys fs.FS, opts ...Option) (*Validator, error) {
```
jsoriano: All `NewFrom*` methods are structurally similar:
* Check the validity of the mode.
* Get an FS.
* Decide how to handle links.

Specially the last part is repeated, the different logic we have in the different methods is only cutting cases.
Think for example on the zip case, if we apply the logic in this function we would get the same result, as it just uses `ModeBuild`.

It would be better to let `NewFromPath` and `NewFromZip` do only what is particular to them (getting the FS, hard-coding BuildMode in the zip case), and call at some point `NewFromFS`, that peforms the common logic in a single place.

.gitignore
```diff
@@ -6,3 +6,5 @@ fuzz
 /build/
 .vscode/
 .DS_Store
+
+.scratch/
```
jsoriano: What is this? Consider using `~/.gitignore_global` instead.

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	if !info.IsDir() {
+		return nil, fmt.Errorf("invalid package filesystem at %q: root is not a directory", location)
+	}
+
+	_, isLinkedFS := fsys.(*linkedfiles.FS)
+	switch mode {
+	case ModeSource:
+		if !isLinkedFS {
+			fsys = linkedfiles.NewFS(location, fsys)
```
jsoriano: On what cases we receive a linked FS here?

Nit. Maybe `NewFS` and `NewBlockFS` should decide what to do when called with a `linkedfiles` FS of each kind, so their callers don't need to care.

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	case ModeBuild:
+		// Always block, even a pre-wrapped *linkedfiles.FS.
+		fsys = linkedfiles.NewBlockFS(fsys)
```
jsoriano: Umm, not sure if we should do this at this level. Could we instead reject `.lnk` files in built packages as we will reject `_dev` files?

Or we need to do this because the spec always receives the link files resolved?

code/go/pkg/validator/api.go
```diff
@@ -0,0 +1,203 @@
...
+	default: // ModeLegacy
+		// Preserve a pre-wrapped *linkedfiles.FS; block everything else.
+		// Matches the old validateFromFS behaviour.
```
jsoriano: Not sure if this would be needed now. I think this was used before because `NewFromPath` generated its own `linkedfiles.FS`. But not sure if this is the case now.

jsoriano (commented):
Overall looks good. Questioning some small design decisions and doing some nitpicking.

code/go/pkg/validator/api.go
teresaromero: This config is used for the warnings, which is read from the env, so i removed the options/config and read it directly from the env. Changed the test case too to use t.Env instead of a local variable.

```
v := &Validator{
		mode:             mode,
		location:         location,
		fsys:             fsys,
		warningsAsErrors: common.IsDefinedWarningsAsErrors(),
		closer:           closer,
}
```

code/go/internal/validator/modes.go
teresaromero: the modes are also used at the internal/spec so this was here to avoid importing it from pkg/validator. Looking it again i am having doubts on the implementation at spec. So i will work on this to change it. thanks

code/go/internal/validator/spec.go
teresaromero: my train of thought was that the spec was defined by the mode, as the definition needs to be source or build, and i did not think of a spec that can be both at the same time. so a package is either a built bundle or a source bundle. 
with this the yml spec gets more broad... and the semantic validation enables the spec to be one or the other, if that makes sense. so i see the first one fitting this. 
with this said i think internal/spec should export the BuildSpec and SourceSpec (not the modes) and we might need to directly create a public function to use one mode or the other but not expose the modes perse. i am giving this a second thought as i did not see it this way until now :D

code/go/pkg/validator/api.go
teresaromero: i am not sure i understand this, or the lines have moved around. here i am validating the fsys, not any linked files. The fs is checked dependeing on the mode "selected" if it allows or not .link files. For build validation there link should be resolved.

code/go/pkg/validator/api.go
teresaromero: i need to review this whole refactor, as i am thinking it will be better to have direct constructors for modes, instead of using the mode as a switch...

teresaromero (commented):
thanks for the feedback

teresaromero:
@jsoriano i've changed the aproach, having in mind the mode belongs to the spec as a compliment of the definition. this is ready for re-review. thnks

code/go/pkg/validator/limits_test.go
```diff
@@ -116,7 +118,11 @@ func TestLimitsValidation(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.title, func(t *testing.T) {
 			t.Parallel()
-			err := ValidateFromFS("test-package", c.fsys)
+
+			specFn := func(version semver.Version) (*validator.Spec, error) {
+				return validator.NewLegacySpec(version)
+			}
+			err := validateFromFS("test-package", c.fsys, specFn)
```
jsoriano: `validator.NewLegacySpec` already has this signature, isn't it? 🤔
```suggestion
			specFn := validator.NewLegacySpec
			err := validateFromFS("test-package", c.fsys, specFn)
```

code/go/pkg/validator/validator.go
```diff
@@ -11,21 +11,61 @@ import (
...
 func ValidateFromPath(packageRootPath string) error {
 	// We wrap the fs.FS with a linkedfiles.LinksFS to handle linked files.
 	linksFS := linkedfiles.NewFS(packageRootPath, os.DirFS(packageRootPath))
-	return ValidateFromFS(packageRootPath, linksFS)
+
+	legacySpec := func(version semver.Version) (*validator.Spec, error) {
+		return validator.NewLegacySpec(version)
+	}
+
+	return validateFromFS(packageRootPath, linksFS, legacySpec)
```
jsoriano: These functions already have the expected signature, no need to wrap them.
```suggestion
	return validateFromFS(packageRootPath, linksFS, validator.NewLegacySpec)
```

code/go/pkg/validator/validator.go
```diff
@@ -11,21 +11,61 @@ import (
...
+// This function uses the build specification, appropriate for packages produced by
+// elastic-package build, distributed as zip files, or served by the package registry.
+// Linked files (.link) are blocked; source-only artifacts are rejected.
+func ValidateFromBuildPath(packageRootPath string) error {
+	fs := os.DirFS(packageRootPath)
+
+	buildSpec := func(version semver.Version) (*validator.Spec, error) {
+		return validator.NewBuildSpec(version)
+	}
+	return validateFromFS(packageRootPath, fs, buildSpec)
```
jsoriano:
```suggestion
	return validateFromFS(packageRootPath, fs, validator.NewBuildSpec)
```

code/go/pkg/validator/validator.go
```diff
@@ -11,21 +11,61 @@ import (
...
+// Linked (.link) files are resolved transparently.
+func ValidateFromSourcePath(packageRootPath string) error {
+	// We wrap the fs.FS with a linkedfiles.LinksFS to handle linked files.
+	linksFS := linkedfiles.NewFS(packageRootPath, os.DirFS(packageRootPath))
+
+	sourceSpec := func(version semver.Version) (*validator.Spec, error) {
+		return validator.NewSourceSpec(version)
+	}
+
+	return validateFromFS(packageRootPath, linksFS, sourceSpec)
```
jsoriano:
```suggestion
	return validateFromFS(packageRootPath, linksFS, validator.NewSourceSpec)
```

code/go/pkg/validator/validator.go
```diff
@@ -46,16 +86,16 @@ func ValidateFromZip(packagePath string) error {
...
-}
+	buildSpec := func(version semver.Version) (*validator.Spec, error) {
+		return validator.NewBuildSpec(version)
+	}
 
-// ValidateFromFS validates a package against the appropiate specification and returns any errors.
-// Package files are obtained through the given filesystem.
-func ValidateFromFS(location string, fsys fs.FS) error {
-	return validateFromFS(location, fsys, common.IsDefinedWarningsAsErrors())
+	return validateFromFS(packagePath, subDir, buildSpec)
```
jsoriano:
```suggestion
	return validateFromFS(packagePath, subDir, validator.NewBuildSpec)
```

code/go/internal/validator/spec.go
jsoriano: Ok, so we go with the first option, the mode is part of a given spec, so it has a single validation mode.

Only instead of
```
spec, _ := NewSpec(version, mode)
errs := spec.ValidatePackage(package)
```

We have specific methods for each mode:
```
spec, _ := NewSourceSpec(version)
errs := spec.ValidatePackage(package)
```

To me the option of leaving the spec defined only by the version, with different methods for the different validation modes still look more natural.

But given that we have a reduced set of modes your option also looks good.

Thanks for the discussion.

code/go/pkg/validator/api.go
jsoriano: Part of the original design in https://github.com/elastic/package-spec/issues/549 was actually to introduce the "Option C" API shape, a new "Validator" object that can be somehow customized with different options.

With current code we are back to something like "Option A": methods with the mode as parameter. Only that we have one method per mode instead of having the mode as a parameter.

I still think it would be valuable to introduce this "Validator" object. Having the object we can add configuration options later without needing to add more functions to the package, or parameters to these functions.

For this initial version it could have `warningsAsErros` as single configuration, and it could have two validation methods (build and source) if we prefer this approach instead of modes as a parameter.

code/go/pkg/validator/api.go
jsoriano: When validating a built package there should not be any `.link` files.

I was wondering how we are ensuring that, and if the different FSs used was because of that.

jsoriano (commented):
Thanks, it looks simpler now.

Some notes:
* Using specific methods for the validation modes LGTM, at the end I don't think we are going to have more modes beyond Source and Build.
* With the last changes we are losing the "Option C" API shape, that would be a "Validator" object that could have different settings, so we don't need to add additional methods or parameters to these methods if we need more settings in the future.
* Please check that links are not allowed in built packages. They should have been resolved when building the package.

code/go/pkg/validator/limits_test.go
teresaromero: yes 🤦🏻‍♀️  its redundant, i will change it thanks

code/go/internal/validator/modes.go
jsoriano: Could this be moved directly from `internal/validator/models` to `internal/validator` then?

code/go/pkg/validator/validator.go
```diff
@@ -8,77 +8,161 @@ import (
...
-	if _, ok := fsys.(*linkedfiles.FS); !ok {
-		fsys = linkedfiles.NewBlockFS(fsys)
+// Validate runs package validation and returns any errors encountered.
+func (v *Validator) Validate() (err error) {
+	if v.closer != nil {
+		defer func() {
+			if cerr := v.closer.Close(); cerr != nil {
+				err = errors.Join(err, cerr)
+			}
+		}()
```
jsoriano: `Validate` should not close the underlying filesystem. It would be an unexpected side effect.

If `Validator` creates its own filesystems or resources that need to be cleaned up, it should expose its own `Close` method.

code/go/pkg/validator/validator.go
```diff
@@ -8,77 +8,161 @@ import (
...
+// NewFromPath returns a Validator for the package rooted at packageRootPath.
+// The mode determines which validation rules apply and how linked files are
+// handled: ModeBuild blocks linked files, ModeSource and ModeLegacy resolve them.
+// Returns an error if mode is not a recognised value.
+func NewFromPath(mode Mode, packageRootPath string, opts ...Option) (*Validator, error) {
+	if !mode.internal.Valid() {
+		return nil, fmt.Errorf("invalid validation mode %q", mode.internal)
 	}
-	defer r.Close()
+	fsys := mode.wrapFS(packageRootPath, os.DirFS(packageRootPath))
```
jsoriano: Nit. If this wrapFS is only used in this case, it would simplify things to handle it only here.
```suggestion
    fsys := os.DirFS(packageRootPath)
    if mode == ModeBuild {
        fsys = linkedfiles.NewBlockFS(fsys)
    } else {
        fsys = linkedfiles.NewFS(packageRootPath, fsys)
    }
```

code/go/pkg/validator/validator.go
```diff
@@ -8,77 +8,161 @@ import (
...
 	}
 
-	return ValidateFromFS(packagePath, subDir)
+	fsys := linkedfiles.NewBlockFS(subDir)
+	return buildValidator(ModeBuild, zipPath, fsys, r, opts), nil
 }
 
-// ValidateFromFS validates a package against the appropiate specification and returns any errors.
-// Package files are obtained through the given filesystem.
-func ValidateFromFS(location string, fsys fs.FS) error {
```
jsoriano, line 54: This method is removed, right? This would be a breaking change in the public API.

teresaromero:
> * With the last changes we are losing the "Option C" API shape, that would be a "Validator" object that could have different settings, so we don't need to add additional methods or parameters to these methods if we need more settings in the future.


> * Please check that links are not allowed in built packages. They should have been resolved when building the package.

I've reintroduced the Validator struct. in order to have the modes as public there is the Mode type which owns the fs for each mode. This way fs are determined. I've kept `NewFromFS` as a constructor with the fs open for the user to use under its responsability

code/go/pkg/validator/validator.go
```diff
@@ -8,77 +8,161 @@ import (
...
+	return v
 }
 
-func validateFromFS(location string, fsys fs.FS, warningsAsErrors bool) error {
-	// If we are not explicitly using the linkedfiles.FS, we wrap fsys with
-	// a linkedfiles.BlockFS to block the use of linked files.
-	if _, ok := fsys.(*linkedfiles.FS); !ok {
-		fsys = linkedfiles.NewBlockFS(fsys)
+// Validate runs package validation and returns any errors encountered.
+func (v *Validator) Validate() (err error) {
```
jsoriano: Going back to the API design question. Sorry for the back and forth.

With current design we have the object, but given that it already receives everything in the constructor, it is just like a function with delayed execution.

These two shapes are really not very different, and the first one has more complexity with little value:
```go
v, err := validator.NewFromZip(path, options...)
err = v.Validate()
```
And:
```go
err := validator.ValidateFromZip(path, options...)
```

What would give more value to the API shape, and what I'd say the "Option C" was, is to have a stateless validator that clearly separates validation configuration from the validation of one, or more packages.

Something like this:
```go
// Configuration-time
v, err := validator.New(validator.ModeSource,
  validator.WithWarningsAsErrors(true),
  validator.WithFilter(filterConfig),  // future
  validator.WithLogger(log),           // future
)


// Verb-time, repeatable, no shared mutable state
err = v.ValidateFromPath("/pkg1")
err = v.ValidateFromPath("/pkg2")
err = v.ValidateFromZip("/pkg.zip")
err = v.ValidateFromFS("loc", fsys)
```

code/go/pkg/validator/validator.go
teresaromero: i initially had this if/else but i thought to enforce the fs is owned by the mode, so wrapFs represents the fs asigned by the mode... so this constructor just uses whatever the mode has and is not responsible of creating it, if that makes sense? happy to change it if this is not correct

code/go/pkg/validator/validator.go
teresaromero: `ValidateFromFS` although is a public method is just being used at tests files. I've reviewed elastic-package and dont see this function used explicitly so, its a breaking change because is no longer available but is not being used 🤔  I will revert and keep it, although i think we should deprecate it in favour of NewFromFS

code/go/pkg/validator/validator.go
jsoriano: Functionality seems to be correct, the only issue I see is that it is adding complexity with an abstraction that is only used in a place.

In general all this handling of links gives me a bit of bad feeling, but I don't have better ideas.

In any case all these are internal details, so let it with the approach you prefer.

jsoriano: If you keep modes as objects with `wrapFS` methods I would suggest to convert `Mode` in an interface, and make `wrapFS` a method of the implementing objects.
```go
type Mode interface {
    mode() modes.Mode
    wrapFS(location string, fsys fs.FS) fs.FS
}

type SourceMode struct {}
func (SourceMode) mode() modes.Mode { return mode.Source }
func (SourceMode) wrapFS(location string, fsys fs.FS) fs.FS {
    return linkedfiles.NewFS(location, fsys)
}

...
```

This looks more aligned with Go conventions.

code/go/pkg/validator/validator.go
jsoriano: We should deprecate all `ValidateFrom` methods at the package level, and keep in the long term only the `Validator` object.
In the meantime let's keep all of them, just to follow good practices, just in case.

code/go/pkg/validator/validator.go
```diff
@@ -13,29 +13,83 @@ import (
...
+// WithWarningsAsErrors controls whether validation warnings are promoted to errors.
+// When enabled is true, warnings are reported as errors regardless of the
+// PACKAGE_SPEC_WARNINGS_AS_ERRORS environment variable. When enabled is false,
+// warnings remain warnings even if the environment variable is set.
+func WithWarningsAsErrors(enabled bool) Option {
+	return func(v *Validator) { v.warningsAsErrors = enabled }
+}
+
+// NewValidator creates a Validator for the given mode and options.
+func NewValidator(mode Mode, opts ...Option) (*Validator, error) {
```
jsoriano: `NewValidator` is redundant now, as the package is already called validator. This could be directly `New`.
```suggestion
// New creates a Validator for the given mode and options.
func New(mode Mode, opts ...Option) (*Validator, error) {
```

code/go/pkg/validator/validator.go
```diff
@@ -46,39 +100,74 @@ func ValidateFromZip(packagePath string) error {
...
+		// If we are not explicitly using the linkedfiles.FS, we wrap fsys with
+		// a linkedfiles.BlockFS to block the use of linked files.
+		if _, ok := fsys.(*linkedfiles.FS); !ok {
+			fsys = linkedfiles.NewBlockFS(fsys)
+		}
+	} else if _, ok := fsys.(*linkedfiles.FS); ok && v.mode == ModeBuild {
+		return errors.New("linked files are not supported in ModeBuild")
+	} else if _, ok := fsys.(*linkedfiles.BlockFS); ok && v.mode == ModeSource {
+		return errors.New("block linked files are not supported in ModeSource")
+	}
```
jsoriano, line 122: I still don't like the handling of linked file systems, but this is internal and we can revisit later, I don't have much better ideas now.

code/go/pkg/validator/validator.go
```diff
@@ -13,29 +13,83 @@ import (
 
 	"github.com/elastic/package-spec/v3/code/go/internal/linkedfiles"
 	"github.com/elastic/package-spec/v3/code/go/internal/packages"
-	"github.com/elastic/package-spec/v3/code/go/internal/validator"
+	internalvalidator "github.com/elastic/package-spec/v3/code/go/internal/validator"
```
jsoriano: Nit. We don't need to change this, right?

code/go/pkg/validator/validator.go
```diff
@@ -13,29 +13,83 @@ import (
...
+// Mode is the validation context that controls semantic rules and linked-file handling.
+type Mode = internalvalidator.Mode
+
+var (
+	// ModeLegacy preserves the original validation behavior.
+	ModeLegacy Mode = internalvalidator.ModeLegacy
+	// ModeSource validates a checked-out source tree.
+	ModeSource Mode = internalvalidator.ModeSource
+	// ModeBuild validates a built package artifact.
+	ModeBuild Mode = internalvalidator.ModeBuild
```
jsoriano: It would more idiomatic, and reads more natural, to use Mode as suffix.
```suggestion
	// LegacyMode preserves the original validation behavior.
	LegacyMode Mode = internalvalidator.LegacyMode
	// ModeSource validates a checked-out source tree.
	SourceMode Mode = internalvalidator.SourceMode
	// BuildMode validates a built package artifact.
	BuildMode Mode = internalvalidator.BuildMode
```

code/go/pkg/validator/validator.go
```diff
@@ -46,39 +100,74 @@ func ValidateFromZip(packagePath string) error {
...
 		return errors.New("could not determine specification version for package")
 	}
 
-	spec, err := validator.NewSpec(*pkg.SpecVersion)
+	spec, err := internalvalidator.NewSpec(*pkg.SpecVersion, v.mode)
 	if err != nil {
 		return err
 	}
-	spec.WarningsAsErrors = warningsAsErrors
+	spec.WarningsAsErrors = v.warningsAsErrors
```
jsoriano, line 141: Maybe we should append a technical preview warning when Source or Build modes are used, at least till we complete the validators.

jsoriano (commented):
Looks good overall, added some comments about some small details.

code/go/pkg/validator/validator.go
teresaromero: I've added just using the log pkg as i havent seen any logger implemented on the package https://github.com/elastic/package-spec/pull/1177/changes/d895e496b1d4ebc2e6685135382e2f3f80b88b14 

i followed the same pattern as in https://github.com/elastic/package-spec/blob/56193f9315aec38b5403f8ede03edd3770f6877c/code/go/internal/validator/semantic/warning.go#L27

jsoriano: Yeah, we don't have an abstraction for warnings, maybe we should add it.

We could leverage `warningsAsErrors`, something like this:
```go
errs := spec.ValidatePackage(*pkg)
if v.mode != LegacyMode {
    err := specerrors.NewStructuredErrorf("validation mode '%s' is in technical preview", v.mode)
    if v.warningsAsErrors() {
        errs = append(errs, err)
    } else {
        log.Printf("Warning: %s", err.Message())
    }
}
if len(errs) > 0 {
    return errs
}
return nil
```

jsoriano (dismissed):

code/go/pkg/validator/validator.go
teresaromero: opened issue for tracking this refactor https://github.com/elastic/package-spec/issues/1185

code/go/pkg/validator/validator.go
teresaromero: updated on last commit

code/go/pkg/validator/validator.go
jsoriano: Thanks!

spec/changelog.yml
```diff
@@ -8,6 +8,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add support for mode-aware constructors and validation APIs.
```
jsoriano, line 11: Nit. Maybe we should be more specific about the modes we are referring to.
```suggestion
    - description: Add different validators for source and built packages.
```

jsoriano (approved):
LGTM, thanks!

spec/changelog.yml
teresaromero: i think that will be the next pr where the validators are actually mapped for source and built. this pr adds the validator api but all validators are now the same

PR #1178, opened 2026-06-02
Validation rules: build/source mode enforcement

### LINKED ISSUE #549: [Change Proposal] Add different validation modes for source and built packages

teresaromero:
## What does this PR do?

Wires concrete semantic validation rules to the `source` / `build` / `legacy` modes introduced in #1177.

**New build-mode rules** — reject source artifacts that must not survive the build:

| Rule | Rejects |
|---|---|
| `validationMode: source` in spec YAML (folder validator) | any `_dev/` directory in build mode |
| `.link` check in folder validator | any `.link` file in build mode |
| `ValidateNoExternalFields` | fields with `external: ecs` (build mode) |
| `ValidateStreamInputBundled` | streams with `package:` or missing `input:`; policy template inputs with `package:` instead of `type:` (build mode, integration type) |

**New source-mode rule:**

| Rule | Rejects |
|---|---|
| `ValidateNoEmbeddedEcsInDynamicTemplates` | `_embedded_ecs` keys in `dynamic_templates` (injected by the build tooling, must not appear in source; integration type) |

**Re-scoped existing rules:**

| Rule | Before | After |
|---|---|---|
| `ValidateExternalFieldsWithDevFolder` | all modes | `legacy` + `source` |
| `ValidateIntegrationInputQualifier` | all modes | `legacy` + `build` |
| `ValidateTestPackageRequirements` | all modes | `legacy` + `source` |

**Composable package false-positive fixes** (source mode, entries that set `package:`):
- `ValidateIntegrationPolicyTemplates`: skip template-existence check when no local `template_path` is declared and `package:` is set — the template comes from the dependency.
- `spec/integration/manifest.spec.yml` and `spec/integration/data_stream/manifest.spec.yml`: `vars` base schema relaxed to `array of objects`; strict `$ref` validation restored conditionally via `if/then` for non-composable entries only.
- Added `required: [type]` guards to `if` blocks for `password`, `select`, and `duration` var conditionals — prevents vacuous matches when `type` is absent (composable vars that omit `type`).
- Resolved stale `TODO: Allow _embedded_ecs only on built packages` comment in the schema.

**`validationMode` spec field** replaces the old boolean `sourceOnly` field with a string (`"source"`, `"build"`, or omitted for both modes), so spec directives can be scoped to either validation direction. All `_dev/` entries now declare `validationMode: source`.

## Why is it important?

`elastic-package` validates both source packages (contain `_dev/`, `.link` files, `external: ecs` references) and built packages (those artifacts are absent, build-injected ones appear instead). A unified rule set causes false positives in both directions. Composable packages (`package:` delegation to input packages) add a third dimension: templates and full var definitions only exist after the build materialises the dependency.

## Checklist

## Related issues

- Closes #549 (partially)
- Requires #1177

## Release Notes

* **New Features**
  * Introduced tighter `source` and `build` validation modes, with mode-aware checks.
  * Improved composable template handling: `package:`-based streams/inputs skip template-path validation when no `template_paths` are provided.

* **Bug Fixes & Validations**
  * Rejection of `_embedded_ecs` dynamic template keys in `source` mode.
  * `build` mode now blocks source-only filesystem artifacts (`_dev`, `.link`) and non-materialized/external ECS references.
  * Only actual `data_stream/*/` directories are considered for stream validation.

* **Documentation**
  * Updated integration/package schemas and changelogs to reflect the new mode rules.


test/built_packages/bad_built_missing_input/changelog.yml
```diff
@@ -0,0 +1,6 @@
+# newer versions go on top
+- version: "0.0.1"
+  changes:
+    - description: Initial draft of the package
```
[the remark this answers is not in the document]
teresaromero: these are test fixtures where this is irrelevant. not addressing it

test/packages/bad_embedded_ecs/changelog.yml
```diff
@@ -0,0 +1,5 @@
+- version: 0.0.1
+  changes:
+    - description: Initial release
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1
```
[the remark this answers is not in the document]
teresaromero: this is a test fixture where this is irrelevant

code/go/internal/validator/spec.go
```diff
@@ -252,10 +253,19 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
...
-		{fn: semantic.ValidateTestPackageRequirements, types: []string{"integration"}, since: semver.MustParse("3.6.0")},
+		{fn: semantic.ValidateTestPackageRequirements, types: []string{"integration"}, since: semver.MustParse("3.6.0"),
+			modes: []Mode{LegacyMode, SourceMode}},
+		{fn: semantic.ValidateNoEmbeddedEcsInDynamicTemplates, types: []string{"integration"},
+			modes: []Mode{SourceMode}},
+		{fn: semantic.ValidateNoDevFolder, modes: []Mode{BuildMode}},
+		{fn: semantic.ValidateNoLinkFiles, modes: []Mode{BuildMode}},
+		{fn: semantic.ValidateNoExternalEcs, modes: []Mode{BuildMode}},
+		{fn: semantic.ValidateStreamInputMaterialized, modes: []Mode{BuildMode},
+			types: []string{"integration"}},
```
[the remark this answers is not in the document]
teresaromero: the modes are transversal to all the versions, as packages are now being validated in source and build form. this is not applicable to versions, although it will be available since 3.7.0

code/go/internal/validator/semantic/types.go
```diff
@@ -352,9 +352,11 @@ func listDataStreams(fsys fspath.FS) ([]string, error) {
...
 	}
 
-	list := make([]string, len(dataStreams))
-	for i, dataStream := range dataStreams {
-		list[i] = dataStream.Name()
+	var list []string
+	for _, dataStream := range dataStreams {
+		if dataStream.IsDir() {
+			list = append(list, dataStream.Name())
+		}
```
jsoriano, line 359: What is the motivation of this change? Can we have anything under `dataStreamDir` that is not a directory?

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -110,6 +111,14 @@ func ValidateIntegrationPolicyTemplates(fsys fspath.FS) specerrors.ValidationErr
...
 func validateIntegrationPolicyTemplateInputs(fsys fspath.FS, policyTemplate integrationPolicyTemplate) error {
 	for _, input := range policyTemplate.Inputs {
+		// Composable inputs reference an input package via 'package:'. When no
+		// explicit template_path/template_paths is set, all templates come from
+		// the dependency and are only present after build. Skip those.
+		// If the composable input defines its own template_path or template_paths
+		// (overlay templates that live in the source package), those are validated.
+		if input.Package != "" && input.TemplatePath == "" && len(input.TemplatePaths) == 0 {
+			continue
+		}
```
jsoriano: Is this exception needed?

I see two options for that:
* We don't validate template paths on source packages. Then this validator is never run with `input.Package != ""`.
* We validate template paths on source packages. Then the value of `input.Package` should not matter. 

Also, unit tests pass without this code.
```suggestion
```

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -141,6 +150,17 @@ func validateAllDataStreamStreamTemplates(fsys fspath.FS, dsMap map[string]dataS
...
+			// no explicit template_path/template_paths is set on the stream, all
+			// templates come from the dependency and are only present after build.
+			// Skip those — ValidateStreamInputMaterialized enforces that 'package:'
+			// is replaced by 'input:' in build mode.
+			// However, if the composable stream defines its own template_path or
+			// template_paths, those files must exist in the source package and are
+			// validated here.
+			if s.Package != "" && s.TemplatePath == "" && len(s.TemplatePaths) == 0 {
+				continue
+			}
```
jsoriano: In this case I guess we need to maintain this condition if we want to validate templates at source packages. This case is different because a default template is expected if none is defined, and for composable packages it would be fine to don't have one in the source.

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -141,6 +150,17 @@ func validateAllDataStreamStreamTemplates(fsys fspath.FS, dsMap map[string]dataS
...
 		manifest := dsMap[dsDir]
 		for _, s := range manifest.Streams {
+			// Composable streams reference an input package via 'package:'. When
+			// no explicit template_path/template_paths is set on the stream, all
+			// templates come from the dependency and are only present after build.
+			// Skip those — ValidateStreamInputMaterialized enforces that 'package:'
+			// is replaced by 'input:' in build mode.
+			// However, if the composable stream defines its own template_path or
+			// template_paths, those files must exist in the source package and are
+			// validated here.
```
jsoriano: Nit. A bit too verbose.
```suggestion
			// Don't validate template paths if they use a package as input
			// and they don't define any template.
```

code/go/internal/validator/semantic/validate_no_dev_folder.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+)
+
+// ValidateNoDevFolder errors for any _dev/ directory found in the package.
+// _dev/ is a source-only artifact used during development (tests, deploy
+// configs, build manifests). It must not appear in a built package that is
+// validated with ModeBuild.
+func ValidateNoDevFolder(fsys fspath.FS) specerrors.ValidationErrors {
```
jsoriano: There would be any way to define this in the spec?

Maybe folder spec should be also aware of the validation mode, and it should not accept files marked with some "source" marker.

And we could reuse it for link files, or other files that should be there only in source packages.

code/go/internal/validator/semantic/validate_no_external_ecs.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/elastic/package-spec/v3/code/go/internal/fspath"
+	"github.com/elastic/package-spec/v3/code/go/pkg/specerrors"
+)
+
+// ValidateNoExternalEcs errors for any field referencing external: ecs.
+// The build process materializes ECS field references — once built, fields
+// should carry full definitions, not external pointers. A built package must
+// not contain any fields with external: ecs when validated with ModeBuild.
+func ValidateNoExternalEcs(fsys fspath.FS) specerrors.ValidationErrors {
```
jsoriano: I also wonder if this could be modeled in the spec. It is fine by now.

code/go/internal/validator/semantic/validate_stream_input_materialized.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// ValidateStreamInputMaterialized errors when build-mode manifests carry
+// source-only 'package:' fields that the build process should have materialized:
+//
+//   - data_stream/*/manifest.yml: each stream must have 'input:' set and must NOT
+//     have 'package:' (composable-input pattern, source-only).
+//   - manifest.yml: each policy_template input must have 'type:' set and must NOT
+//     have 'package:' (package-reference pattern, source-only).
+func ValidateStreamInputMaterialized(fsys fspath.FS) specerrors.ValidationErrors {
```
jsoriano: Nit. I think we call it "Bundled" in other places.
```suggestion
func ValidateStreamInputBundled(fsys fspath.FS) specerrors.ValidationErrors {
```

code/go/internal/validator/semantic/validate_stream_input_materialized.go
```diff
@@ -0,0 +1,164 @@
...
+	Inputs []policyTemplateInputMaterialization `yaml:"inputs"`
+}
+
+type packageMaterializationManifest struct {
+	Type            string                          `yaml:"type"`
+	PolicyTemplates []policyTemplateMaterialization `yaml:"policy_templates"`
+}
+
+// ValidateStreamInputMaterialized errors when build-mode manifests carry
+// source-only 'package:' fields that the build process should have materialized:
```
jsoriano: It would be also great if we could model this in the spec.

code/go/internal/validator/semantic/validate_stream_input_bundled.go
```diff
@@ -0,0 +1,164 @@
...
+				// errors would be misleading — the user intended the composable-input
+				// pattern and needs to materialise it, not simply add 'input:'.
+				continue
+			}
+			if s.Input == "" {
+				errs = append(errs, specerrors.NewStructuredErrorf(
+					"file %q: stream[%d] missing required 'input:' field",
+					fullPath, i,
+				))
+			}
```
jsoriano, line 104: This condition is already checked by the spec right? It checks that either `package` or `input` should be set.
I think this validator should only check that `package` is not defined in built packages. The rest is modeled in the spec.

code/go/internal/validator/semantic/validate_no_external_ecs.go
```diff
@@ -0,0 +1,29 @@
...
+	validateFunc := func(metadata fieldFileMetadata, f field) specerrors.ValidationErrors {
+		if f.External != "ecs" {
```
jsoriano: No external field should be allowed on built packages.
```suggestion
func ValidateNoExternalFields(fsys fspath.FS) specerrors.ValidationErrors {
	validateFunc := func(metadata fieldFileMetadata, f field) specerrors.ValidationErrors {
		if f.External == "" {
```

code/go/pkg/validator/validator_test.go
```diff
@@ -1480,9 +1485,13 @@ func TestValidateFromZip_modeRestrictions(t *testing.T) {
 
 	t.Run("build_allowed", func(t *testing.T) {
 		t.Parallel()
+		// Use the built fixture: source packages have _dev/ and external:ecs which
+		// build mode rejects. good_built is the correct built-package counterpart.
+		builtPkg := filepath.Join("..", "..", "..", "..", "test", "packages", "build_mode", "good_built")
```
jsoriano: Nit. Maybe place these built packages under a different directory.
```suggestion
		builtPkg := filepath.Join("..", "..", "..", "..", "test", "built_packages", "good_built")
```

jsoriano (dismissed):
Adding some suggestions and requesting some small changes.

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
teresaromero: i removed the check. i prefer the validation runs on source and build. source is run on elastic-package lint - build - install; if there is any missing file we will notice before getting to the bundle.

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
teresaromero: yes, i am keeping this as the source validation is mantained

code/go/internal/validator/semantic/types.go
teresaromero: by the spec definition there are only directories on the datastream dir, but, fs.ReadDir is gathering directories and files, so i thought it was interesting to add this guard at this layer to ensure it. based on the docs https://pkg.go.dev/io/fs#DirEntry the direntry interface could be one or the other...

code/go/internal/validator/semantic/validate_no_dev_folder.go
teresaromero: implemented this from the spec, both dev and .link files are checked based on mode on the spec. i was suggested to use the `visibility: private` setting which was shared with all _dev folder placements, but i thought this might have more semantical use so i created a sourceOnly boolean property for this case... should we use visbility field instead? i have mixed feelings as i think visibility could mean a broader scope...

code/go/internal/validator/semantic/validate_no_external_ecs.go
teresaromero: i was exploring how to add the mode to the fields and i think we could address this as a follow up enhacement, keeping the semantical validation for now. i will create an issue to discuss a solution

code/go/internal/validator/semantic/validate_stream_input_materialized.go
teresaromero: same as for ecs fields, this will followup on the field-mode awarenes

code/go/internal/validator/semantic/validate_stream_input_bundled.go
teresaromero: yes, you are right. i've removed the condition so the bundled is only checked for `package`, if there is no package, by spec its forced to have `input`. this will be not needed if we implement mode-aware fields. i will add it to the follow up issue

teresaromero:
Opening issue for a followup on fields mode-aware https://github.com/elastic/package-spec/issues/1188

code/go/internal/specschema/folder_item_spec.go
```diff
@@ -142,6 +148,9 @@ type folderItemSpec struct {
 	DevelopmentFolder  bool              `json:"developmentFolder" yaml:"developmentFolder"`
 	AllowLink          bool              `json:"allowLink" yaml:"allowLink"`
 
+	// SourceOnly marks items that must not appear in built packages (build mode).
+	SourceOnly bool `json:"sourceOnly" yaml:"sourceOnly"`
```
jsoriano: Great to see this working 👍 Thanks.

Could we make this more flexible and allow to set the mode the item should appear on? In case we need items that should only appear in built packages.

So instead of setting `sourceOnly: true`, we set `validationMode: source`.

code/go/internal/validator/semantic/validate_integration_policy_template_path.go
```diff
@@ -111,14 +111,9 @@ func ValidateIntegrationPolicyTemplates(fsys fspath.FS) specerrors.ValidationErr
...
-		// explicit template_path/template_paths is set, all templates come from
-		// the dependency and are only present after build. Skip those.
-		// If the composable input defines its own template_path or template_paths
-		// (overlay templates that live in the source package), those are validated.
-		if input.Package != "" && input.TemplatePath == "" && len(input.TemplatePaths) == 0 {
-			continue
-		}
+		// Only validate template files that are explicitly declared; if none are set there
+		// is nothing to check (composable inputs without overlay templates source them from
+		// the dependency package, which is absent from the source tree).
```
jsoriano, line 116: Thanks for making these comments more concise 👍

code/go/internal/specschema/folder_item_spec.go
teresaromero: yup, makes sense. i was thinking on this while creating the issue for the followup, to declare the "specific mode" instead of booleans. i will update the pr with these change

jsoriano (approved):
Thanks!

PR #1180, opened 2026-06-04
Add provider_permissions field for declaring integration permission requirements

jeniawhite:
## What does this PR do?

Adds a new optional provider_permissions field to the package manifest schema for both integration and input package types. The field can be declared at four levels - package, policy_template, input, and data_stream - allowing permission requirements to be expressed at the right granularity. Each entry in the array targets a specific provider (e.g. aws, gcp, azure, kubernetes) and carries two optional sub-fields:

- permissions: individual permission grants (e.g. s3:GetObject, storage.objects.get, Microsoft.Compute/virtualMachines/read)
- roles: pre-defined roles or managed policies to attach alongside inline permissions (e.g. SecurityAudit, roles/logging.viewer)

The field is introduced in spec version 3.7.0 and is removed via version patches for older format_version values.

## Why is it important?

Integration packages currently have no machine-readable way to declare what provider permissions they require. This means permission requirements live only in documentation, drift over time, and cannot be used by tooling. With provider_permissions in the spec, tooling can read exactly which permissions each enabled integration needs and generate appropriately scoped infrastructure-as-code (IaC) artifacts - rather than granting a static kitchen-sink policy. The permission declarations are versioned alongside the package, making them auditable and kept in sync with the data streams they support.

## Checklist

## Related issues

- Closes https://github.com/elastic/security-team/issues/16737

## Summary by CodeRabbit

* **New Features**
  * Added support for provider_permissions across package, policy template, input, and data stream levels for finer permission declarations.

* **Documentation**
  * Added changelog entry documenting provider_permissions support (in-development 3.6.4-next).

* **Tests**
  * Added test fixtures covering valid and invalid provider_permissions cases, including missing required fields and disallowed extra fields.


spec/changelog.yml
```diff
@@ -8,6 +8,9 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    - description: Add provider_permissions field to package, policy_template, input, and data_stream levels for declaring provider-specific permissions.
```
teresaromero, line 13: is there any change required on the kibana side for this to be released?

spec/integration/manifest.spec.yml
```diff
@@ -736,6 +736,85 @@ spec:
       type: string
       examples:
         - httpjson
+    provider_permissions:
+      description: >
+        Permissions and roles this integration unit requires from the named provider.
+        May be declared at package, policy_template, input, and data_stream levels.
```
teresaromero: Could we document how different levels of declaring this permissions work? what if i declare it at package and then at an input? does it override? Are there any cases where we might need to check semantically the meaning of this permissions across levels? Is there any case we want to avoid? or validate?

spec/integration/manifest.spec.yml
```diff
@@ -736,6 +736,85 @@ spec:
...
+      items:
+        type: object
+        additionalProperties: false
+        required:
+          - provider
+        properties:
+          provider:
+            description: >
+              Identifier of the provider these permissions apply to
+              (e.g. "aws", "gcp", "azure", "kubernetes", "okta").
```
teresaromero:
```suggestion
```

as we have them as example on the `examples` i think this might be redundant. Else, we can keep them here and perhaps add a single example (reducing the lines)

spec/changelog.yml
juliaElastic: we will use the provider_permissions from kibana to generate a list of required permissions in the onboarding flow: https://github.com/elastic/ingest-dev/issues/7927
there should be no dependency on kibana, the new fields are additive and won't be used on old kibana versions (should be smoke tested)

spec/input/manifest.spec.yml
```diff
@@ -172,6 +176,12 @@ spec:
 
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.7.0
+    patch:
+      - op: remove
+        path: "/properties/provider_permissions"
+      - op: remove
+        path: "/properties/policy_templates/items/properties/provider_permissions"
```
jsoriano: JSON patches would not be needed if this new setting doesn't affect existing versions of Fleet.

```suggestion
```

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -706,10 +706,17 @@ spec:
...
+      $ref: "../../integration/manifest.spec.yml#/definitions/provider_permissions"
   required:
   - title
 # JSON patches for newer versions should be placed on top
 versions:
+  - before: 3.7.0
+    patch:
+      # provider_permissions field for provider-specific permission declarations.
+      - op: remove
+        path: "/properties/provider_permissions"
```
jsoriano: JSON patches would not be needed if this new setting doesn't affect existing versions of Fleet.
```suggestion
```

spec/integration/manifest.spec.yml
```diff
@@ -991,6 +1076,17 @@ spec:
...
+    patch:
+      # provider_permissions field for provider-specific permission declarations.
+      - op: remove
+        path: "/properties/provider_permissions"
+      - op: remove
+        path: "/properties/policy_templates/items/properties/provider_permissions"
+      - op: remove
+        path: "/properties/policy_templates/items/properties/inputs/items/properties/provider_permissions"
+      - op: remove
+        path: "/definitions/provider_permissions"
```
jsoriano:
```suggestion
```

spec/changelog.yml
jsoriano: As this doesn't affect existing versions of Kibana we can release this in 3.6.4, and use it on any package.

jsoriano (commented):
FTR, as discussed over slack, as these new settings won't affect current versions of kibana, we can release this in next 3.6 version, and we can remove the version patches for 3.7.0.

This way this can be earlier used in packages without needing to bump their kibana or format versions.

jeniawhite:
@teresaromero @jsoriano 
Addressed your review comments, please re-review to approve to merge if no further action needed.
Thank you.

spec/changelog.yml
```diff
@@ -8,6 +8,11 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+- version: 3.6.4
+  changes:
+    - description: Add provider_permissions field to package, policy_template, input, and data_stream levels for declaring provider-specific permissions.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1180
```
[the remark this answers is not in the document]
jeniawhite: @jsoriano 
Since we want to release as part of `3.6.4`, I assume that pushing this to `3.7.0-next` is not the intended change. Please correct me if I'm wrong.

spec/changelog.yml
```diff
@@ -8,6 +8,11 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+- version: 3.6.4
```
```suggestion
- version: 3.6.4-next
```

we need to open a new "next" release for this change and when preparing the release we remove the suffix and release it.

jsoriano (approved):

PR #1189, opened 2026-06-18
Replace profiling data stream type with profiles

teresaromero:
## What does this PR do?

Replaces the `profiling` data stream type with `profiles` in the integration spec:

- `spec/integration/data_stream/manifest.spec.yml`: renames the enum value `profiling` → `profiles`
- `validate_profiling_nonga.go` → `validate_profiles_nonga.go`: renames the validator and its exported/unexported functions accordingly
- Introduces `tracesDataStreamType` and `profilesDataStreamType` constants in `types.go` (shared across the `semantic` package) to replace inline string literals
- Updates test fixtures and the error message assertion in `validator_test.go`

## Why is it important?

Mixed usage of `profiling` (integration data streams) and `profiles` (input packages, Fleet data stream type) was causing inconsistency across the Package Spec and Fleet. `profiles` is the correct term, aligning with the [OTel Profiles signal name](https://opentelemetry.io/docs/concepts/signals/profiles/).

The change is breaking but safe: the only published package in the integrations repository using this type is `profiling_otel`, which already uses `profiles` — no packages use the deprecated `profiling` string.

## Checklist

 > Note on version patches: No backward-compatibility patch has been added for before: 3.7.0. This is intentional — the rename is a hard break: 
 > profiling is invalid in all spec versions going forward, regardless of format_version. This is safe because no packages in the integrations repository currently use type: profiling.

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/7797

## Summary by CodeRabbit

* **Documentation**
  * Updated best practices guidance to use package-level constants for data stream type definitions instead of inline literals.

* **Bug Fixes**
  * Renamed data stream type from "profiling" to "profiles" across schema, validation rules, and test packages.

* **Chores**
  * Added internal constants for data stream types to maintain consistency across validators.


teresaromero:
test integrations

spec/changelog.yml
```diff
@@ -11,6 +11,9 @@
     - description: Add support for mode-aware constructors and validation APIs.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1177
+    - description: Replace `profiling` data stream type with `profiles` to align with the OTel Profiles signal name.
+      type: breaking-change
```
jsoriano: We can release it in 3.6 as bugfix, as this was not implementing what was expected.

AGENTS.md
```diff
@@ -376,6 +376,7 @@ func validateMyRule(manifest pkgpath.File) specerrors.ValidationErrors {
 5. **Error handling**: Return structured errors with file paths and context
 6. **Use descriptive variable names**: Avoid abbreviations in validators. Use full names like `packageName` (not `pkgName`), `dataStreamManifests` (not `dsManifests`), `templateIndex` (not `ptIdx`) for better readability.
 7. **Use fsys.Path() for error messages**: When creating error messages, use `fsys.Path("relative/path")` to get the full package path, not just the relative path from a file object. This ensures error messages match the test framework's expectations.
+8. **Data stream type constants**: Never use data stream type strings (e.g. `"traces"`, `"profiles"`) as inline literals. Declare them as package-level constants in `code/go/internal/validator/semantic/types.go` alongside the existing `tracesDataStreamType` and `profilesDataStreamType`. The canonical set mirrors the `type` enum in `spec/integration/data_stream/manifest.spec.yml`.
    ```go
```
jsoriano, line 380: This code snippet is related to the point 7 above, right?

jsoriano (changes requested):
LGTM, only a couple of things on the changelogs and the docs for agents.

AGENTS.md
teresaromero: yup! updated

jsoriano (approved):

PR #1196, opened 2026-06-25
code/go/internal/validator/semantic: narrow by-reference panel flagging to exclude saved searches

efd6:
Saved searches are inherently referenced and should not be flagged as by-reference panels. Restrict the check to visualization, lens, and map reference types only.

## Checklist

## Related issues

-


P1llus (approved):

teresaromero (commented):
could you add a changelog entry to the spec? also, can we relate any context or issue that triggered this change? thanks

teresaromero (approved):

PR #1209, opened 2026-07-20
Support dynamic IaC for integrations

jeniawhite:
## What does this PR do?
We would like to support dynamic IaC generation for integrations, this change will allow integrations to declare IaC template manifests that will be used to create dynamic IaC for them to be deployed in a frictionless manner.

## Checklist

## Related issues

-

### DIFF
--- a/spec/blueprint/manifest.spec.yml
+++ b/spec/blueprint/manifest.spec.yml
@@ -1,0 +1,54 @@
+##
+## Describes the specification for the blueprint package's main manifest.yml file
+##
+spec:
+  # Everything under here follows JSON schema (https://json-schema.org/), written as YAML for readability
+  type: object
+  additionalProperties: false
+  properties:
+    format_version:
+      description: The version of the package specification format used by this package.
+      $ref: "../integration/manifest.spec.yml#/definitions/version"
+    name:
+      description: The name of the package.
+      type: string
+      pattern: '^[a-z0-9_]+$'
+      examples:
+      - elastic_iac_blueprints
+    title:
+      $ref: "../integration/manifest.spec.yml#/definitions/title"
+    description:
+      $ref: "../integration/manifest.spec.yml#/definitions/description"
+    version:
+      description: The version of the package.
+      $ref: "../integration/manifest.spec.yml#/definitions/version"
+    source:
+      $ref: "../integration/manifest.spec.yml#/definitions/source"
+    type:
+      description: The type of package.
+      type: string
+      enum:
+      - blueprint
+      examples:
+      - blueprint
+    conditions:
+      description: Conditions under which this package can be installed.
+      type: object
+      additionalProperties: false
+      properties:
+        elastic:
+          $ref: "../integration/manifest.spec.yml#/definitions/conditions/properties/elastic"
+        kibana:
+          $ref: "../integration/manifest.spec.yml#/definitions/conditions/properties/kibana"
+    owner:
+      $ref: "../integration/manifest.spec.yml#/definitions/owner"
+    deprecated:
+      $ref: "../integration/manifest.spec.yml#/definitions/deprecated"
+  required:
+  - format_version
+  - name
+  - title
+  - description
+  - version
+  - type
+  - owner
--- a/code/go/pkg/validator/validator_test.go
+++ b/code/go/pkg/validator/validator_test.go
@@ -86,6 +86,21 @@ func Test_ValidateFromPath(t *testing.T) {
 				"field group: Does not match pattern '^[a-z0-9_]+$'",
 			},
 		},
+		"good_iac_blueprints":                    {},
+		"good_iac_blueprints_input":              {},
+		"bad_iac_blueprints_missing_required": {
+			"manifest.yml",
+			[]string{
+				`field iac_blueprints.0: format is required`,
+				`field iac_blueprints.0: patches is required`,
+			},
+		},
+		"bad_iac_blueprints_invalid_format": {
+			"manifest.yml",
+			[]string{
+				`field iac_blueprints.0.format: iac_blueprints.0.format must be one of the following: "cloudformation", "arm", "terraform", "deployment-manager", "helm", "kustomize", "ansible", "bicep"`,
+			},
+		},
 		"bad_duration_vars": {
 			"manifest.yml",
 			[]string{
--- a/code/go/pkg/validator/validator_test.go
+++ b/code/go/pkg/validator/validator_test.go
@@ -86,8 +86,9 @@ func Test_ValidateFromPath(t *testing.T) {
 				"field group: Does not match pattern '^[a-z0-9_]+$'",
 			},
 		},
-		"good_iac_blueprints":                    {},
-		"good_iac_blueprints_input":              {},
+		"good_iac_blueprints":       {},
+		"good_iac_blueprints_input": {},
+		"good_blueprint":            {},
 		"bad_iac_blueprints_missing_required": {
 			"manifest.yml",
 			[]string{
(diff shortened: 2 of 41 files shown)


teresaromero (changes requested):
Please, lets separate this PR into two where the new type is implemented and other where the changes on input and integrations referencing the new fields.

bhapas:
Per @teresaromero's review request to split this PR, the changes have been separated into two independent PRs off `main`:

1. **Blueprint package type** (+ related tests): https://github.com/elastic/package-spec/pull/1224
2. **`iac_blueprints` / scoped `provider_permissions` on integration & input** (+ related tests): https://github.com/elastic/package-spec/pull/1225

Please close this PR in favor of those two when convenient. Credit for the original work remains with @jeniawhite (co-authored on the split commits).

spec/blueprint/manifest.spec.yml
```diff
@@ -0,0 +1,54 @@
+##
+## Describes the specification for the blueprint package's main manifest.yml file
```
cmacknz, line 2: Do you need a separate blueprint package type? Isn't it enough to allow input and integration packages to define blueprints?

I would expect that the permissions and other definitions in a blueprint would need to be tied to specific integration versions, so it seems preferable to bundle them into the existing package types, rather than have a separate package the likely introduces the need for version compatibility rules.

It seems much nicer to be able to test the AWS package as a complete unit for example.

cmacknz: CC @mrodm @teresaromero

bhapas: @cmacknz How do we deal if there are like multiple, like 20 odd AWS integrations that need same blueprint , If we limit the blueprint inside one integration then probably we would have to duplicate it for other similar integrations?

The purpose was to have one place of truth for all these blueprints that could be used by multiple packages. Reusing `input` / `integration` types might causes problems in Fleet UI if we do not have any Kibana/ES assets probably? 

If the naming needs to be generic , like to reuse the package type for any general purpose template / prompt / text / asset storage mechanism then we can use more generic / base package naming convention.

cmacknz: Needing to reuse these is valid, but creating a separate package type creates a compatibility check problem as I mentioned so if reuse is the only reason for a separate package we maybe need another way to reuse configuration in packages.

It should be possible for us to create a way to reuse configuration across integrations without creating a new releasable user artifact. We should not be publishing packages whose main purpose is just config reuse in other packages.

teresaromero: i was thinking on this re-use and had the links idea since the last meeting, so i went ahead and checked this out.

The proposal: use elastic-package .link files to share the canonical blueprint asset across integration packages.

The .link mechanism already exists — a .link file is a plain text pointer (<relative-path> <sha256>) that elastic-package resolves transparently at build time. The built package zip contains the real file content, not the pointer. elastic-package links update/check keeps checksums in sync and fails the build on drift.

I ran a POC on the elastic/integrations repo to verify this approach: https://github.com/elastic/integrations/pull/20721 

- Created account.cloudformation.json.link files inside the iac/ directories of aws, cloud_asset_inventory, and cloud_security_posture, all pointing to a single shared source.
- elastic-package links update computed checksums for all three.
- elastic-package build on cloud_asset_inventory produced a zip with iac/account.cloudformation.json — real content, correct filename — confirming that LinksFS resolves iac/*.link at build time, not just elasticsearch/ingest_pipeline/ and fields/.
- Cross-package links (pointing outside a package's own tree) work fine.

Where to store the shared source:

1. shared/blueprints/aws/federated-identity/account.cloudformation.json at repo root — neutral, no ownership coupling.
2. Inside one authoritative integration (e.g., aws) — makes ownership explicit but creates a dependency.

What this avoids: no new type: blueprint in the spec, no EPR changes, no new elastic-package support for a new package type. The canonical template is just a file — shared via a mechanism the toolchain already understands.

@bhapas  The iac folder will contain "blueprint" + patches; would the service that patches permissions be able to grab from this folder all the files? 

@mrodm will this be a challenge for backporting a package? having a linked file source on root repo? would it be better to have it on a package and link the other from there?

from this we could also update elastic-package so it can simulate the permissions patching and have this asset tested during development; as we do now with policies with fleet (thinking out loud)

mrodm: > @mrodm will this be a challenge for backporting a package? having a linked file source on root repo? would it be better to have it on a package and link the other from there?

@teresaromero  Yes, I think linking files between different packages is going to be problematic when doing backports.

Currently, we suggest that new backport branches should be created just with the required package and remove the others. If the linked file is going to be located in another package, when creating the backport branch it is likely that the linked file does not exist. In that case one option is that the owners should copy the source of the linked file from the commit where the backport branch was created.

In order to support that, scripts that copy the contents of the backport branches should also to take into account the source of the linked files to ensure that any other necessary package is kept. And this is not considered currently in this process.

If the linked file is in the root of the repository could also cause some issues. For instance, when a change in that blueprint folder in the root of the repository is introduced, it is not going to be easy to know to which packages affect and which branches should be backported. If possible, I would try to avoid this (thinking in backports).

EDIT:

And for instance, changes in the packages containing the source of the links will not be suggested as part of the backport process automation.

It is difficult to ensure that the backport process automation would take care of all these scenarios.

mrodm: Created issue to ensure that the source packages of the linked files are copied too https://github.com/elastic/integrations/issues/20726

mrodm: Created issue to ensure that the source packages of the linked files are copied too https://github.com/elastic/integrations/issues/20726

bhapas: @cmacknz Agree that we should not be publishing packages that just package a reusable config for other [ specific set of ] packages.

@teresaromero 

>The .link mechanism already exists — a .link file is a plain text pointer ( ) that elastic-package resolves transparently at build time. The built package zip contains the real file content, not the pointer. elastic-package links update/check keeps checksums in sync and fails the build on drift.

This sounds good. Except that I have a few questions around this.

- This approach means that every built package will have a copy of blueprint in it , right?
- Let's say , we want to develop a new development version of the blueprint  [ Like .../aws/fi/account-v2-preview ] , I guess this should be possible?
- If we want multiple versions of blueprints to co-exist I guess we can just create additional folders for the versions [ aws/federated-identity/account-v1 ] [ aws/federated-identity/account-v2 ] where major bump is for non-backwards compatible changes to the template. 
- For simplicity , I guess each package has to have one base template reference at all levels - policy_template , data_stream , integration , input levels.
- When there is a base template update , I guess all the downstream packages need a new version release [ Probably rightly so , rather than silently updating a template ].


>@bhapas The iac folder will contain "blueprint" + patches; would the service that patches permissions be able to grab from this folder all the files?

Yes , it should be possible if we can read the specific resources from the gcloud bucket.

>@mrodm will this be a challenge for backporting a package? having a linked file source on root repo? would it be better to have it on a package and link the other from there?

For backport branches , specific package versions can have specific base template instead of a link. That way we work with the specific template manually in that backport branch.

cmacknz: The link concept solves this, the backporting issues can be solved through the tooling updates Mario suggested hopefully. One day might just have to commit to maintaining backport branches of the entire integrations repository if this is troublesome.

I don't think the link concept prevents anything related to breaking changes or multiple versions of a blueprint at once. You can change what the links point to, have current and next versions of a blueprint etc. they would just all coexist on the main branch. Potentially you might want a version 

Maybe we need to iterate on the ergonomics of this a bit but conceptually this direction makes sense to me.

PR #1213, opened 2026-07-29
Add optional top-level `group` field to package manifests

teresaromero:
## What does this PR do?

Adds an optional top-level `group` field to `manifest.yml` for **integration**, **input**, and **content** package types.

```yaml
format_version: 3.5.0
name: nginx_input_otel
title: Nginx (OpenTelemetry)
type: input
group: nginx
# ...
```

Field contract:
- **Type**: `string`, optional
- **Pattern**: `^[a-z0-9_]+$` (same as package `name`)
- **Semantics**: opaque group id — all packages sharing the same value are members of the same marketplace group
- **No existence validation** — the spec does not validate the id against a known list; Fleet ignores unknown ids and group activation also requires a Kibana descriptor entry

## Why is it important?

The Elastic marketplace needs a way to collapse related packages (e.g. `nginx`, `nginx_input_otel`, `nginx_ingress_controller`) into a single tile. Kibana owns the tile title/icon/description via a static `INTEGRATION_GROUPS` config keyed by the group id; packages only declare membership.

**No version gate** — the field is pure metadata and requires no Fleet/Kibana runtime support. Other recently-added optional fields with the same nature (`fips_compatible`, `provider_permissions`) are also ungated. Gating behind a new `format_version` would force every adopting package to bump its format version, pulling in unrelated schema constraints. Most packages in the integrations repo are at `3.0.x`–`3.5.x` (e.g. `nginx` at `3.0.2`, `nginx_input_otel` at `3.5.0`); those packages can adopt `group` as-is.

Content packages are included for schema parity even though their tiles are hidden by default.

## Checklist

  - `good_integration_group` — valid `group: nginx` on an integration (format_version 3.3.2)
  - `good_input_group` — valid `group: nginx` on an input package (format_version 3.5.0)
  - `bad_integration_group` — invalid value (`"Nginx OTel"`) fails the pattern
  - `bad_input_group` — invalid value (`"Nginx-OTel"`) fails the pattern

## Related issues

- Part of https://github.com/elastic/ingest-dev/issues/8082

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -19,7 +19,7 @@
       link: https://github.com/elastic/package-spec/pull/1196
     - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
       type: enhancement
-      link: https://github.com/elastic/package-spec/pull/XXXX
+      link: https://github.com/elastic/package-spec/pull/1213
 - version: 3.6.5
   changes:
     - description: Resolve test requires `source` paths relative to the package root instead of the test config directory.
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -17,6 +17,8 @@
     - description: Remove the search type from by-reference validation checks.
       type: bugfix
       link: https://github.com/elastic/package-spec/pull/1196
+- version: 3.6.6-next
+  changes:
     - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1213
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -4,9 +4,6 @@
 ##
 - version: 3.7.0-next
   changes:
-    - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
-      type: enhancement
-      link: https://github.com/elastic/package-spec/pull/1213
     # Pending on https://github.com/elastic/kibana/issues/220294
     - description: Add support for semantic_text field definition.
       type: enhancement
@@ -20,6 +17,9 @@
     - description: Remove the search type from by-reference validation checks.
       type: bugfix
       link: https://github.com/elastic/package-spec/pull/1196
+    - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/XXXX
 - version: 3.6.5
   changes:
     - description: Resolve test requires `source` paths relative to the package root instead of the test config directory.
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -4,6 +4,9 @@
 ##
 - version: 3.7.0-next
   changes:
+    - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1213
     # Pending on https://github.com/elastic/kibana/issues/220294
     - description: Add support for semantic_text field definition.
       type: enhancement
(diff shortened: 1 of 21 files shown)


spec/changelog.yml
```diff
@@ -17,6 +17,9 @@
     - description: Remove the search type from by-reference validation checks.
       type: bugfix
       link: https://github.com/elastic/package-spec/pull/1196
+    - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
```
teresaromero, line 22: i am adding this to 3.7 as i understand is a "new feature" for a new minor (not breaking change) but this will mean to move some pending things into a future 3.8 .

wdyt @mrodm I dont like the idea of doing this on a patch :/

spec/changelog.yml
```diff
@@ -17,6 +17,9 @@
...
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1213
```
mrodm, line 24: Is there any error when installing the package with this `group` field defined ? 

If there is no error, I think we could be moving this entry to a 3.6.6-next section since it would not require changes in Kibana.

spec/changelog.yml
mrodm: According to how it is managed the specification (versioning) here in the spec, I think this could be moved to a new section `3.6.6-next`. Since it does not require changes in Kibana , or at least adding it, it will not cause errors. Is my understanding correct?

spec/integration/manifest.spec.yml
```diff
@@ -549,6 +549,17 @@ spec:
...
         - since
         - description
+    group:
+      description: >
+        Opaque identifier declaring membership in a marketplace group. All packages
+        that share the same value are considered members of the same group.
+        Kibana owns tile title/icon/description via a static descriptor keyed by
+        this id; packages only declare membership.
+      type: string
+      pattern: '^[a-z0-9_]+$'
```
mrodm, line 559: I wonder if could set some kind of enumeration here to validate the values before those packages are being released cc @Supplementing

spec/integration/manifest.spec.yml
```diff
@@ -549,6 +549,17 @@ spec:
       required:
         - since
         - description
+    group:
+      description: >
+        Opaque identifier declaring membership in a marketplace group. All packages
```
mrodm: what do you mean with "Opaque identifier"?

spec/integration/manifest.spec.yml
teresaromero: i don't think we can add here any enum https://github.com/elastic/ingest-dev/issues/8082#issuecomment-5092435697

spec/changelog.yml
teresaromero: i've unsed local ngnix with the group field, build and install with elastic-package using local package-spec and it works ok.

spec/integration/manifest.spec.yml
teresaromero: This was handpicked by my friend Claude and i understood it was an identifier which we did not have access as semantics live in Kibana... i agreed as it make sense in my head, but... asking Cursor ...

> Your reading is close, but "opaque" is doing a slightly different job than “from Kibana’s list.”
> 
>   In this wording, opaque means: treat the string as a bare key. Don’t parse meaning from it, don’t derive title/icon/description from it, and don’t expect package-spec to know which values are valid. Equality is what matters — same group → same marketplace group.
>   Kibana owns the display descriptor keyed by that id; packages only declare membership.
> 
>   That matches the design from ingest-dev#8082 (https://github.com/elastic/ingest-dev/issues/8082): group existence lives in Kibana’s 
>   constants; Fleet ignores unknown groups; package-spec only checks format (^[a-z0-9_]+$).

I am changing the description so its more clear :D thanks!

mrodm (approved):
LGTM

PR #1214, opened 2026-07-29
spec/{input,integration}/_dev/test/config.spec.yml: allow policy test s to ignore fields

efd6:
## What does this PR do?

Allows (dynamic) fields to be ignored in integration policy tests.

## Why is it important?

Dynamic fields break existing tests.

## Checklist

## Related issues

- For elastic/elastic-package#3771

### DIFF
--- a/spec/changelog.yml
+++ b/spec/changelog.yml
@@ -17,9 +17,6 @@
     - description: Remove the search type from by-reference validation checks.
       type: bugfix
       link: https://github.com/elastic/package-spec/pull/1196
-    - description: Allow policy tests to ignore fields.
-      type: enhancement
-      link: https://github.com/elastic/package-spec/pull/1214
 - version: 3.6.6-next
   changes:
     - description: Add support for saved search assets in content packages.
@@ -28,6 +25,9 @@
     - description: Add optional top-level `group` field to integration, input, and content package manifests for declaring marketplace group membership.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/1213
+    - description: Allow policy tests to ignore fields.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1214
 - version: 3.6.5
   changes:
     - description: Resolve test requires `source` paths relative to the package root instead of the test config directory.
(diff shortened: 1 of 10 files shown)


spec/integration/_dev/test/config.spec.yml
```diff
@@ -21,6 +21,29 @@ spec:
...
+          items:
+            $ref: "../../data_stream/_dev/test/config.spec.yml#/definitions/package_requirement"
+        ignore_fields:
+          description: >
+            Stream-level field paths to strip before comparing the actual policy
+            against the expected file. Paths use dot notation scoped to the stream
+            level (e.g. "state.user_agent" strips inputs[].streams[].state.user_agent).
+          type: array
+          items:
+            type: string
```
teresaromero, line 41:
```suggestion
      allOf:
        - $ref: "#/definitions/config_tests"
        - type: object
           additionalProperties: false
            properties:
               ignore_fields:
                  description: >
                     Stream-level field paths to strip before comparing the actual   policy against the expected file. Paths use dot notation scoped to the stream level (e.g. "state.user_agent" strips inputs[].streams[].state.user_agent).
                    type: array
                    items:
                        type: string
```

can we do something like this so the common is taken from the config test and its clear we are adding a specific policy field? not tested, i am just thinking on avoiding common fields to be repeated

teresaromero (dismissed):
tested locally ok. Some tests packages could be added just adding the new field to good_input and good_v3 ; perhaps add a bad case for adding the field at a system config. 

thanks

spec/changelog.yml
```diff
@@ -17,6 +17,9 @@
     - description: Remove the search type from by-reference validation checks.
       type: bugfix
       link: https://github.com/elastic/package-spec/pull/1196
+    - description: Allow policy tests to ignore fields.
+      type: enhancement
+      link: https://github.com/elastic/package-spec/pull/1214
```
mrodm: As this change does not require any support in Kibana or in the stack, it should be moved to `3.6.6-next`

teresaromero (approved):

mrodm (approved):

PR #1145, opened 2026-04-21
Prepare release 3.6.1

teresaromero:
## What does this PR do?

Prepares the **3.6.1** package-spec release: finalizes `spec/changelog.yml`, points latest-stack compliance CI at spec **3.6.1**, refreshes the Go patch in `.go-version`, and updates `AGENTS.md` examples to match.

Follow-up commits on this branch also:

- Move the Fleet UI enhancements from [PR #1133](https://github.com/elastic/package-spec/pull/1133) into the **3.6.1** changelog (top-level `sections` / per-var `section`, and `show_divider` on inputs).
- Remove `@skip` from the OTel `dynamic_signal_types` dataset compliance scenarios in `compliance/features/dataset.feature` so those scenarios run again.

## Why is it important?

Ships accumulated spec changes as **3.6.1** so integrators can depend on the released semver and CI validates the stack against that version. Re-enabling the dataset scenarios keeps compliance coverage aligned with current routing behavior for dynamic signal types.

### Release contents (from `spec/changelog.yml` for 3.6.1)

- **Enhancement:** Add var_groups support to policy template and input levels in integration packages, and to policy template and package root levels in input packages. ([PR #1120](https://github.com/elastic/package-spec/pull/1120))
- **Enhancement:** Add support for named inputs in policy templates to disambiguate multiple inputs of the same type. ([PR #1135](https://github.com/elastic/package-spec/pull/1135)) — changelog notes pending [Kibana #262138](https://github.com/elastic/kibana/pull/262138).
- **Enhancement:** Add validation for Fleet-reserved variables (`use_apm`, `data_stream.dataset`) to enforce correct types and constraints when explicitly overridden in package manifests. ([PR #1134](https://github.com/elastic/package-spec/pull/1134))
- **Enhancement:** Add transform `settings.num_failure_retries` option. ([Issue #1124](https://github.com/elastic/package-spec/issues/1124))
- **Enhancement:** Add `geo_shape` as a valid field type for integrations. ([PR #1132](https://github.com/elastic/package-spec/pull/1132))
- **Enhancement:** Allow `_dev/build/docs` folder in content packages to support README auto-generation. ([PR #1144](https://github.com/elastic/package-spec/pull/1144))
- **Enhancement:** Add top-level `sections` list for defining named section headers, and `section` attribute on vars for assigning variables to a section in the Fleet UI. ([PR #1133](https://github.com/elastic/package-spec/pull/1133))
[...]

PR #1216, opened 2026-07-31
Prepare release 3.6.6

teresaromero:
## What does this PR do?

Preparation for release 3.6.6

- update go minor to 1.26.5
- update changelog and move some entries to this version
- updated compliance

## Why is it important?

This release includes some new enhacements. See changelog

## Checklist


.buildkite/pipeline.trigger.compliance.tests.sh
```diff
@@ -28,8 +28,9 @@ steps:
 EOF
 
 # Generate each test we want to do.
-compliance_test 9.5.0-SNAPSHOT 3.6.4
```
teresaromero, line 31: @mrodm i am not sure which versions we should update here; 9.5 is not yet released.. and 9.6 is already in snapshot.

mrodm: We could be using snapshot versions also for 9.4.x WDYT ?
```sh
compliance_test 9.6.0-SNAPSHOT 3.6.6
compliance_test 9.5.0-SNAPSHOT 3.6.6
compliance_test 9.4.5-SNAPSHOT 3.6.6
```

And maybe we could update the test related to 8.19.x to the latest stable version too:
```
compliance_test 8.19.19 3.4.2
```

mrodm (commented):
Should we update AGENTS.md as in this PR ? #1145 

Could we remove the skip from this compliance test ? 
https://github.com/elastic/package-spec/blob/ee85d0922ffa8e389ebcdf05a3390a5f058c72b2/compliance/features/package-dependencies.feature#L4-L9

AGENTS.md
```diff
@@ -249,11 +249,11 @@ go run github.com/elastic/elastic-package stack up -d --version 9.4.0-SNAPSHOT
 eval $(go run github.com/elastic/elastic-package stack shellinit)
 cd ..
 
-# Run compliance tests against spec version 3.6.4
-TEST_SPEC_VERSION=3.6.4 make -C compliance test
+# Run compliance tests against spec version 3.6.5
```
mrodm: Should it be updated to 3.6.6 and in all the document?


```suggestion
# Run compliance tests against spec version 3.6.6
```

mrodm (dismissed):
LGTM
Just a minor comment about `AGENTS.md` file. Probably, this could be updated in a follow-up too.

AGENTS.md
teresaromero: yes, thanks!

mrodm (approved):
Thanks!

PR #1250, opened 2026-09-14
add logsdb, logsdb_columnar, and columnar index modes

mrg-elastic:
Add three new values to the elasticsearch.index_mode enum (spec 3.7.0):
- logsdb: logs profile with default sort on host.name / @timestamp
- logsdb_columnar: logsdb profile + columnar storage (tech preview)
- columnar: base columnar mode without logs-profile defaults (tech preview)

Introduce ValidateColumnarModeConstraints for data streams using either columnar mode. Hard errors: doc_values: false and mapping-level runtime fields (Elasticsearch rejects these at template install time). Filterable warnings (SVR00011-SVR00013): nested types (limited support), dynamic: false and enabled: false on objects (data-loss risk — columnar has no stored _source).

The validator is registered as warnOn since 3.7.0 so warnings can be suppressed per-package via validation.yml while hard blockers always fail.

## Checklist

## Related issues

-


spec/integration/data_stream/manifest.spec.yml
```diff
@@ -295,9 +295,12 @@ spec:
         This setting must be installed in the composable index template, not in the package component templates.
       type: string
       enum:
-      - "time_series" # Enables time series data streams https://www.elastic.co/guide/en/elasticsearch/reference/master/tsds.html
+      - "time_series"     # Enables time series data streams (TSDB)
+      - "logsdb_columnar" # Tech preview: logsdb profile + columnar storage (no inverted indexes for non-text fields)
+      - "columnar"        # Tech preview: base columnar mode (no logs-profile defaults)
```
tommyers-elastic, line 300: do you think we'll ever need plain columnar mode for integrations?

code/go/pkg/validator/validator_test.go
```diff
@@ -74,6 +74,34 @@ func Test_ValidateFromPath(t *testing.T) {
 		"good_provider_permissions_input":        {},
 		"good_integration_group":                 {},
 		"good_input_group":                       {},
+		"good_columnar_index_mode":               {},
+		"good_columnar_base_mode":                {},
+		"good_columnar_override":                 {},
+		"good_columnar_supported":                {},
+		"bad_columnar_supported_blocker": {
+			"data_stream/logs/fields/base-fields.yml",
```
teresaromero, line 83: you will need to use there a path that is transformed to windows (path.Join()? - haven't verified)

Issue #1260, opened 2026-09-21
Add toggle for .otel suffix index patterns


P1llus:
## Problem

Fleet today applies the `.otel` index-template suffix based on whether a data stream has an `otelcol` input. Integration packages that ship only field mappings and ingest pipelines (no agent inputs) cannot opt into that naming.

## Proposal

Add an optional boolean `use_otel_suffix` on integration data stream manifests (`data_stream/*/manifest.yml`). When true, Fleet should install index templates with the `.otel` suffix (same naming as otelcol), without requiring an input.

## Checklist

- [x] package-spec: add `use_otel_suffix` to data stream manifest schema (3.7.0-next)
- [ ] Kibana Fleet: honor `use_otel_suffix` in `getRegistryDataStreamAssetBaseName` / install-time naming
- [x] Compliance tests: assert `.otel-*` pattern for a no-input integration package

PR #1261, opened 2026-09-21
Add otel suffix toggle

P1llus:
## What does this PR do?

Currently we only build/install component templates with the `.otel` index pattern suffix when the input is `otelcol`, this does not provide any potential ways for integrations that might just define datastreams, field mappings or pipelines (non content-type assets) for `.otel` based indices.

PR is based on initial discussions we have had, which was mainly between a more generic "custom suffix" argument, or just a toggle between otel/ecs based suffixes, the latter was chosen for now.

## Why is it important?

Blocks the possibility to deploy assets required for Cloud related services, as the data is sent outside agent policies but to a `.otel` destination

## Checklist

## Related issues

Relates: https://github.com/elastic/package-spec/issues/1260


juliaElastic (dismissed):
LGTM

spec/changelog.yml
```diff
@@ -8,6 +8,10 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+    # Pending on https://github.com/elastic/package-spec/issues/1260
```
teresaromero: here we will need the kibana relaated change

spec/changelog.yml
```diff
@@ -8,6 +8,10 @@
...
+    - description: Add optional use_otel_suffix on integration data streams so index templates can use the .otel suffix without an otelcol input.
+      type: enhancement
+      link: TBD
```
```suggestion
      link: https://github.com/elastic/package-spec/pull/1261
```

spec/integration/data_stream/manifest.spec.yml
```diff
@@ -627,6 +627,17 @@ spec:
       description: if true, the index pattern in the ES template will contain the dataset as a prefix only
       type: boolean
       default: false
+    use_otel_suffix:
```
teresaromero, line 630: few questions related to this:

- would this just apply to integrations data streams; what happens when a stream is referenced under the package manifest policy template?
- we need some semantical validation to prevent this is used when inputs are in place; what would happen if this value is used with beats packages instead of otel ones?
- As far as i know, the suffix is added by the collector , not fleet; how is fleet intended to manage this in order to add the suffix?

teresaromero (commented):
thanks, i think we should narrow the change with some semantical validation so its not used when the package has inputs (for example, check if there is any other situation to define when is good to use)

Also, in order for this feature to be scoped to 3.7; should we add a spec patch? cc @mrodm

spec/changelog.yml
P1llus: I have a draft one that il open soon and link here :+1:

spec/integration/data_stream/manifest.spec.yml
P1llus: Policy templates should not really be relevant here since they usually do not define a datastream specifically? They are more focused on agent policies, while this flag do not impact agent policies at all, just asset installations.

If a data stream with the flag enabled was used in a beats agent policy the assets index pattern would be wrong, but it is disabled by default and should not provide any difference for any existing integrations today.
I can add in some validation for that usecase just to prevent it in the integration itself + another for the kibana side if needed, but installation of assets is checked differently than agent policies.

The suffix is already added to fleet installed assets (component templates that defines ingest pipelines and field mapping), the kibana side simply apply a separate condition to the existing "isOtelDataStream" type check.

test/packages/bad_integration_otel_suffix_old_version/data_stream/logs/manifest.yml
teresaromero, line 1: could we have this bad example to trigger the semantic validation instead of the additional field related to package versions?

code/go/internal/validator/spec.go
```diff
@@ -264,6 +264,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateNoExternalFields, modes: []Mode{BuildMode}},
 		{fn: semantic.ValidateStreamInputBundled, modes: []Mode{BuildMode},
 			types: []string{"integration"}},
+		{fn: semantic.ValidateUseOtelSuffix, types: []string{"integration"}, since: semver.MustParse("3.7.0")},
```
teresaromero, line 267:
```suggestion
		{fn: semantic.ValidateUseOtelSuffix, types: []string{"integration"}
```

Are we requiring packages to be on 3.7 to adopt this? related to my patch question earlier; if we open this to all versions (although is released since 3.7 validation) it will not force packages to adopt it by upgrading the spec

teresaromero (commented):
you need to run `make check` locally to solve the ci

code/go/internal/validator/semantic/validate_use_otel_suffix.go
```diff
@@ -0,0 +1,76 @@
...
+		return specerrors.ValidationErrors{specerrors.NewStructuredError(err, specerrors.UnassignedCode)}
+	}
+
+	var errs specerrors.ValidationErrors
+	for _, dataStreamName := range dataStreamNames {
+		manifestPath := path.Join(dataStreamDir, dataStreamName, "manifest.yml")
+		data, err := fs.ReadFile(fsys, manifestPath)
+		if err != nil {
+			errs = append(errs, specerrors.NewStructuredErrorf(
+				"file %q is invalid: failed to read manifest: %w", fsys.Path(manifestPath), err))
```
teresaromero:
```suggestion
				"file \"%s\" is invalid: failed to read manifest: %w", fsys.Path(manifestPath), err))
```

code/go/internal/validator/semantic/validate_use_otel_suffix.go
```diff
@@ -0,0 +1,76 @@
...
+			continue
+		}
+
+		var manifest useOtelSuffixManifest
+		if err := yaml.Unmarshal(data, &manifest); err != nil {
+			errs = append(errs, specerrors.NewStructuredErrorf(
+				"file %q is invalid: failed to parse manifest: %w", fsys.Path(manifestPath), err))
```
teresaromero:
```suggestion
				"file \"%s\" is invalid: failed to parse manifest: %w", fsys.Path(manifestPath), err))
```

code/go/internal/validator/semantic/validate_use_otel_suffix.go
```diff
@@ -0,0 +1,76 @@
...
+		if !manifest.UseOtelSuffix {
+			continue
+		}
+
+		var inputs []string
+		for _, stream := range manifest.Streams {
+			if stream.Input == "" {
+				continue
+			}
+			inputs = append(inputs, fmt.Sprintf("%q", stream.Input))
```
teresaromero:
```suggestion
			inputs = append(inputs, fmt.Sprintf(\"%s\", stream.Input))
```

code/go/internal/validator/semantic/validate_use_otel_suffix.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+		if len(inputs) == 0 {
+			continue
+		}
+
+		errs = append(errs, specerrors.NewStructuredError(fmt.Errorf(
+			"file %q is invalid: use_otel_suffix is only allowed on data streams without inputs (inputs defined: %s). If this data stream needs an input, use the otelcol input type instead, which already applies the .otel index pattern suffix",
```
teresaromero:
```suggestion
			"file \"%s\" is invalid: use_otel_suffix is only allowed on data streams without inputs (inputs defined: %s). If this data stream needs an input, use the otelcol input type instead, which already applies the .otel index pattern suffix",
```

teresaromero (commented):
Windows error might be because of this...

P1llus:
> Windows error might be because of this...

Yeah it was mainly because of full path being used together with q which escapes backslashes, I fixed 3 of the 4 you mentioned because the fourth did not return a path so it wouldn't have the same issue.

PR #1265, opened 2026-09-23
feat(validation): warn on package-level conditions.agent.version (SVR00011)

teresaromero:
## Why

`conditions.agent.version` at the package level is deprecated and being phased out. Packages using it should migrate to per-data-stream agent conditions. This PR introduces a validation warning (SVR00011) so authors are notified during `lint`/`check` rather than discovering it after the field is removed.

Resolves elastic/ingest-dev#9431.

## What

- Adds `SVR00011`: emits a deprecation warning when `conditions.agent.version` is present in a package manifest
- Updates `spec/integration/manifest.spec.yml` to mark the field as deprecated
- Adds changelog entry and documents the new validation in `docs/validations.md`

## Testing

Tested locally by pinning this change in an `elastic-package` build and running `epb lint` against the `aws` package:

```
Warning: package-level conditions.agent.version is deprecated, and will be disallowed in a future spec version (SVR00011)
```

Warning fires as expected; no other regressions observed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


teresaromero:
test integrations

juliaElastic (dismissed):
The warning LGTM.
Before we close the issue, I think we need a follow up to attempt removing the condition from packages that have it: https://github.com/elastic/ingest-dev/issues/9431#issuecomment-5775115745
We could create (draft) PRs in integrations and discuss alternatives with the package owners.

spec/changelog.yml
```diff
@@ -8,6 +8,11 @@
     - description: Add support for semantic_text field definition.
       type: enhancement
       link: https://github.com/elastic/package-spec/pull/807
+- version: 3.6.7-next
+  changes:
+    - description: Warn when a package declares a package-level `conditions.agent.version`.
+      type: deprecation
+      link: https://github.com/elastic/package-spec/pull/TBD
```
mrodm, line 15:
```suggestion
      link: https://github.com/elastic/package-spec/pull/1265
```

spec/integration/manifest.spec.yml
```diff
@@ -224,7 +224,12 @@ spec:
           properties:
             version:
               type: string
-              description: Elastic Agent versions compatible with this package.
+              deprecated: true # See https://github.com/elastic/ingest-dev/issues/9431
```
mrodm, line 227: This is an internal issue, could we add the PR link ? Should we create an issue in package-registry for this instead?

docs/validations.md
```diff
@@ -135,3 +137,53 @@ on_failure:
         in pipeline '{{{ _ingest.pipeline }}}'
         failed with message '{{{ _ingest.on_failure_message }}}'
 ```
+
+## SVR00010 - Integration input qualifier required
+[SVR00010]: #svr00010---integration-input-qualifier-required
+
+**Available since [3.6.0](https://github.com/elastic/package-spec/releases/tag/v3.6.0)**
```
mrodm, line 144: What about adding the text that was set in this other PR ?

https://github.com/elastic/package-spec/pull/1244/changes#diff-5da8bd75fdd3133e874744d4d616b1b66412fde5a24d3093f3a488409c0814f7R146-R160


> When a policy template declares more than one input of the same `type`, every one of
> them must have a `name`. Data streams refer to inputs by type, so without names Fleet
> cannot tell which of the inputs a stream belongs to. This is common with `otelcol`.
> 
> ```yaml
> policy_templates:
>   - name: nginx
>     inputs:
>       - type: otelcol
>         name: nginx_metrics
>         title: Nginx metrics
>       - type: otelcol
>         name: nginx_logs
>         title: Nginx logs
> ```
>

code/go/internal/validator/spec.go
```diff
@@ -244,7 +244,7 @@ func (s Spec) rules(pkgType string, rootSpec spectypes.ItemSpec) validationRules
 		{fn: semantic.ValidateInputPackagesPolicyTemplates, types: []string{"input"}},
 		{fn: semantic.ValidateInputDynamicSignalTypes, since: semver.MustParse("3.6.0")},
 		{fn: semantic.ValidateFleetReservedVars, types: []string{"integration", "input"}, since: semver.MustParse("3.6.1")},
-		{fn: semantic.ValidateMinimumAgentVersion},
+		{fn: warnOn(semantic.ValidateMinimumAgentVersion)},
```
mrodm, line 247: Should it be reported as an error now (e.g. in 3.7.0) or should we wait for that ?

teresaromero: this is somthing i was wondering, why do a phased deprecation if this needs to be fixed anyways on the integrations side?

not sure what is the timeline on the kibana side for the migration to take place? if we go with making this error exit we could open the fixes on the packages ourselves when we bump elastic package or even before the release so they are prepared. @juliaElastic for the changes on the packages, are those ready to be done so fleet can use it? related to this comment https://github.com/elastic/package-spec/pull/1265#pullrequestreview-5288425118

